PR #846, opened 2014-12-02
Fixes #838 Entity transform timezone

### LINKED ISSUE #838: Last activity indicator (messages inbox) shows incorrect time
Just after creating a new message, the inbox overview shows that the last activity was 1 hour ago.

When opening the message thread, the message creation time does display correctly.

rap1ds:
**Fix Time.parse timezone issue**

By default, Ruby parses string times to the system timezone, if the string doesn't contain timezone information. This is problematic, since time is stored to MySQL in UTC, without timezone information. Thus it should be parsed to UTC, not to system's timezone.

**Disable ActiveRecord timezone awareness**

The timezone was set to Helsinki, which is obviously wrong. We don't want to make Active Record to use Helsinki timezone, instead, in the future, we might want to make Active Record aware of user's timezone.

### DIFF
--- a/app/services/marketplace_service/inbox.rb
+++ b/app/services/marketplace_service/inbox.rb
@@ -60,7 +60,7 @@ module MarketplaceService
 
       inbox_row_common_spec = [
         [:conversation_id, :fixnum, :mandatory],
-        [:last_activity_at, :sql_to_time, :mandatory],
+        [:last_activity_at, :utc_str_to_time, :mandatory],
         [:current_is_starter, :mandatory, transform_with: @tiny_int_to_bool],
         [:current_id, :string, :mandatory],
         [:other_id, :string, :mandatory],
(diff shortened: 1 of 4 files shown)


usmanarif:
I am getting "ArgumentError: argument out of range" error for

utc_str_to_time(str)
present in app/utils/time_utils.rb file

dsumeet:
Hi, 

I was getting the same error in app/utils/time_util.rb file

I am using a fresh copy of sharetribe from github . 

After analysis i found that the issue is related to date type conversion so I had to change the following code segment in line 291 services/marketplace_service/inbox.rb

current 
GREATEST(COALESCE(transactions.last_transition_at, 0),
              COALESCE(conversations.last_message_at, 0)) AS last_activity_at,

corrected
cast(GREATEST(COALESCE(transactions.last_transition_at, 0),
              COALESCE(conversations.last_message_at, 0)) as datetime)   AS last_activity_at,

the correction has fixed my Argument_out_range exception , the last_activity_at was showing 2015-0 instead of a complete date.

Can you please put my or a final fix in main release.

PR #848, opened 2014-12-03
create payment_settings table, model, store

ovan:


ovan:
@gonxi Hi, our PayPal code right now is only for the hosted platform. It resides here in the open repository but it cannot be used as is by OS installations. We try to cover this in our readme to: https://github.com/sharetribe/sharetribe#payments

app/controllers/admin/paypal_preferences_controller.rb
```diff
@@ -38,10 +38,16 @@ def index
     currency = @current_community.default_currency
     minimum_commission = paypal_minimum_commissions_api.get(currency)
 
+    tx_settings =
+      Maybe(tx_settings_api.get(community_id: @current_community.id, payment_gateway: :paypal, payment_process: :preauthorize))
```
sktoiva: This same code is used in paypal_helper. Should the paypal helper contain a method like
```ruby
def get_settings
  Maybe(TxApi.settings.get_active(community_id: community_id))
      .select { |result| result[:success] }
      .map { |result| result[:data] }
end
```

PR #900, opened 2015-01-07
Remove plans from community table

Gnito:
and change expires_at datetime for trials


app/services/marketplace_service/api/marketplaces.rb
```diff
@@ -18,7 +18,9 @@ def create(params)
       Helper.create_community_customization!(community, marketplace_name, locale)
       t = Helper.create_transaction_type!(community, p[:marketplace_type])
       Helper.create_category!("Default", community, locale, t.id)
-      Helper.create_community_plan!(community);
+
+      plan_level = p[:plan_level].or_else(CommunityModel::FREE_PLAN)
```
ovan: The home of this enum, `CommunityModel::FREE_PLAN` etc, would now make more sense to be under CommunityPlan ?

Gnito: true.

PR #906, opened 2015-01-13
I18n warning

bingxie:
1. Get rid of the warning: [deprecated] I18n.enforce_available_locales will default to true in the future. If you really want to skip validation of your locale you can set I18n.enforce_available_locales = false to avoid this message.
2. Fix typo

This is my first pull request. I also want to try the process.


sktoiva:
This is good!

bingxie:
Great, thank you.

PR #910, opened 2015-01-13
Don't trust receiver_id in pp ec payment response

ovan:
PayPal backend has a bug where they sometimes send an incorrect (unencrypted) merchant account id back in DoExpressCheckoutPayment response. This leads to payment not being linked to correct receiver paypal account and subsequently in the payment process not to be found. Instead, we use merchant account's payer_id that we know anyway from the context where the API call is triggered.


sktoiva:
_b

PR #913, opened 2015-01-14
Divided one long delay job (CommunityMembersEmailSentJob):

Gnito:
1. create a delay job where you loop through community members
2. create a send message delay job per community member

### DIFF
--- a/app/jobs/create_member_email_batch_job.rb
+++ b/app/jobs/create_member_email_batch_job.rb
@@ -11,12 +11,13 @@ class CreateMemberEmailBatchJob < Struct.new(:sender_id, :community_id, :subject
   end
 
   def perform
-    sender = Person.find(sender_id)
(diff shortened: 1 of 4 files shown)


app/jobs/create_member_email_batch_job.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def perform
+    sender = Person.find(sender_id)
+    current_community = Community.find(community_id)
+
+    Delayed::Job.transaction do
+      current_community.members.each do |recipient|
+        Delayed::Job.enqueue(CommunityMemberEmailSentJob.new(sender.id, recipient.id, community_id, subject, content, locale))
+      end
+    end
```
Gnito: I'm quite confused about this.
transaction creates only one connection to the db, but inside we need to "actions": insert row and enqueue

PR #915, opened 2015-01-15
Fix yens

ovan:


app/controllers/homepage_controller.rb
```diff
@@ -92,14 +92,16 @@ def find_listings(params, listings_per_page)
...
-    filter_params[:price_cents] = if (params[:price_min] && params[:price_max])
-      min = params[:price_min].to_i * 100
-      max = params[:price_max].to_i * 100
-
-      # Search only if range is not from min boundary to max boundary
-      if min != @current_community.price_filter_min || max != @current_community.price_filter_max
-        filter_params[:price_cents] = (min..max)
-      end
+    filter_params[:price_cents] =
+      if (params[:price_min] && params[:price_max])
```
sktoiva: Would it be better to get rid of this style and use extract method here?

app/utils/money_util.rb
```diff
@@ -24,4 +23,9 @@ def to_dot_separated_str(m)
     [units, subs.rjust(2, "0")].join(".")
   end
 
+  # Get only full units. No rounding.
```
sktoiva: It's floor

app/services/braintree_sale_service.rb
```diff
@@ -5,12 +5,14 @@
...
   def initialize(payment, payment_params)
+    subunit_to_unit = Money::Currency.new(payment.currency).subunit_to_unit
+
     @payment = payment
     @community = payment.community
     @payer = payment.payer
     @recipient = payment.recipient
-    @amount = payment.sum_cents.to_f / 100  # Braintree want's whole dollars
-    @service_fee = payment.total_commission.cents.to_f / 100
+    @amount = payment.sum_cents.to_f / subunit_to_unit  # Braintree want's whole dollars
```
sktoiva: It's probably not whole dollars but dollars instead of cents.

app/controllers/homepage_controller.rb
```diff
@@ -92,14 +92,16 @@ def find_listings(params, listings_per_page)
...
+        min = MoneyUtil.parse_str_to_money(params[:price_min], @current_community.default_currency).cents
+        max = MoneyUtil.parse_str_to_money(params[:price_max], @current_community.default_currency).cents
+
+        if ((@current_community.price_filter_min..@current_community.price_filter_max) != (min..max))
+          (min..max)
+        else
+          nil
+        end
     end
```
sktoiva: Indenting is wrong

PR #917, opened 2015-01-15
spinner after paying-with-PayPal redirect

Gnito:


config/locales/en.yml
```diff
@@ -1509,6 +1509,8 @@ en:
     transaction:
       commission_payment_name: "Commission payment for %{listing_title}"
       commission_payment_description: "Marketplace %{service_name} took this commission from transaction regarding %{listing_title}"
+    wait_while_loading: "Please wait."
+    chatting_with_paypal: "We need to chat with Paypal."
```
ovan: The official way to write it is PayPal.

sktoiva: This generates quite an uncertain feeling - should it be just: We are chatting with PayPal.

Otherwise, I like it though ;)

config/locales/fi.yml
```diff
@@ -1509,6 +1509,8 @@ fi:
     transaction:
       commission_payment_name: ~
       commission_payment_description: ~
+    wait_while_loading: "Anteeksi, odota hetki."
+    chatting_with_paypal: "Joudumme juttelemaan PayPalin kanssa."
```
sktoiva: Same here, Juttelemme PayPalin kanssa.

config/locales/fi.yml
```diff
@@ -1509,6 +1509,8 @@ fi:
     transaction:
       commission_payment_name: ~
       commission_payment_description: ~
+    wait_while_loading: "Anteeksi, odota hetki."
```
sktoiva: Probably no need to be sorry?

PR #918, opened 2015-01-16
Pp info notification

sktoiva:


app/view_utils/paypal_helper.rb
```diff
@@ -20,4 +20,8 @@ def user_and_community_ready_for_payments?(user, community)
     PaypalHelper.personal_account_prepared?(user, community) &&
       PaypalHelper.community_ready_for_payments?(community)
   end
+
+  def missing_payment_info?(user, community)
```
ovan: Naming should make it clear that the result is dependent on user having open listings, not just missing payment info.

app/view_utils/paypal_helper.rb
```diff
@@ -20,4 +20,8 @@ def user_and_community_ready_for_payments?(user, community)
...
+    community.paypal_enabled && user.listings.present? && !user_and_community_ready_for_payments?(user, community)
```
ovan: Should this only check for open listings, not closed?

ovan:
Can you rebase on top of paypal-payment-settings? This will only be used after we get that merged and it changes how payment settings for community are stored in case of PayPal.

app/view_utils/paypal_helper.rb
sktoiva: Well, maybe. Don't know if it makes that much difference - but might give a better impression when testing.

sktoiva:
Checked, a really good move to right direction.

PR #920, opened 2015-01-19
Use subunit_to_unit conversion on price validation

ovan:


app/assets/javascripts/kassi.js
```diff
@@ -680,31 +685,39 @@ function initialize_listing_view(locale) {
...
+  form_type,
+  form_id,
+  minimum_price,
+  subunit_to_unit,
+  minimum_price_message) {
+
+  auto_resize_text_areas("text_area");
+  style_action_selectors();
+
+  if (commission_percentage != null) {
```
sktoiva: Should these different branches be extracted into own functions?

ovan: Maybe, but modification to this function is just collateral damage from adding subunit parsing to listing create/edit so I don't feel like putting effort into refactoring this. Also, it's part of almost-deprecated postpay flow.

PR #923, opened 2015-01-20
Fix bug: Show 410 Gone when trying to access profile page of a deleted user

rap1ds:
- Remove filter `person_belongs_to_current_community`. Instead, fetch the `@person` where needed and make sure it belongs to community by manually calling `ensure_person_belongs_to_current_community!` method.
- Fix one review comment from Olli

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -222,10 +222,6 @@ class ApplicationController < ActionController::Base
     end
   end
 
-  def ensure_person_belongs_to_current_community!(person)
-    raise ActiveRecord::RecordNotFound.new('Not Found') unless @person.communities.include?(@current_community)
-  end
-
   def fetch_community_admin_status
     @is_current_community_admin = @current_user && @current_user.has_admin_rights_in?(@current_community)
   end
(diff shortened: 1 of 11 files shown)


app/controllers/application_controller.rb
```diff
@@ -222,8 +222,7 @@ def set_default_url_for_mailer
     end
   end
 
-  def person_belongs_to_current_community
-    @person = Person.find(params[:person_id] || params[:id])
+  def ensure_person_belongs_to_current_community!(person)
```
ovan: Since this is no longer a filter I don't see much reason to keep it in ApplicationController. Why not separate view util for person-related controller utility functions similarly as we have e.g. PaypalHelper?

app/services/transaction_service/transaction.rb
```diff
@@ -17,9 +17,7 @@ def query(transaction_id)
   end
 
   def has_unfinished_transactions(person_id)
-    finished_states = ["free", "rejected", "confirmed", "canceled", "errored"]
-                      .map { |state| "'#{state}'"}
-                      .join(",")
+    finished_states = "'free', 'rejected', 'confirmed', 'canceled', 'errored'"
```
ovan: :+1:

ovan:
Looks good, travis error seems weird though?

PR #924, opened 2015-01-20
Remove 'email' column from 'people' table

rap1ds:
- email column has not been in use for a while
- all emails have been migrated to separate emails table

- [ ] Fix new user form


rap1ds:
Tests are failing. I have to fix those.

rap1ds:
This is too big task to do now. The `Person` object is binded to the view form and highly depends on the email column on db.

PR #925, opened 2015-01-20
Remove unused people listing page

rap1ds:
People listing page hasn't been in use for a long long time...


ovan:
Shoud this also remove the controller action?

ovan:
ok

PR #926, opened 2015-01-20
Refactor atom api

sktoiva:


Gnito:
looks good

app/controllers/atom/listings_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+               :optional_category => @category_label,
+               :community_name => @current_community.name_with_separator(I18n.locale),
+               :listing_type => listing_type_label)
+    @updated = @listings.first.present? ? @listings.first.updated_at : Time.now
+  end
+
+  private
+
+  def set_pagination
+    @page = params["page"] || 1
```
rap1ds: It seems that `@page` and `@per_page` are used only in `index` method, not in the view. So having these as instance variables adds unnecessary complexity.

If you want to keep the `set_pagination` as a separate method, that's ok, but instead of settings the instance variables, you could just return the values.

app/controllers/atom/listings_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  before_filter :set_pagination
+
+  respond_to :atom
+  layout false
+
+  # Render atom feed
+  def index
+    @listings = Listing.find_with(params, @current_user, @current_community, @per_page, @page)
+
+    @total_pages = @listings.total_pages
```
rap1ds: Seems that this is not in use.

app/controllers/atom/listings_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      listing_type_label = t("listings.index.#{params['share_type']+"s"}")
+    else
+      listing_type_label = t("listings.index.listings")
+    end
+
+    @title = t("listings.index.feed_title",
+               :optional_category => @category_label,
+               :community_name => @current_community.name_with_separator(I18n.locale),
+               :listing_type => listing_type_label)
+    @updated = @listings.first.present? ? @listings.first.updated_at : Time.now
```
rap1ds: Is it possible to pass these @variables as locals with the `render` method?

rap1ds:
Added a couple comments but all in all very good clean up +1

PR #930, opened 2015-01-21
Added knowledge base links to some admin pages.

Kusti:


Kusti:
Note: to see the knowledge base articles, remember to switch the "display_knowledge_base_articles" to true in your config.yml (default is false).

app/views/admin/communities/analytics.haml
```diff
@@ -10,7 +10,11 @@
       %h3= t(".google_analytics_key")
     .row
       .col-6
-        = render :partial => "layouts/info_text", :locals => { :text => t(".google_analytics_key_info_text")}
+        - if APP_CONFIG.display_knowledge_base_articles
+          - instructions_link = link_to(t("admin.communities.analytics.google_analytics_instructions_link_text"), "#{APP_CONFIG.knowledge_base_url}/articles/412735-how-to-configure-google-analytics" )
+          = render :partial => "layouts/info_text", :locals => { :text => t(".google_analytics_key_info_text_with_instructions", :instructions_link => instructions_link).html_safe }
+        - else
+          = render :partial => "layouts/info_text", :locals => { :text => t(".google_analytics_key_info_text", :instructions_link => instructions_link).html_safe }
```
sktoiva: Where is instructions_link defined?

app/views/admin/communities/analytics.haml
```diff
@@ -10,7 +10,11 @@
       %h3= t(".google_analytics_key")
     .row
       .col-6
-        = render :partial => "layouts/info_text", :locals => { :text => t(".google_analytics_key_info_text")}
+        - if APP_CONFIG.display_knowledge_base_articles
```
sktoiva: We'd like to keep the logic of views in minimum, so instead of this, construct the instruction link and link text in controller and pass as locals. Then this whole if - else checking downgrades here to:

```ruby
= render :partial => "layouts/info_text", :locals { text: text, instructions_link: instructions_link }
```

app/views/admin/communities/social_media.haml
```diff
@@ -13,7 +13,11 @@
       %h3= t(".twitter_handle")
     .row
       .col-6
-        = render :partial => "layouts/info_text", :locals => { :text => t(".twitter_handle_info_text") }
+        - if APP_CONFIG.display_knowledge_base_articles
```
sktoiva: Same here as in previous comment about logic.

app/views/admin/paypal_preferences/index.haml
```diff
@@ -17,6 +17,10 @@
   .row
     = t("admin.paypal_accounts.integration_info_text").html_safe
 
+  - if APP_CONFIG.display_knowledge_base_articles
```
sktoiva: And here.

app/views/admin/communities/analytics.haml
ovan: Three lines above. But it should come from controller via locals instead.

app/views/admin/communities/analytics.haml
Kusti: Ahem, the "instructions_link" is a parameter to the translation of the text, so your code is not quite correct. But I'll do something and see what you think. :)

app/views/admin/communities/analytics.haml
sktoiva: Yes, but it's inside the the other if branch, or do I now miss something essential about variable scoping?

PR #931, opened 2015-01-21
Add :not_applicable in order to maintain backwards compatibility.

sktoiva:


ovan:
Good catch, ok to merge.

PR #932, opened 2015-01-22
BraintreeService::EscrowReleaseHelper.next_settlement_batch_time not always returns next batch fixed

### LINKED ISSUE #911: Always failed rspec test
I always have a failed rspec test on my local machine, I run the test with:
rake test:prepare
rspec spec

The failed test is:

Failures:

  1) BraintreeService::EscrowReleaseHelper#release_from_escrow when escrow status changed from "hold_pending" to "held" releases from escrow later
     Failure/Error: successes.should == 1
       expected: 1
            got: 0 (using ==)
     # ./spec/services/braintree_service/escrow_release_helper_spec.rb:37:in `block (4 levels) in <top (required)>'

Seems like the failed reason is Delayed::Worker.new.work_off returns [0, 0] not [1, 0], do I need to config anything for the delayed job?

heydiplo:
This PR fixes the issue when next_settlement_batch_time returns the batch time which is more than 24 hours ahead.

* Let’s say we’re in GMT+4 and it's March 11, 20:00 GMT+0.
* Time.now returns time in current timezone, it is Mar 12, 00:00 GMT+4.
* settlement_batch_time_per_date(Time.now) will take date with current timezone applied, which is 12th.
* And will return March 12, 17:00 CST. which is Mar 12 23:00 GMT+0.  And it's not the next batch, it's 27 hours ahead.

In fact, it happens every time current hour is less or equal to current timezone offset plus 6 hours (e.g. 0:00–10:00 for GMT+4). It also means the tests are always failing at some period of the day. The futher to the east you are, the bigger this period is.

Fixes #911.

### DIFF
--- a/app/services/braintree_service/escrow_release_helper.rb
+++ b/app/services/braintree_service/escrow_release_helper.rb
@@ -2,6 +2,9 @@ module BraintreeService
   module EscrowReleaseHelper
     include Braintree::Transaction::EscrowStatus
 
+    BATCH_TIMEZONE = "Central Time (US & Canada)"
+    BATCH_HOUR = 17
+
     module_function
 
     # Get back the next escrow release time
@@ -40,18 +43,11 @@ module BraintreeService
       self.delay(:run_at => next_escrow_release_time, :priority => 6).release_from_escrow_by_community_id(community_id, transaction_id)
     end
 
-    # Give a date and get back time of the given date when batch is run
-    def settlement_batch_time_per_date(date)
-      offset = ActiveSupport::TimeZone.new("Central Time (US & Canada)").formatted_offset
-      Time.new(date.year, date.month, date.day, 17, 0, 0, offset)
-    end
-
     # Give a date and get back time of next batch time
     def next_settlement_batch_time(now=Time.now)
-      todays_batch = settlement_batch_time_per_date(now)
-      tomorrows_batch = settlement_batch_time_per_date(now.tomorrow)
-
-      now < todays_batch ? todays_batch : tomorrows_batch
+      next_batch_time = now.in_time_zone(BATCH_TIMEZONE).change(hour: BATCH_HOUR, min: 0, sec: 0)
+      next_batch_time += 1.day if next_batch_time < now
+      next_batch_time
     end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/services/braintree_service/escrow_release_helper_spec.rb
```diff
@@ -50,17 +50,22 @@
...
   describe "#next_escrow_release_time" do
+
+    def central_time(*args)
+      ActiveSupport::TimeZone.new("Central Time (US & Canada)").local(*args)
+    end
+
     context 'when todays batch has not been run yet' do
       it 'returns time of todays batch and buffer' do
-        next_batch = BraintreeService::EscrowReleaseHelper.next_escrow_release_time(Time.new(2014, 3, 11, 22, 0, 0, 0), 2)
-        next_batch.utc.should be_eql(Time.new(2014, 3, 12, 1, 0, 0, 0))
```
heydiplo: Central Time is GMT-6 without daytime saving and GMT-5 with it. Thus, this line will always fail in summertime. It's better to compare it directly to desired 19:00 in Central Time. Also, 22:00 UTC should resolve to the next day in winter (because 22-5 is 17 and current day's batch is already passed) and to the current day in summer (22-6 is 16, the batch is in an hour). This is why I changed it to 21:00, which should always resolve to current day's batch.

app/services/braintree_service/escrow_release_helper.rb
```diff
@@ -40,18 +43,11 @@ def release_from_escrow_after_next_batch(community_id, transaction_id)
...
-
     # Give a date and get back time of next batch time
     def next_settlement_batch_time(now=Time.now)
-      todays_batch = settlement_batch_time_per_date(now)
-      tomorrows_batch = settlement_batch_time_per_date(now.tomorrow)
-
-      now < todays_batch ? todays_batch : tomorrows_batch
+      now = now.in_time_zone(BATCH_TIMEZONE)
+      now += 1.day if now.hour >= BATCH_HOUR # we've missed today's batch
+      now.change({ hour: BATCH_HOUR, min: 0, sec: 0 })
```
bingxie: According to the API (if only the hour is passed, then minute, sec, and usec is set to 0), so we don't need to pass the min and sec.

heydiplo: I rewrote this piece a little. Decided to leave min and sec though to make it explicit for future generations.

rap1ds:
Thanks a lot! This is very nice fix! The bug must have been there for quite a while without no one noticing it. Very good that it's now fixed!

PR #935, opened 2015-01-22
Don't mention escrow if not using braintree

ovan:


app/controllers/admin/communities_controller.rb
```diff
@@ -221,4 +222,8 @@ def payment_settings_api
     TransactionService::API::Api.settings
   end
 
+  def escrow_payments?(community)
+    MarketplaceService::Community::Query.payment_type(community) == :braintree
```
sktoiva: Is it intentional that you pass the whole community model instead of just the id?

Community.find_by_id(@current_community) handles this well of course.

ovan: Nope, not intentional. Kinda creepy that it worked

sktoiva: Well, these find_by_id, and I think find as well (which is the underlying implementation of payment_type) default to either using passed parameter as id or call .id on passed object.

PR #936, opened 2015-01-22
Rewrite Paypal account logic to separate API and Store

rap1ds:


app/services/paypal_service/store/paypal_account.rb
```diff
@@ -0,0 +1,50 @@
+module PaypalService::Store::PaypalAccount
+  PaypalAccountModel = ::PaypalAccount
+  OrderPermissionModel = ::OrderPermission
+
+  PaypalAccount = EntityUtils.define_builder(
+    [:id, :fixnum], # todo should we return this or not?
```
ovan: No need hopefully

app/services/paypal_service/store/paypal_account.rb
```diff
@@ -0,0 +1,161 @@
...
+    account_model = PaypalAccountModel.create!(account)
+    permission_model = OrderPermissionModel.create!(permission.merge(paypal_account_id: account_model.id))
+
+    from_model(account_model)
+  end
+
+  def update(opts)
+    account_model = find_model(opts[:person_id], opts[:community_id])
+
+    raise "Can not find Paypal account for person_id #{opts[:person_id]} and community_id #{opts[:community_id]}" unless account_model
```
ovan: ArgumentError

app/services/paypal_service/store/paypal_account.rb
```diff
@@ -0,0 +1,161 @@
...
+
+  def delete_billing_agreement(opts)
+    account = find_model(opts[:person_id], opts[:community_id])
+    account.billing_agreement.destroy
+  end
+
+  def delete_billing_agreement_by_payer_and_agreement_id(payer_id, billing_agreement_id)
+    billing_agreement = find_billing_agreement_by_payer_and_agreement_id(payer_id, billing_agreement_id)
+    billing_agreement.destroy
+  end
```
ovan: `def delete_billing_agreement(payer_id:, billing_agreement_id:)`

app/services/paypal_service/api/accounts.rb
```diff
@@ -0,0 +1,225 @@
...
+
+    ## GET /accounts/:community_id(/:person_id?)
+
+    def get(community_id, person_id=nil)
+      Result::Success.new(PaypalAccountStore.get(person_id, community_id))
+    end
+
+    private
+
+    def with_permission_request_response(callback_url, &block)
```
ovan: Instead of per request specific helpers payments.rb has a pattern that uses one more general `with_success` helper to do the request. Upside is that the request type and data used is more apparent already at api implementing method. Could we do something similar here? Have a loko at request_wrapper.rb

spec/services/paypal_service/test_permissions.rb
```diff
@@ -0,0 +1,65 @@
...
+    def self.build(api_builder)
+      PaypalService::Permissions.new(nil, TestLogger.new, TestPermissionsActions.new.default_test_actions, api_builder)
+    end
+  end
+
+  class TestPermissionsActions
+    attr_reader :default_test_actions
+
+    def initialize
+      @fake_pal = FakePal.new
```
ovan: Is this used somewhere?

app/services/transaction_service/transaction.rb
```diff
@@ -51,15 +48,27 @@ def can_start_transaction_braintree(community_id:, author_id:)
...
 
   def can_start_transaction_paypal(community_id:, author_id:)
-    paypal_account = PaypalService::PaypalAccount::Query.personal_account(author_id, community_id)
-    admin_account = PaypalAccountQuery.admin_account(community_id)
-
-    PaypalAccountEntity.paypal_account_prepared?(paypal_account) &&
-      PaypalAccountEntity.order_permission_verified?(admin_account) &&
+    person_account_response = paypal_account_api.get(
+      community_id: community_id,
+      author_id: author_id
```
rap1ds: should be person_id

spec/services/paypal_service/test_permissions.rb
rap1ds: True, not used. I'll remove it.

app/controllers/paypal_accounts_controller.rb
```diff
@@ -32,13 +30,17 @@ def show
...
   end
 
   def new
-    paypal_account = PaypalAccountQuery.personal_account(@current_user.id, @current_community.id)
-    return redirect_to action: :show if PaypalAccountEntity.paypal_account_prepared?(paypal_account)
+    account_response = accounts_api.get(
+      community_id: @current_community.id,
+      person_id: @current_user.id
+    )
+    m_account = Maybe(account_response[:data])
```
ovan: Isnt's this the same as: `account_response.maybe`?

app/controllers/paypal_accounts_controller.rb
```diff
@@ -72,10 +74,14 @@ def new
   def create
     return redirect_to action: :new unless PaypalHelper.community_ready_for_payments?(@current_community)
 
-    paypal_account = PaypalAccountQuery.personal_account(@current_user.id, @current_community.id)
-    order_permission_verified = PaypalAccountEntity.order_permission_verified?(paypal_account)
+    account_response = accounts_api.get(
+      community_id: @current_community.id,
+      person_id: @current_user.id
+    )
+    m_account = Maybe(account_response[:data])
```
ovan: Same here, `account_response.maybe`?

app/services/paypal_service/api/accounts.rb
```diff
@@ -0,0 +1,228 @@
...
+    # The API implmenetation
+    #
+
+    ## POST /accounts/request
+
+    def request(body:)
+      # If a request for new account is made, delete old (not completely actived) accounts
+      # Future note: This may not be sufficient solution when we implement the option to
+      # switch PayPal account
+      delete(
```
ovan: I don't like API operations calling directly each other. There's unnecessary ceremony wrapping results etc. when we are already inside the module. I'd prefer using Store directly.

app/services/paypal_service/api/accounts.rb
```diff
@@ -0,0 +1,228 @@
...
+    # , success_url: "http://marketplace.com/success
+    # , cancel_url: "http://marketplace.com/cancel
+    # }
+    #
+
+    def billing_agreement_request(community_id:, person_id:, body:)
+      # If a request for new billing agreement is made, delete old (not completely actived) billing agreement
+      # Future note: This may not be sufficient solution when we implement the option to
+      # switch PayPal account
+      delete_billing_agreement(
```
ovan: Same comment here about calling other public API operations.

app/services/paypal_service/permissions_injector.rb
```diff
@@ -7,10 +7,14 @@ def paypal_permissions
...
 
     def build_paypal_permissions
-      PaypalService::Permissions.new(
-        build_endpoint(APP_CONFIG),
-        build_api_credentials(APP_CONFIG),
-        PaypalService::Logger.new)
+      config = DataTypes.create_config(
+        {
+          endpoint: build_endpoint(APP_CONFIG),
+          api_credentials: build_endpoint(APP_CONFIG)
```
ovan: Shouldn't it be build_api_credentials?

app/services/paypal_service/store/paypal_account.rb
```diff
@@ -0,0 +1,220 @@
...
+  # Rename map for PaypalAccount values that are stored to billing_agreement table
+  BILLING_AGREEMENT_MAP = {
+    billing_agreement_billing_agreement_id: :billing_agreement_id,
+    billing_agreement_paypal_username_to: :paypal_username_to,
+    billing_agreement_request_token: :request_token
+  }
+
+  module_function
+
+  def create(opts:)
```
ovan: The :opts keyword param feels a bit unnecessary since it's the only parameter and also generic name opts doesn't add anything to document the call site.

rap1ds: This is only for consistency.

In `update` method there are three params, the two search params (community_id and person_id) and then the third `opts`, which is the "body" of the method call. The create method follows the same pattern, however, when we are creating a new account, we do not have the search params, only the content `opts` for the new account.

PR #937, opened 2015-01-22
Don't show credit icons for free transaction types

ovan:


app/views/listings/_listing_actions.haml
```diff
@@ -69,7 +69,7 @@
         .content
           = @listing.transaction_type.action_button_label(I18n.locale)
 
-    - if payment_gateway == :paypal
+    - if payment_gateway == :paypal && @listing.transaction_type.status_after_reply != "free"
```
rap1ds: Actually, we don't want to show the icons for any transaction type that does not involve money. So I would compare `@listing.transaction_type.price_field == true`

rap1ds: It seems that `status_after_reply` takes price into account, so my previous comment is not true.

PR #948, opened 2015-01-26
Rely on javascript on start and end validations

sktoiva:


app/models/booking.rb
```diff
@@ -17,8 +17,6 @@ class Booking < ActiveRecord::Base
   attr_accessible :transaction_id, :end_on, :start_on
 
   validates :start_on, :end_on, presence: true
-  validates_date :start_on, on: :create, on_or_after: :today
-  validates_date :end_on, on_or_after: :start_on
```
ovan: I think this second validation still makes sense (although might be the controller already takes care of this and that's where it belongs).

PR #949, opened 2015-01-27
Add and_then and maybe to Result classes

rap1ds:


app/services/result.rb
```diff
@@ -9,6 +9,56 @@ def initialize(data = nil)
...
+    #   if account[:payer_id].nil?
+    #     Result::Error.new(:not_found)
+    #   else
+    #     payments_api.do_payment(payer: account[:payer_id])
+    #   end
+    # }
+    def and_then(&block)
+      result = block.call(data)
+      result.tap do |res|
+        raise "Block must return Result" unless (res.is_a?(Result::Success) || res.is_a?(Result::Error))
```
ovan: `raise ArgumentError.new(...`

PR #955, opened 2015-01-30
Remove private repository dependency

rap1ds:
Currently we have a private Github repo which contains:

- Mangopay key file (not needed anymore)
- ss-pika/ss-social webfonts

This pull request removes the mangopay key because it's not needed, and it moves webfonts to S3.

The good thing is that:

- Now we can link Heroku to Github and see exactly which commit was deployed and also we can see the code diff easily
- New developer does not need to setup the private repository access
- Simplifies the deployment

Bad:

- New developer has to copy webfonts to local machine manually

### DIFF
--- a/config/config.defaults.yml
+++ b/config/config.defaults.yml
@@ -182,7 +182,7 @@ default: &default_settings
 
   # If icon_pack is set to "ss-pika", set the location for the icon pack
   # This setting is irrelevant to Open Source developers, since ss icon pack
-  # does not include to the Open Source distribution
+  # is not included in the Open Source distribution
   ss_pika_location: "ss-pika"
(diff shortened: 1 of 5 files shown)


config/config.defaults.yml
```diff
@@ -180,6 +180,12 @@ default: &default_settings
   # Currently only othe option is "ss-pika" (a proprietary icon set)
   icon_pack: "font-awesome"
 
+  # If icon_pack is set to "ss-pika", set the location for the icon pack
+  # This setting is irrelevant to Open Source developers, since ss icon pack
+  # does not include to the Open Source distribution
```
ovan: is not included in the...

PR #957, opened 2015-02-02
Change the semantic of 'active' PayPal account

rap1ds:
Change the semantic of 'active' PayPal account

Old semantic:

- Personal account was marked active when both order permissions and billing agreement were verified
- Community account was marked active when order permissions were verified

New semantic:

- All accounts are marked active after order permissions are verified
- Being 'active' does not anymore mean that the account setup is fully completed

### DIFF
--- a/app/services/transaction_service/transaction.rb
+++ b/app/services/transaction_service/transaction.rb
@@ -48,23 +48,15 @@ module TransactionService::Transaction
   end
 
   def can_start_transaction_paypal(community_id:, author_id:)
-    person_account_response = paypal_account_api.get(
-      community_id: community_id,
-      person_id: author_id
-    )
-    personal_account_active = Maybe(person_account_response)[:data][:active]
-
-    admin_account_response = paypal_account_api.get(
-      community_id: community_id
-    )
-    admin_account_active = Maybe(admin_account_response)[:data][:active]
+    personal_account_verified = paypal_account_verified?(community_id: community_id, person_id: author_id)
+    community_account_verified = paypal_account_verified?(community_id: community_id)
 
     payment_settings_available =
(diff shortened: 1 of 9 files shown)


app/services/transaction_service/transaction.rb
```diff
@@ -48,22 +48,14 @@ def can_start_transaction_braintree(community_id:, author_id:)
...
-      community_id: community_id,
-      person_id: author_id
-    )
-    personal_account_active = Maybe(person_account_response)[:data][:active]
-
-    admin_account_response = paypal_account_api.get(
-      community_id: community_id
-    )
-    admin_account_active = Maybe(admin_account_response)[:data][:active]
+    personal_account_prepared = PaypalHelper.account_prepared_for_user?(author_id, community_id)
```
rap1ds: Err... I think this is not right. Transaction service doesn't know about PaypalHelper, which is a module in the "main" web service.

ovan: Hmmm... I'd like to get rid of some of the use of PaypalHelper around the code. Could the account object returned from the API have some enum field that indicates if the account is ready for action? The logic of what it means for an account to be prepared is logically owned by paypal service that manages the accounts.

ovan: So basically a new field that would mean what the active used to mean? verified, setup_complete, fully_connected, or something like that.

rap1ds: That's a good suggesting. For the sake of consistency, let's make it `state: one_of [:verified, :not_verified]`

Then another point: `active` is actually only used by the PayPal API to internally select which account to return, so I guess there's no reason why we would want to expose that. Anyway, it's always `true`.

PR #961, opened 2015-02-03
New PayPal commission model

ovan:


app/assets/javascripts/kassi.js
```diff
@@ -722,17 +722,28 @@ function initialize_accept_transaction_form(
   }
 }
 
-function updateSellerGetsValue(priceInputSelector, youWillGetSelector, currencySelector, communityCommissionPercentage, minCommission) {
-  $display = $(youWillGetSelector);
+function updateSellerGetsValue(priceInputSelector, displayTargetSelector, currencySelector, communityCommissionPercentage, minCommission, showReversed) {
```
sktoiva: Now when we are doing a bigger refactoring and new functionality related to this, should we move this to own js module? We had the goal of getting rid of one big kassi.js file...

app/mailers/transaction_mailer.rb
```diff
@@ -93,6 +93,7 @@ def braintree_new_payment(payment, community)
           payment_total: humanized_money_with_symbol(payment.total_sum),
           payment_service_fee: humanized_money_with_symbol(service_fee),
           payment_seller_gets: humanized_money_with_symbol(you_get),
+          paypal_gateway_fee: nil, # This is probably a sign saying we shouldn't be sharing a template?-)
```
sktoiva: Totally agree with the comment. Should we fix it now?

app/services/transaction_service/transaction.rb
```diff
@@ -48,19 +48,15 @@ def can_start_transaction_braintree(community_id:, author_id:)
   end
 
   def can_start_transaction_paypal(community_id:, author_id:)
-    personal_account_verified = paypal_account_verified?(community_id: community_id, person_id: author_id)
+    payment_settings = Maybe(PaymentSettingsStore.get_active(community_id: community_id))
+                       .select {|set| set[:payment_gateway] == :paypal && set[:commission_from_seller] && set[:minimum_price_cents]}
```
sktoiva: The contents of select-block could be extracted to method. Would be faster to read and better documented.

app/services/transaction_service/transaction.rb
```diff
@@ -48,19 +48,15 @@ def can_start_transaction_braintree(community_id:, author_id:)
...
-      .select {|set| set[:payment_gateway] == :paypal && set[:commission_from_seller] && set[:minimum_price_cents]}
+    payment_settings_available = payment_settings.map {|_| true }.or_else(false)
 
-    case [personal_account_verified, community_account_verified, payment_settings_available]
-    when matches([true, true, Some])
-      true
-    else
-      false
-    end
+    [personal_account_verified, community_account_verified, payment_settings_available].all?
```
sktoiva: :+1:  for this and previous line.

rap1ds: @sktoiva No love for the fancy `matches` method? :cry: :broken_heart:

sktoiva: Yeah, it's fancy and I have love for it, but not in simple cases like this :)

app/services/transaction_service/transaction.rb
```diff
@@ -369,27 +367,23 @@ def checkout_details(model)
...
+    # Ensure we never charge more than what the seller received after payment processing fee
+    [commission_total, payment_total - fee_total].min
   end
 
-  def paypal_account_verified?(community_id:, person_id:nil)
-    acc = paypal_accounts_api.get(
-      community_id: community_id,
-      person_id: person_id
-    ).maybe
+  def paypal_account_verified?(community_id:, person_id: nil, settings: Maybe(nil))
```
sktoiva: The exact same method is in paypal_helper, should this logic be somehow published through service api?

app/assets/javascripts/kassi.js
ovan: Normally I'd agree but knowing some of the plans we have for mid-term future with rethinking the UI I don't think it's worth the time and effort to clean up the existing JS since it's likely to be thrown out anyway.

app/services/transaction_service/transaction.rb
ovan: Conceptually it's wrong to reach inside transaction service in controllers to do a purely functional data validation. Equally wrong would be to call from a service implementation to PaypalHelper controller utility. This logic is part of a "PayPal service client", i.e. something that is implemented at the side that calls out to paypal service. To share it would mean to implement a client package and introduce it as a dependency to anyone who needs to call out to paypal service. But that's clearly overkill now so instead we have this "intentional copy & paste". One way to get around this in some scenarios is to introduce calculated enum fields to the service resource representation but that's always a trade-off because then you end up calculating and sharing the data to everyone whether or not they need it.

app/services/transaction_service/transaction.rb
rap1ds: 🙌

PR #962, opened 2015-02-04
fix limiting on atom format

sktoiva:
Fixes https://sharetribe.airbrake.io/projects/15448/groups/1347153929703886658

Without format: true, constraints on format do not work so that they would limit access only to the specified one.

### DIFF
--- a/config/locales/nb.yml
+++ b/config/locales/nb.yml
@@ -1502,6 +1502,7 @@ nb:
       billing_agreement_wrong_account: ~
       something_went_wrong: ~
       account_not_verified: ~
+      account_restricted: ~
   paypal:
     checkout_with_paypal: ~
     cancel_succesful: ~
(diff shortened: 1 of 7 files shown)


ovan:
Seems to be something relevant reported by cucumber:
```
 No route matches {:format=>nil, :controller=>"atom/listings", :locale=>:en} (ActionView::Template::Error)
```

Gemfile.lock
```diff
@@ -398,7 +398,7 @@ GEM
     ruby-ole (1.2.11.6)
     ruby-prof (0.14.2)
     ruby-progressbar (1.4.2)
-    rubyzip (1.1.6)
+    rubyzip (1.1.7)
```
rap1ds: What's this?

sktoiva: Dunno, dependency to selenium webdriver.

rap1ds: Why is it updated?

sktoiva: Check commits. I updated selenium webdriver.

rap1ds: Ok, but selenium webdriver is not updated, I guess? But the ffi is, which is probably something related.

PR #963, opened 2015-02-04
Switch Paypal account

rap1ds:


app/controllers/people_controller.rb
```diff
@@ -226,7 +226,7 @@ def destroy
       MarketplaceService::Listing::Command.delete_listings(@current_user.id)
 
       communities.each { |community_id|
-        PaypalService::API::Api.accounts_api.delete(community_id: @current_community.id, person_id: @current_user.id)
```
rap1ds: I don't understand why current master doesn't fail tests on Travis because of this...

rap1ds:
There's one thing I have in mind what to improve Account store:

We have set up implicit rules of combinations that MUST return a unique account. They are:

- Personal account: person_id, community_id, payer_id
- Active personal account: person_id, community_id, active: true
- Community account: person_id: nil, community_id, payer_id
- Active community account: person_id: nil, community_id, active: true

Would be nice to somehow make these rules explicit and enforce correct usage. There's a great possibility for a bug, if a developer forgets one of the fields and the database query results a non unique result, from which the first one is selected.

Improvement ideas I had in mind:

- Throw an exception, if database query results more than one account. (easy and fast solution)
- Create a datatype for each combination. AccountStore methods would take the datatype entity instead of a huge list of keyword arguments. Data types would have correct mandatory settings to enforce the correct usage. (robust, documents nicely, laborious)
- Do nothing. This whole thing is only inside the Paypal service.

Any ideas how to continue?

app/controllers/paypal_accounts_controller.rb
```diff
@@ -71,7 +73,13 @@ def new
     })
   end
 
-  def create
+  def ask_order_permission
+    return redirect_to action: :new unless PaypalHelper.community_ready_for_payments?(@current_community)
+
+    create_paypal_account
```
ovan: This might not be entirely about new code in this branch but I'm having hard time understanding the need to have separate create_paypal_account method as it stands. This is the only place it is called from and this action just transforms control of flow to a private method. Make sit hard to understand what are the possible outcomes of invoking this action. A better style would be to move the code that decides what to flash and render into this action and have a helper for fetching redirect_url from PayPal.

app/controllers/paypal_accounts_controller.rb
```diff
@@ -84,7 +92,7 @@ def create
     when Some(:verified)
       create_billing_agreement
     else
-      create_paypal_account
+      redirect_to action: ask_order_permission
```
ovan: Similarly here we have inconsistent abstraction level in the method making it hard to understand. The above branch delegates control and outcome to a helper method but else branch deals with sending back a redirect. Again, my suggestion would be to handle rendering / redirecting here and use helpers to deal with pp api if needed.

app/controllers/people_controller.rb
ovan: Interesting. This is wrong on multiple levels. We shouldn't be deleting accounts connected to other communities and also there's the bug of not using the iterated over collection contents for anything.

ovan: And missed the fact that this was people controller. So that comment of mine makes no sense.

app/services/paypal_service/api/accounts.rb
```diff
@@ -68,17 +63,45 @@ def create(community_id:, person_id: nil, order_permission_request_token:, body:
...
+                opts: opts
+              )
+
+              # Delete the 'new' account
+              PaypalAccountStore.delete(
+                community_id: community_id,
+                person_id: person_id,
+                order_permission_request_token: order_permission_request_token
+              )
+            end
```
ovan: This is a bit too long to be readable. Pull the if-part to a private helper, something like: `create_verified_account(opts, existing)`. That way you can also eliminate the opts variable and declare the data directly as call parameter. Also, there might be a bug in the else branch because it returns the result of delete operation, not create operation.

app/services/paypal_service/api/accounts.rb
```diff
@@ -68,17 +63,45 @@ def create(community_id:, person_id: nil, order_permission_request_token:, body:
...
+                person_id: person_id,
+                order_permission_request_token: order_permission_request_token,
+                opts: opts
+              )
+            else
+              # Update the 'existing' account
+              PaypalAccountStore.update(
+                community_id: community_id,
+                person_id: person_id,
                 payer_id: personal_data[:payer_id],
```
ovan: Isn't this attribute passed in as part of opts?

app/services/paypal_service/api/lookup.rb
```diff
@@ -21,15 +21,15 @@ def with_account(cid, pid, &block)
       end
     end
 
-    def with_accounts(cid, pid, &block)
+    def with_accounts(cid, pid, receiver_id, &block)
       admin_acc = AccountStore.get(community_id: cid)
```
ovan: Should this have active: true also?

app/services/paypal_service/api/accounts.rb
rap1ds: Yes, but it's passed as a separate argument because it is used for the search.

app/services/paypal_service/api/lookup.rb
rap1ds: That's right :) And a very good example about my comment above about the implicit uniqueness rules :)

app/services/paypal_service/store/paypal_account.rb
```diff
@@ -103,16 +127,23 @@ def delete_billing_agreement_by_payer_and_agreement_id(payer_id:, billing_agreem
...
 
-  def delete(person_id:nil, community_id:)
-    find_model(person_id: person_id, community_id: community_id).each { |account| account.destroy }
+  def delete(person_id:nil, community_id:, order_permission_request_token: :all)
+    find_model(person_id: person_id, community_id: community_id, order_permission_request_token: order_permission_request_token).each { |account| account.destroy }
   end
 
-  def get(person_id:nil, community_id:)
-    from_model(find_model(person_id: person_id, community_id: community_id))
+  def get(person_id:nil, community_id:, active: :all, payer_id: :all)
```
ovan: This method signature is starting to feel like too many different use cases crammed into one shape. What if we'd have:

```ruby
def get(person_id:, community_id: nil, payer_id:)
def get_active(person_id:, community_id: nil)
```

So either you are looking for a specific account with all three parameters and don't care if it's active or not (inside transaction process) or then you just need the account to use for a person / community to start a transaction or to show in UI. Would that cover it?

ovan: Oh and then there's the special case, `get_personal_account_by_payer_id`

PR #965, opened 2015-02-04
Select accepted admins only.

Gnito:
Fixing a bug with banned admins.
Urgent misc task: https://trello.com/c/LMYZxXnu/1321-banned-admin-receive-some-email-admins

### DIFF
--- a/app/controllers/admin/community_memberships_controller.rb
+++ b/app/controllers/admin/community_memberships_controller.rb
@@ -12,6 +12,12 @@ class Admin::CommunityMembershipsController < ApplicationController
 
   def ban
     membership = CommunityMembership.find_by_id(params[:id])
+
+    if membership.person == @current_user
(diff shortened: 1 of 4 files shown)


app/controllers/admin/community_memberships_controller.rb
```diff
@@ -13,6 +13,7 @@ def index
   def ban
     membership = CommunityMembership.find_by_id(params[:id])
     membership.update_attributes(:status => "banned")
+    membership.update_attributes(:admin => 0) if membership.admin == 1
```
rap1ds: We should move to the new syntax hash `admin: 0`, but for the sake of consistency, this is ok here.

rap1ds:
The code looks good :+1: 

I don't fully understand the problem behind this. Why would anyone want to ban an admin user?

Another way to tackle this issue would be to set a limitation that admins can not be banned. If user want to ban admin user, he should first remove the admin rights and then ban.

PR #966, opened 2015-02-04
Add :receiver_id to paypal_tokens

rap1ds:


Gnito:
Didn't test this one, but looks pretty clear. :+1:
So, token contains both ids to seller: our person_id and now also paypal's "payer_id" renamed as receiver_id
(a bit strange this PayPal's name for the field)

app/models/paypal_account.rb
```diff
@@ -19,7 +19,7 @@
 #
 
 class PaypalAccount < ActiveRecord::Base
-  attr_accessible :email, :payer_id, :person_id, :community_id, :active
+  attr_accessible :email, :payer_id, :person_id, :community_id, :active, :receiver_id
```
rap1ds: This should be removed

PR #969, opened 2015-02-05
Refactor ListingsController and Atom::ListingsController

sktoiva:
Remove unnecessary routes and make application respond correctly to different formats.


config/routes.rb
```diff
@@ -95,10 +95,6 @@
     match '/:person_id/settings/payments/paypal_account/show' => 'paypal_accounts#show', :as => :show_paypal_account_settings_payment
     match '/:person_id/settings/payments/paypal_account/create' => 'paypal_accounts#create', :as => :create_paypal_account_settings_payment
 
-    scope module: "atom", constraints: { format: "atom" } do
```
rap1ds: Should we do a redirect, so that URLs pointing to this route will not fail. My assumption is that there are a lot of users using exactly this route, since this is the "new" route.

rap1ds:
One comment, otherwise looks ok :+1:

config/routes.rb
sktoiva: This generates route: ```<locale>/listings.atom?<params>``` which is duplicate route with what ```resources :listings``` creates. Moreover, the first one does not limit the format to "atom" because of how rails works. 

This fixes some requests to urls ```<locale>/listings``` that have generated now a TemplateNotFoundError.

sktoiva: Basically, this doesn't change any of our routes.

PR #970, opened 2015-02-05
Locale was not set when calling fetch_community filter

Gnito:
application_controller:
1) *fetch_community* filter is called before filter *set_locale*
2) *fetch_community* calls *fetch_community_by_strategy*
3) *fetch_community_by_strategy* calls *service_name* 
  [ ApplicationHelper.store_community_service_name_to_thread(service_name) ]
=> *service_name* uses **I18n.locale** which is not set from url! i.e. it's default.


rap1ds:
Maybe a better would be to move the service name storing to the thread to the `set_locale` filter.

rap1ds:
:+1:

PR #973, opened 2015-02-05
[Add merchant_id: 2/3] Migrate merchant_id to paypal_payments from transactions table

rap1ds:
- Data migration

This PR should be merged to master and deployed to production only after PR #972

### DIFF
--- a/db/migrate/20150205155519_populate_merchant_id_to_paypal_payments.rb
+++ b/db/migrate/20150205155519_populate_merchant_id_to_paypal_payments.rb
@@ -37,9 +37,12 @@ class PopulateMerchantIdToPaypalPayments < ActiveRecord::Migration
       SET paypal_payments.merchant_id = COALESCE(listings.author_id, accounts.person_id)
 
       WHERE
+        # pick author_id
         (listings.author_id IS NOT NULL AND accounts.person_id IS NULL) OR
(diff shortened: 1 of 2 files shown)


db/migrate/20150205155519_populate_merchant_id_to_paypal_payments.rb
```diff
@@ -0,0 +1,50 @@
...
+      # Join by matching `payer_id` and `community_id`
+      ON (paypal_payments.receiver_id = accounts.payer_id AND paypal_payments.community_id = accounts.community_id)
+
+      # Update `merchant_id` by selecting the first non-nil of author_id and person_id
+      SET paypal_payments.merchant_id = COALESCE(listings.author_id, accounts.person_id)
+
+      WHERE
+        (listings.author_id IS NOT NULL AND accounts.person_id IS NULL) OR
+        (listings.author_id IS NULL AND accounts.person_id IS NOT NULL) OR
+        (listings.author_id = accounts.person_id)
```
ovan: This allows nil ids if both are nil.

rap1ds: Good catch!

rap1ds:
Add `null: false`

PR #975, opened 2015-02-06
Remove community name

rap1ds:
Community name is not used anymore. We use always the translated community name from community customizations.

- [x] Before deploy, make sure that all communities have translations for the community name

### DIFF
--- a/app/models/community.rb
+++ b/app/models/community.rb
@@ -263,7 +263,13 @@ class Community < ActiveRecord::Base
   attr_accessor :terms
 
   def name(locale)
-    community_customizations.where(locale: locale).first.name
+    customization = community_customizations.where(locale: locale).first
+
+    if customization
+      customization.name
+    else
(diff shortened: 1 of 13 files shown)


rap1ds:
-[ ] Before deploy, make sure that all communities have translations for the community name

app/models/community.rb
```diff
@@ -264,18 +262,8 @@ class Community < ActiveRecord::Base
...
-      else
-        community_customizations.find_by_locale(locales.first).name
-      end
-    else
-      # TODO: this is not required any more when we remove "name" column,
-      # from community, this should be removed after that.
-      read_attribute(:name)
-    end
+  def name(locale)
+    community_customizations.where(locale: locale).first.name
```
ovan: In case of missing locale this fails with null pointer. Should decide if it's an error to call with non-existing locale and throw an ArgumentError with good error message or then decide this is acceptable and define a default value.

PR #982, opened 2015-02-11
Paypal error handling

rap1ds:
- Mark commission payment status to `errored` if an error occures.
- Improve error logging by adding community and transaction ids

### DIFF
--- a/app/services/paypal_service/api/billing_agreements.rb
+++ b/app/services/paypal_service/api/billing_agreements.rb
@@ -95,7 +95,8 @@ module PaypalService::API
           }),
         error_policy: {
           codes_to_retry: ["10001", "x-timeout", "x-servererror"],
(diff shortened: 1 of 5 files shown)


app/services/paypal_service/api/billing_agreements.rb
```diff
@@ -114,5 +115,22 @@ def do_charge_commission(community_id, info, m_acc, admin_acc, payment)
       end
     end
 
+    def error_payment(cid, txid, payment)
```
ovan: Needs better name, mark_payment_errored, record_payment_error, or something. Now you can get the impression that the job of this method is to cause a payment error but actually is just records an error already happened.

ovan:
Check the travis failure, rspec failed.

ovan:
:+1:

PR #984, opened 2015-02-12
Show missing payment warning only for listings with price

rap1ds:
- If user has a listing with transaction type "Give", there's no need for PayPal account, thus we should not show the warning about missing account


ovan:
ok, makes sense.

PR #985, opened 2015-02-12
Clean up the logic how we choose the path for new transaction page

rap1ds:
- Also, fixes potential bug: If preauthorize is true, but price field is false, do not go to PayPal payment page

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -108,26 +108,19 @@ class ListingsController < ApplicationController
       [nil, nil]
     end
 
-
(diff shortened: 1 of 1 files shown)


ovan:
Yeah, I like the change. Let's merge it. I think the logic is overly complex because of the way how the source data for the decision is spread to to many different places. Also naming the of the fields in transaction_type and the effect them having a value has is not exactly obvious. We'll be likely to revisit this with quantity / booking options implementation. Anyway, good stuff, makes the logic the way it exists today clear.

sktoiva:
:+1: 
This actually caused already confusion, when price field was not present but preauthorize was - and this prompted user to contact author.

PR #986, opened 2015-02-12
Shipping fields

sktoiva:


app/models/listing.rb
```diff
@@ -406,4 +409,10 @@ def answer_for(custom_field)
   def payment_required_at?(community)
     transaction_type.price_field? && community.payments_in_use?
   end
+
+  def currency=(currency)
+    raise ArgumentError.new "Invalid currency. Tried to change #{self[:currency]} to #{currency}." if self[:currency] && self[:currency] != currency
```
rap1ds: What does this do? Did I understand right, it throws when ever someone is trying to change the currency?

sktoiva: Yes. It does just that.

db/migrate/20150212124533_add_shipping_enabled_to_community.rb
```diff
@@ -0,0 +1,5 @@
+class AddShippingEnabledToCommunity < ActiveRecord::Migration
+  def change
+    add_column :communities, :shipping_enabled, :integer, default: 0
```
rap1ds: Is this meant to be integer, or should it be boolean?

app/models/listing.rb
rap1ds: Could you elaborate why is this needed?

db/migrate/20150212124533_add_shipping_enabled_to_community.rb
sktoiva: Yeah, should be boolean.

ovan: I'm not sure if it's good idea to put this into community table even temporarily. I spent considerable amount of time moving the paypal enabled out of community table after the first release when the check had been embedded already into a lot of places in the code base. What about marketplace_settings or listing_settings or selling_settings or something like that, per community table?

db/migrate/20150212125111_add_shipping_fields_to_listing.rb
```diff
@@ -0,0 +1,8 @@
+class AddShippingFieldsToListing < ActiveRecord::Migration
+  def change
+    add_column :listings, :require_shipping_address, :boolean, default: false
+    add_column :listings, :pickup_enabled, :boolean, default: true
```
Gnito: Is true good for default (sounds like it, but might need some discussion)

Gnito:
Update export script

Gnito:
Those are the only things that came to mind.
so, :+1:

PR #992, opened 2015-02-13
2/2 Redirect to current locale

rap1ds:
Merge only after #991


rap1ds:
This is pending. I will maybe (or maybe not) finish this later.

rap1ds:
I don't have time to finish this, so I will close the pull request but leave the branch here.

PR #993, opened 2015-02-13
Show filters

Gnito:


rap1ds:
:+1:

app/controllers/homepage_controller.rb
```diff
@@ -27,8 +27,9 @@ def index
     # only 1 sub share type, as that would make the listing type menu visible and it would look bit silly
     @transaction_type_menu_enabled = @transaction_types.size > 1
     @show_categories = @current_community.categories.size > 1
+    @filters_enabled = @current_community.custom_fields.size > 0 || @current_community.show_price_filter
```
rap1ds: This variable is only used once and it is not meant to be used in the views, so it shouldn't be instance variable. Remove the `@`.

Gnito: Yeah, that's true.

Gnito:
joins -fail :/

PR #996, opened 2015-02-16
Fix listingpost localebug

ovan:


app/views/listings/form/_price.haml
```diff
@@ -33,7 +33,7 @@
                 "#fee_display",
                 "#listing_currency",
                 #{commission_from_seller},
-                #{minimum_commission},
+                #{minimum_commission.format(decimal_mark: ".", symbol: false)},
```
rap1ds: This is not right for all languages. Finnish for example uses `,` as a decimal mark.

sktoiva:
:+1:

app/views/listings/form/_price.haml
ovan: Yep, that's the point. We are formatting number for JS to consume and localization messes things up. :)

rap1ds: ...and I just realized that this value is probably passed to JS as an integer and thus my comment should be ignored.

PR #997, opened 2015-02-17
Update listings_controller.rb

samikha:
I was getting the error:
ActionView::Template::Error (undefined local variable or method `payment_gateway' for #<#<Class:0x0000000d581aa8>:0x00000008aad2e8>):
Just made sure payment_gateway is the same as payment_type

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -153,6 +153,7 @@ class ListingsController < ApplicationController
       logger.info "Category: #{@listing.category.inspect}"
 
       payment_type = MarketplaceService::Community::Query.payment_type(@current_community.id)
+      payment_gateway = payment_type
       allow_posting, error_msg = payment_setup_status(
                        community: @current_community,
                        user: @current_user,


rap1ds:
Hi @samikha ! Thanks for you Pull Request!

Can you please explain how this fixes the issue you are having? What's happening here is that you are setting a new local variable `payment_gateway` which is neither used in the controller action method nor passed to the view.

samikha:
Hi @rap1ds 
The listings/form/form_content which is rendered in the ListingController/new does use payment_gateway. I guess it might be better to fix it there and change it to  payment_type in _form_content.haml, but I wasn't sure if this would break anything else.

S

PS. I come from a Django background and completely new to RoR, so I could be wrong, but I was getting the error when creating a new listing.

rap1ds:
@samikha 

There are two ways to pass variables to views:

1. Make is an instance variable (e.g. `@my_variable = "mikko"`)
2. Pass it to `render` method in a `locals` hash (e.g. `render locals: {name: "mikko"})`)

The `ListingController#new` action uses the option 2. It passes the result of a method call `commission(@current_community)` to the `render` method in a `locals` hash. The result of the method call is a hash itself, and that hash includes the `payment_gateway`

So, to my understanding, the change you made has no effect to the code. The local variable `payment_gateway` is not used and it's not passed to the view. Does this sound correct?

Anyway, would be very nice to find out the source of the issue you were having.

rap1ds:
I'll close the PR, but if you find out more information about the undefined local variable or method `payment_gateway' would be nice to hear.

PR #1001, opened 2015-02-18
Remove dependency to listing in tx service

ovan:


app/models/transaction.rb
```diff
@@ -64,6 +71,7 @@ class Transaction < ActiveRecord::Base
   validates_presence_of :payment_gateway
 
   monetize :minimum_commission_cents, with_model_currency: :minimum_commission_currency
+  monetize :unit_price_cents, with_model_currency: :unit_price_currency
```
sktoiva: Why is there different currency for commission and units?

db/migrate/20150213091223_add_listing_fields_to_transaction.rb
```diff
@@ -0,0 +1,8 @@
+class AddListingFieldsToTransaction < ActiveRecord::Migration
+  def change
+    add_column :transactions, :listing_title, :string, after: :listing_quantity
+    add_column :transactions, :listing_author_id, :string, after: :listing_quantity
+    add_column :transactions, :unit_price_cents, :integer, after: :listing_title
+    add_column :transactions, :unit_price_currency, :string, limit: 8, after: :unit_price_cents
```
sktoiva: I assume you want to avoid using add_money and its automagic features?

app/models/transaction.rb
ovan: I prefer all monetary values to be stored with separate currency to avoid update on one value to affect others. In this case the purpose is to never modify these amounts after insert but still it feels more clean this way.

Haven't used the money rails gem too much so being explicit helped me understand what actually happens here. Makes the code more readable to me.

db/migrate/20150213091223_add_listing_fields_to_transaction.rb
ovan: Yeah, I don't see the point of extra syntax to learn, write and read for a simple thing like this. Also not sure if you can control the varchar length like this with it.

PR #1002, opened 2015-02-18
Don't throw on nil community link

ovan:


rap1ds:
Why is this error happening? Probably because there's broken data in the database. In this case, there's probably a community_id that points to an unexisting community (probable an earlier deleted community).

Instead of adding this code, should we just clean the broken data from database?

rap1ds:
The code changes look ok :+1:

ovan:
Yes, true, it's about tolerating broken data we shouldn't have in the first place. I already fixed the data for this case but just thought it would be nice to tolerate it with little effort.

PR #1003, opened 2015-02-19
Move model access from tx service to store

ovan:


app/services/transaction_service/store/transaction.rb
```diff
@@ -0,0 +1,130 @@
...
+      conversation.messages.build({
+          content: tx_data[:content],
+          sender_id: tx_data[:starter_id]})
+    end
+  end
+
+  def build_booking(tx_model, tx_data)
+    if is_booking?(tx_data)
+
+      # TODO What's the correct place for the booking calculation logic?
```
sktoiva: A good question, because when adding shipping this same comes to mind. 

We need to calculate OrderTotal, ItemTotal and any additional costs and then make sure everything adds up. Is this store correct place to do this kind of logic?

ovan: Even more relevant with upcoming booking options. At the moment this just keeps the existing functionality which is to verify that the incoming listing_quantity matches with the booking dates. Let's revisit this when are further along with shipping / booking and see if this check should just be eliminated here.

PR #1008, opened 2015-02-22
Moved asset precompilation step before indexing.

klemensschindler:
Asset precompilation should occur before indexing, otherwise indexing exits with an error on the assets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,9 @@ Note: If you encounter problems with the installation, you can try asking for he
 * Install [Imagemagick](http://www.imagemagick.org)
 * run `bundle install` in the project root directory (sharetribe) to install required gems
 * (In the following commands, leave out the `RAILS_ENV=production` part if you want to get Sharetribe running in development mode.) Load the database structure to your database: `rake RAILS_ENV=production db:schema:load`
+* If you want to run Sharetribe in production mode (i.e. you are not developing the software) you'll need to precompile the assets. This puts the Javascript and CSS files in right places. Use command: `rake assets:precompile`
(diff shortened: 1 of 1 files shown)


rap1ds:
Hi @klemensschindler 

Thansk for your pull request!

However, I don't completely understand the rationale behind this. 1) Why would you have an error on the assets? Shouldn't that be fixed if it happens? 2) These commands are separated from each other and error in assets does not affect to the indexing, or does it?

klemensschindler:
Hi Mikko,

To be honest, I don't know too much about the internals of sharetribe, but
this is what happened:
I tried running a site in production mode, following the instructions.
Indexing failed on some assets (I forgot to save the concrete error though).
Not only did indexing fail, the indexing daemon could not be started.
Only after precompilation of the assets. indexing succeeded and the daemon
be started.

So you could be right that this pull request only fixes a symptom and not a
the root cause, but you can probably judge this better than I. In that case
we should probably turn this into an issue.

Furthermore: I did not try it, but I suspect that this problem only occurs
when running in production mode, not in development mode (since assets
don't have to be precompiled there). If this is so, it would explain why
this does not occur more often during sharetribe develpoment.

Kind Regards,

Klemens

rap1ds:
Hi @klemensschindler 

Ok! Thanks for your detailed explanation!

It's very hard to me to see the connection between these two commands, so I guess this was fixing only symptom as you said. So I'm going to close the pull request.

Anyway, thanks for reporting the issue! We will keep our ears and eyes open and if there are more reports regarding this, then there must be something that needs to be fixed.

PR #1011, opened 2015-02-23
Remove previous hotfix and correctly handle xhr requests

sktoiva:


app/controllers/listings_controller.rb
```diff
@@ -32,31 +32,15 @@ class ListingDeleted < StandardError; end
...
-      # Returns the listings for one person formatted for profile page view
-      per_page = params[:per_page] || 200 # the point is to show all here by default
-      page = params[:page] || 1
-      render :partial => "listings/profile_listings", :locals => {:person => @person, :limit => per_page}
-      return
-    end
-    # -- HOTFIX, REMOVE
-
     respond_to do |format|
-      format.js do
```
ovan: So was the problem that format.js is not the same as request.xhr?

sktoiva: Yes, format.js is not the same. It responds to the corresponding HTTP_ACCEPT header-field. Moreover, the order is important as if the HTTP_ACCEPT field contains '*/*', it'll pick the first one.

ovan: Ok, makes sense. PR looks good to me.

PR #1012, opened 2015-02-23
Tweak signup without first and last name

bingxie:
The user's requirement is when I configurated the community with real_name_required: false, I don't want show the first name and last name on signup form.

### DIFF
--- a/app/views/people/new.haml
+++ b/app/views/people/new.haml
@@ -47,7 +47,7 @@
     - if @current_community.only_organizations
       = form.label :organization_name, t('.organization_name')
       = form.text_field :organization_name, :class => :text_field, :maxlength => "30"
-    - else
+    - elsif @current_community.real_name_required?
       = form.label :given_name, t('.given_name')
       = form.text_field :given_name, :class => :text_field, :maxlength => "30"
       = form.label :family_name, t('.family_name')
--- a/features/people/user_creates_a_new_account.feature
+++ b/features/people/user_creates_a_new_account.feature
@@ -68,8 +68,9 @@ Feature: User creates a new account
     Then I should see "This field is required."
     When given name and last name are not required in community "test"
     And I am on the signup page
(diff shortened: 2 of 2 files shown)


gnomet:
Hi @bingxie!

Thanks for the pull request! This makes sense indeed regarding the feature.

However, this feature (real name not required) has had so little amount of use, so it's planned to be removed in the upcoming new versions of Sharetribe, and thus I won't be merging this improvement to the main code base now, as for other users of the open source version it might seem strange if this is feature is first extended but then soon after removed.

I'm very sorry for this approach towards your contribution, which by itself does make a lot of sense for a marketplace that wants to operate with usernames only. However as we strive to keep our code base simple and maintainable (aiming to get it easier to understand than it is now) we're checking the usage of many features and removing the most rarely used ones in future version.

It doesn't mean that this addition couldn't be useful for you, and I hope you manage to keep using it. Just pay attention when upgrading Sharetribe version as you might need to do some manual steps in order to keep the "real name not required" feature active.

It always annoys me to say no to contributions, but at the moment we are in a challenging phase of trying to improve the core, and the aim is to develop towards a system where it would be easier to add contributions and extend functionality without needing to plug those extensions as part of the core code. I don't know yet how that will look like and when it will be reality, but hopefully that would increase our ability to accept pull requests.

So I thank you very much for wanting to share your code with others, but unfortunately I'll be closing this for now. Hope you understand.

bingxie:
Hi @gnomet , thanks for your reply. I totally understand you and the consideration for the open source project. 
Here I also want to let you know, the users are not used to input real name separately(first name + last name) in China. So for the system, it needs find a way to extend. 
Looking forward to the change on open source sharetribe.

gnomet:
Hi @bingxie, thanks for understanding and also the comment about Chinese system for entering names. It is a good point and probably needs to be taken into account at some point.

PR #1013, opened 2015-02-23
Specify to infer spec type from file location

bingxie:
To get rid of the warning for the coming rspec3.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -66,6 +66,10 @@ prefork = lambda {
     config.mock_with :rspec
     config.include Devise::TestHelpers, :type => :controller
     config.include SpecUtils
+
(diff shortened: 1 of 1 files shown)


rap1ds:
I will leave this open for discussion, but I have a feeling that instead of this, we could just add the test type to each RSpec file. To me this feels like unnecessary magic, which also prevents us from arranging the specs in different directory structure.

bingxie:
This will let us still keep the dir convention in mind. But if we decided to add type to each file. After we finished the change, we can get rid of this magic.

ovan:
Hi @bingxie. I'm closing this now as we have not yet upgraded to rspec3. I prefer to force an explicit choice about this at upgrade time instead.

bingxie:
That's ok!

PR #1014, opened 2015-02-23
Translation service

Gnito:


db/migrate/20150224140913_create_community_translations.rb
```diff
@@ -0,0 +1,12 @@
+class CreateCommunityTranslations < ActiveRecord::Migration
+  def change
+    create_table :community_translations do |t|
+      t.integer :community_id,    null: false
+      t.string  :locale,          null: false
```
ovan: VARCHAR 256 is a bit much for this. :) 16 should do?

app/services/translation_service/store/translation.rb
```diff
@@ -0,0 +1,201 @@
...
+        { translation_key: key,
+          translations: translations
+        }
+      }
+  end
+
+  # Get translations
+  # Format for params:
+  # {community_id: 1, translation_keys: ["aa", "bb", "cc"], locales: ["en", "fi-FI", "sv-SE"], options: {}}
+  def get(community_id:, translation_keys: [], locales: [], options: {})
```
ovan: :use_fallbacks is the only option so it would be more descriptive to just have that as a keyword param with default value. This makes it clear both at the call site as well as here when reading the code.

app/services/translation_service/store/translation.rb
```diff
@@ -0,0 +1,201 @@
...
+  #       , translation: "Tervetuloa"
+  #       }
+  #     ]
+  #   }
+  # ]
+  def create(community_id:, translation_groups: [])
+    if translation_groups.empty?
+      msg = "You must specify 'translation_groups' as an array of hash-objects containing translation_key and array of translations - like translation_groups: [ { translation_key: nil, translations: [ { locale: 'en-US' , translation: 'Hi!'}, { locale: 'fi-FI', translation: 'Moi!'}]}]"
+      raise ArgumentError.new(msg)
+    end
```
ovan: I'm being using the style where API input parameter validation belong to API implementation instead of store and the store only focuses on implementing the storing and retrieving of hashes.

app/services/translation_service/api/translations.rb
```diff
@@ -0,0 +1,58 @@
+module TranslationService::API
+
+  class Translations
+
+    TranslationStore = TranslationService::Store::Translation
+
+    ## GET /translations/:community_id/
+    def get(community_id, find_params = {})
```
ovan: What about using EntityUtils.define_builder to document and validate the find_params input hash?

db/migrate/20150224140913_create_community_translations.rb
Gnito: Yeah, you are right. 
(It seems that we have the same 256 in other translation-tables too, but hopefully rows are not using it all :)

app/services/translation_service/store/translation.rb
Gnito: So, there is 1-on-1 mapping between api calls (/function) and store functions? 
I mean that if this is not the case, then we are repeating argument validation in several locations.

app/services/translation_service/api/translations.rb
Gnito: Good point :)

app/services/translation_service/store/translation.rb
ovan: No there's no 1-on-1 mapping between api and store. The idea would be that API is the public interface and good validation + error msgs is important. Store is internal interface so less validation and fail with nil pointers / less descriptive errors is fine. No need to double validate.

Gnito: yeah, my point was just that there is less repeated code if arguments are checked in single point instead of all the places that redirects to that one point.

ovan: Oh ok, now I see your point. I think you go a long with avoiding code dup with private helper functions. And then when you need it, you have more context, like api method name, etc., at API endpoint to produce an error msg to user than in some later shared part of the code path. A bit generic comment but as a principle I like API validates what comes in and further down the road we can trust more on input data.

app/services/translation_service/translations_api.md
```diff
@@ -0,0 +1,318 @@
...
+        , locale: "fi-FI"
+        , translation: "Tervetuloa"
+        }
+      ]
+  }
+, { translation_key: "tr54rdfgdrted"
+  , translations:
+      [ { translation_key: "tr54rdfgdrted"
+        , locale: "en-US"
+        , translation: "Welcome"
```
bingxie: According to the request, here should be "Hi".

spec/services/translation_service/api/translation_spec.rb
```diff
@@ -0,0 +1,191 @@
+require_relative '../api'
+
+describe TranslationService::API::Translations do
```
bingxie: So this file name should be translations_spec.rb

spec/services/translation_service/api.rb
```diff
@@ -0,0 +1,11 @@
+module TranslationService::API
```
bingxie: I think we don't need this class here.

app/models/community_translation.rb
```diff
@@ -0,0 +1,25 @@
+# == Schema Information
+#
+# Table name: community_translations
+#
+#  id              :integer          not null, primary key
+#  community_id    :integer          not null
+#  locale          :string(16)       not null
+#  translation_key :string(255)      not null
+#  translation     :string(255)
```
rap1ds: Hmm... This may not be enough. We already have many translated strings that are longer than this. For example, the length of `getting_started.info_texts` under `admin` is 462

app/services/translation_service/api/translations.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  class Translations
+
+    TranslationStore = TranslationService::Store::Translation
+
+
+    ## GET /translations/:community_id/
+    def get(community_id, request_params = {})
+      begin
+        params = TranslationService::DataTypes::Translation
```
rap1ds: I saw from earlier comments that @ovan suggested using DataTypes to validate the params. @ovan was this what you had in your mind?

I'm not completely confident with this solution. To my understanding throwing an exception and immediately catching it is considered as a bad practice. I've always heard the mantra that "exceptions are for exceptional situations", e.g. network failure, disk failure etc. Someone calling API with invalid params is not that exceptional.

app/services/translation_service/api/translations.rb
```diff
@@ -0,0 +1,66 @@
...
+      rescue Exception => error
+        msg = "Translation_groups data structure is not valid"
+        Result::Error.new(msg)
+      end
+
+    end
+
+    # PUT /translations/:community_id/
+    def update(community_id, translation_groups = [])
+      raise NoMethodError.new("Not implemented")
```
rap1ds: I think we can remove this.

spec/services/translation_service/api/translation_spec.rb
Gnito: This is in line with our other services. 
This way, it is easier to understand what test is in question since our folder structure is separated by function (and different folders might contain files with similar names)
some_service
> api
+->file1_spec.rb
> datatypes
+->file1_spec.rb
> store
+->file1_spec.rb

app/services/translation_service/translations_api.md
Gnito: Good point!

app/models/community_translation.rb
Gnito: Good point. Should these be text instead of varchars?

Or should we extend this structure when we add longer translations.
I.e. one option could be that we add separate table for 'long texts' and add just column to this table 
(some kind of long_text_ref_id)

spec/services/translation_service/api.rb
Gnito: Also in line with other services, but lets see what the team thinks about this.

app/services/translation_service/api/translations.rb
Gnito: Good question!
Personally, I felt that this (DataTypes) is a good way to test the structure of parameters. 
But how to  do it - suggestions are welcomed.

spec/services/translation_service/api/translation_spec.rb
bingxie: I see.

app/models/community_translation.rb
Gnito: Since we are not using this as a search key, I guess we could go just with text?

PR #1016, opened 2015-02-24
Fix bug: No error message because of double redirect

rap1ds:
When there's an error from PayPal, we lost the error message because of double rendering when user is changing her PayPal account.

Flash messages are only persisted for the next redirect. However, we are redirecting the user from `permissions_verified` -> `new` -> `show`. Instead of this, redirect from `permissions_verified` -> `show` if user account exists.

### DIFF
--- a/app/controllers/paypal_accounts_controller.rb
+++ b/app/controllers/paypal_accounts_controller.rb
@@ -214,7 +214,15 @@ class PaypalAccountsController < ApplicationController
       end
 
     flash[:error] = error_msg
-    redirect_to new_paypal_account_settings_payment_path(@current_user.username)
+    redirect_to action: error_redirect_action
+  end
+
+  def error_redirect_action
+    if PaypalHelper.account_prepared_for_user?(@current_user.id, @current_community.id)
+      :show
+    else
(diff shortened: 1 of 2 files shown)


app/controllers/paypal_accounts_controller.rb
```diff
@@ -214,7 +214,18 @@ def flash_error_and_redirect_to_settings(error_response: nil, error_msg: nil)
       end
 
     flash[:error] = error_msg
-    redirect_to new_paypal_account_settings_payment_path(@current_user.username)
+    redirect_to error_redirect_to
+  end
+
+  def error_redirect_to
```
ovan: I'd have this return just the action. That way you would avoid the assign from conditional -pattern that is always a bit hard to read, and also an extra local variable.

rap1ds: The rationale is this:

`redirect_to redirect_destination`

And then the implementation can be:

```ruby
def redirect_destination
  {action: :whatever}
end

# or

def redirect_destination
  new_settings_whatever_path(@current_user.id)
end
```

PR #1019, opened 2015-02-25
Marketplace username 3/3

rap1ds:
- [x] Remove `domain_alias` column
- [x] Remove all subdomain from `domain` column
- [x] Change community fetch so that always assumes full domain in `communities.domain`

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -177,24 +177,7 @@ class ApplicationController < ActionController::Base
   end
 
   def needs_redirect?(host, domain)
-    if domain.nil?
-      false
-    elsif is_subdomain?(domain)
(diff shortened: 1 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -164,6 +165,38 @@ def fetch_community
...
+
+    if needs_redirect?(host, domain)
+      redirect_to "#{request.protocol}#{domain}"
+    end
+  end
+
+  def needs_redirect?(host, domain)
+    if domain.nil?
+      false
+    elsif is_subdomain?(domain)
```
Gnito: Needs to be removed (as stated by branch creator in pr 2/3)

Gnito:
:+1:

PR #1022, opened 2015-02-26
Marketplace username 1/3

rap1ds:
- [x] Add a `username` column to `communities`
- [x] Populate `username` column by values from `domain_alias` and `domain`
- [x] Save `username` to communities table from a new marketplace is created

### DIFF
--- a/app/models/community.rb
+++ b/app/models/community.rb
@@ -3,6 +3,7 @@
 # Table name: communities
 #
 #  id                                         :integer          not null, primary key
+#  username                                   :string(255)      not null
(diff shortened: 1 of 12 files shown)


Gnito:
Looks good (even though username sounds a bit strange)

gnomet:
Looks good to me. As Vesa mentioned, the username term is a bit strange, but as long as it's not in the UI it's clear enough in my opinion.

rap1ds:
Ok, I'll change the `username`, it is clearly confusing.

PR #1023, opened 2015-02-26
Marketplace username 2/3

rap1ds:
- [x] Use username to fetch community, if community is not found by domain
- [x] Do a redirect to the full domain, if needed.
- [x] Do NO save `domain` to communities table from a new marketplace is created
- [x] Fix tests and change factories
- [x] Remove the custom `find_by_domain` method from `Community`
- [x] Remove all usage of `find_by_domain` (expect `ApplicationController#fetch_community`)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,6 +17,7 @@ class ApplicationController < ActionController::Base
     :check_auth_token,
     :fetch_logged_in_user,
     :fetch_community,
+    :redirect_to_marketplace_domain,
     :fetch_community_membership,
     :set_locale,
     :generate_event_id,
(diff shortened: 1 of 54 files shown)


app/controllers/application_controller.rb
```diff
@@ -164,6 +165,38 @@ def fetch_community
...
+
+    if needs_redirect?(host, domain)
+      redirect_to "#{request.protocol}#{domain}"
+    end
+  end
+
+  def needs_redirect?(host, domain)
+    if domain.nil?
+      false
+    elsif is_subdomain?(domain)
```
rap1ds: Remove this is 3/3

Gnito:
Uh, this was a long one. I didn't spot anything strange - hopefully Travis likes it too.

rap1ds:
Yeah, a bit too long :( `domain` was used in many places so I had to touch a lot of files. Thanks @Gnito for checking this out!

gnomet:
Yea, lots of stuff.. Won't re-read if one review is done.

one thought that came to me was if installation instruction need update, but it seems to be better now than I remembered, so probably this won't change anything there..

PR #1026, opened 2015-02-26
Tx service gw adapters 2/X

ovan:


app/services/transaction_service/gateway/braintree_adapter.rb
```diff
@@ -20,12 +22,41 @@ def create_payment(tx:, gateway_fields:, prefer_async: nil)
...
+      if !result.success?
+        SyncCompletion.new(Result::Error.new(result.message))
+      end
+
+      SyncCompletion.new(Result::Success.new({result: true}))
+    end
+
+    def get_payment_details(tx:)
+      payment_total = Maybe(PaymentModel.where(transaction_id: tx[:id]).first).total_sum.or_else(nil)
+      total_price = tx[:unit_price] * 1 # TODO fixme for booking (model.listing_quantity)
```
rap1ds: This needs to be done before merge?

app/services/transaction_service/gateway/checkout_adapter.rb
```diff
@@ -3,9 +3,20 @@ module TransactionService::Gateway
...
 
+    PaymentModel = ::Payment
+
     def implements_process(process)
       [:none, :postpay].include?(process)
     end
 
+    def get_payment_details(tx:)
+      payment_total = Maybe(PaymentModel.where(transaction_id: tx[:id]).first).total_sum.or_else(nil)
+      total_price = tx[:unit_price] * 1 # TODO fixme for booking (model.listing_quantity)
```
rap1ds: Same question here

config/routes.rb
```diff
@@ -299,8 +299,7 @@
             put :confirmation, to: 'confirm_conversations#confirmation' #TODO these should be under transaction
             get :accept_preauthorized, to: 'accept_preauthorized_conversations#accept'
             get :reject_preauthorized, to: 'accept_preauthorized_conversations#reject'
-            put :acceptance_preauthorized, to: 'accept_preauthorized_conversations#accepted', constraints: ParamsConstraints.new({listing_conversation: {status: "paid"}})
-            put :acceptance_preauthorized, to: 'accept_preauthorized_conversations#rejected', constraints: ParamsConstraints.new({listing_conversation: {status: "rejected"}})
+            put :acceptance_preauthorized, to: 'accept_preauthorized_conversations#accepted_or_rejected'
```
rap1ds: I guess this is the only place where ParamsConstraints was used, so the file should be removed.

rap1ds:
You found out the bug (or at least very confusing behaviour) of `result.maybe` returning None if the data is empty. What do you think, shouldn't we fix it quite soon? I mean in general, not maybe in this branch.

rap1ds:
TransactionStore is now tightly coupled to the Conversation (`add_message`, `mark_as_unseen`, e.g.) What do you think, is it ok for now or should it be changed?

rap1ds:
`TransactionService.charge_commission` seems to be Paypal specific. Should it be in adapter instead?

ovan:
result.maybe: Yes, should be fixed to avoid others having to discover this odd behavior and unnecessary debugging. Not in this branch.

coupled conversation: This is not modelled optimally but also not in the scope of this change. This just moves model level operations behind a store interface but doesn't change the way things are modelled under the hood. Hopefully this makes changing later easier because the surface area of the modelling implications is smaller.

Good point about charge commission. Adapter is not exactly the correct place, it's a process level thing somehow. I left out the part where processes would handle the async behavior consistently with the sync behavior because this includes moving actions away from statesman and MarketplaceService::Transaction. Not sure when this work will happen but at least we need to clarify what to do with preauth flow before that. Charge commission should be addressed then too.

config/routes.rb
ovan: Ok, will do.

app/services/transaction_service/gateway/braintree_adapter.rb
ovan: I need to check this once more. It's a copy & paste comment so no behavior change here between master and this branch but I could either remove the comment or then change something.

PR #1027, opened 2015-02-27
Confirm two days after booking ended

rap1ds:


rap1ds:
Waiting tests to pass...

PR #1028, opened 2015-02-27
Handle delayed delta indexing quietly without waiting

sktoiva:


config/thinking_sphinx.yml
```diff
@@ -6,6 +6,7 @@ production:
   address: <%= ENV['SPHINX_HOST'] || '' %>
   enable_star: true
   min_infix_len: 3
+  quiet_deltas: true
```
ovan: What does this do?

PR #1033, opened 2015-03-04
Add shipping fields to new listing and edit listing forms

sktoiva:


app/controllers/listings_controller.rb
```diff
@@ -171,7 +171,8 @@ def create
 
     params[:listing] = normalize_price_param(params[:listing]);
 
-    @listing = Listing.new(params[:listing])
+    create_params = create_listing_params(params[:listing])
```
ovan: unnecessary variable

app/controllers/listings_controller.rb
```diff
@@ -559,4 +566,10 @@ def delivery_config(require_shipping_address, pickup_enabled, shipping_price, cu
       []
     end
   end
+
+  def create_listing_params(params)
```
ovan: This could and should be implemented as pure function instead of modifying input params.

app/controllers/listings_controller.rb
```diff
@@ -222,9 +223,11 @@ def update
 
     @listing.custom_field_values = create_field_values(params[:custom_fields])
 
-    params[:listing] = normalize_price_param(params[:listing]);
+    params[:listing] = normalize_price_param(params[:listing])
+
+    update_params = create_listing_params(params[:listing])
```
ovan: unnecessary var

app/services/marketplace_service/community.rb
```diff
@@ -36,6 +36,12 @@ def payment_type(community_id)
           .or_else(nil)
       end
 
+      def shipping_enabled?(community_id)
+        Maybe(MarketplaceSettings.where(community_id: community_id).first)
+          .map { |settings| settings.shipping_enabled }
+          .or_else(nil)
+      end
```
ovan: since this is temporary it could only exist in controller instead of in deprecated service

PR #1034, opened 2015-03-05
Support chinese, korean and japanese search

bingxie:


rap1ds:
Thanks @bingxie !

@sktoiva Can you take a look at this? If I remember right you knew something about sphinx charsets etc.

gnomet:
I checked and tested and worked nicely. Thanks @bingxie!

bingxie:
Thanks!

PR #1035, opened 2015-03-05
Transaction process 1/4: Add new schema, save new data to new schema

rap1ds:
- [x] Create a TransactionProcess model
- [x] Each TransactionType has one TransactionProcess. Actually, the relation is `belongs_to` since the foreign key is in the TransactionType model
- [x] Save TransactionProcess in TransactionTypeCreator

### DIFF
--- a/app/services/transaction_type_creator.rb
+++ b/app/services/transaction_type_creator.rb
@@ -68,7 +68,7 @@ module TransactionTypeCreator
 
   module_function
 
-  def create(community, transaction_type_class_name)
+  def create(community, transaction_type_class_name, opts = {})
     throw "Transaction type #{transaction_type_class_name} not available. Available types are: #{available_types.join(', ')}" unless available_types.include? transaction_type_class_name
 
     transaction_type_description = TRANSACTION_TYPES[transaction_type_class_name]
@@ -92,6 +92,8 @@ module TransactionTypeCreator
     #enable preauthorized payments
     transaction_type.preauthorize_payment = true
 
+    create_transaction_process!(community, transaction_type, opts[:payment_gateway_available])
+
     transaction_type.save!
 
     # Categories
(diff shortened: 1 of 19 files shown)


db/migrate/20150303134630_create_transaction_processes.rb
```diff
@@ -0,0 +1,9 @@
+class CreateTransactionProcesses < ActiveRecord::Migration
+  def change
+    create_table :transaction_processes do |t|
+      t.string :process, null: false
```
Gnito: btw. what do you think, should this string be limited from default 255?

Gnito:
:+1:

test/helper_modules.rb
```diff
@@ -48,11 +63,13 @@ def self.load_test_categories_and_transaction_types_to_db(community)
 
     def self.load_categories_and_transaction_types_to_db(community, transaction_types, categories)
       # Load transaction types
-      transaction_types.each do |type, translations|
+      transaction_types.each do |type, opts|
 
         transaction_type = Object.const_get(type.to_s).create!(:type => type, :community_id => community.id)
+        transaction_type.create_transaction_process(process: opts[:process])
```
sktoiva: It'd be better to use ```create_transaction_process!``` as it throws instead of returning false on invalid association object.

test/helper_modules.rb
```diff
@@ -48,11 +63,13 @@ def self.load_test_categories_and_transaction_types_to_db(community)
...
+        transaction_type.save!
```
sktoiva: Why is this needed?

app/services/transaction_type_creator.rb
```diff
@@ -110,4 +112,15 @@ def use_in_category(category, transaction_type)
...
 
+  # This method is only temporary implementation for the transaction_process change
+  def create_transaction_process!(community, transaction_type, payment_gateway_available)
+    payment_gateway_available = payment_gateway_available || MarketplaceService::Community::Query.payment_type(community.id).present?
+    process = ListingsController.select_payment_process(
+          price_field: transaction_type.price_field?,
+          preauthorize: transaction_type.preauthorize_payment?,
+          payment_gateway_available: payment_gateway_available)
+
+    transaction_type.create_transaction_process({ process: process })
```
sktoiva: Better to use ```create_transaction_process!``` as it throws.

test/helper_modules.rb
rap1ds: You mean `save!`?

The relation between TransactionType and TransactionProcess is so that TransactionType _belongs\_to_ TransactionProcess, which means that the foreign key is store to TransactionType.transaction\_process\_id

When `transaction_type.create_transaction_process` is called, a new TransactionProcess is instantiated AND saved to database. The id of the newly created TransactionProcess is set to TransactionType but NOT saved. That's why `save` is needed.

This wouldn't be the case if the relationship was TransactionType _has\_one_ TransactionProcess. The calling `transaction_type.create_transaction_process` would instantiate and save the TransactionProcess, including the foreign key `transaction_type_id`

PR #1036, opened 2015-03-05
Transaction process 1/3: Add new schema, migrate data from old schema to new schema

rap1ds:
- [x] Create a TransactionProcess model
- [x] Each community has one TransactionProcess per process type
- [x] Save TransactionProcess in TransactionTypeCreator

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -110,7 +110,7 @@ class ListingsController < ApplicationController
 
     payment_gateway = MarketplaceService::Community::Query.payment_type(@current_community.id)
 
-    payment_process = select_payment_process(
+    payment_process = ListingsController.select_payment_process(
       price_field: @listing.transaction_type.price_field?,
       preauthorize: @listing.transaction_type.preauthorize_payment?,
       payment_gateway_available: payment_gateway.present?)
@@ -512,7 +512,8 @@ class ListingsController < ApplicationController
     end
   end
 
-  def select_payment_process(price_field:, payment_gateway_available:, preauthorize:)
+  # This is wrong place, yes, but this method will be soon removed
+  def self.select_payment_process(price_field:, payment_gateway_available:, preauthorize:)
     case [price_field, payment_gateway_available, preauthorize]
     when matches([false])
       :none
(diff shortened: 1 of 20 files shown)


Gnito:
:+1:

rap1ds:
@Gnito Yes they do. When running down-stream migration, how can you tell which were null before running the initial up-stream migration? :) That's the problem.

app/controllers/listings_controller.rb
```diff
@@ -512,7 +512,8 @@ def payment_setup_status(community:, user:, listing:, payment_type:)
     end
   end
 
-  def select_payment_process(price_field:, payment_gateway_available:, preauthorize:)
+  # This is wrong place, yes, but this method will be soon removed
+  def self.select_payment_process(price_field:, payment_gateway_available:, preauthorize:)
```
ovan: Why class method?

app/services/transaction_type_creator.rb
```diff
@@ -66,17 +66,26 @@ module TransactionTypeCreator
...
   module_function
 
-  def create(community, transaction_type_class_name)
-    throw "Transaction type #{transaction_type_class_name} not available. Available types are: #{available_types.join(', ')}" unless available_types.include? transaction_type_class_name
+  def create(community, transaction_type_class_name, process)
+    throw "Transaction type '#{transaction_type_class_name}' not available. Available types are: #{available_types.join(', ')}" unless available_types.include? transaction_type_class_name
+    throw "Transaction process '#{process}' not available. Available processes are: #{TRANSACTION_PROCESSES.join(', ')}" unless TRANSACTION_PROCESSES.include? process.to_sym
+
+    author_is_seller = transaction_type_class_name != "Request"
+    transaction_process = TransactionProcess.where(community_id: community.id, author_is_seller: author_is_seller, process: process).first_or_create
```
ovan: Is there a bang version of first_or_create?

db/migrate/20150303140556_add_transaction_process_id_to_transaction_type.rb
```diff
@@ -0,0 +1,10 @@
+class AddTransactionProcessIdToTransactionType < ActiveRecord::Migration
+  def up
+    add_column :transaction_types, :transaction_process_id, :int, after: :community_id
+    add_index :transaction_types, :transaction_process_id
```
ovan: What do we need this index for? What's the use case for looking up a transaction type by transaction process?

app/services/transaction_type_creator.rb
rap1ds: Yes there is!

PR #1037, opened 2015-03-05
Transaction process 2/3: Read from new schema

rap1ds:
- [x] Read from TransactionProcess

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -113,7 +113,11 @@ class ListingsController < ApplicationController
     # TODO Change this so that the path is always the same, but the controller
     # decides what to do. We don't want to make a API call to TransactionService
     # just to show a listing details
-    process = TransactionProcess.find(@listing.transaction_type.transaction_process_id).process
+    process_res = TransactionService::API::Api.processes.get(
+      community_id: @current_community.id,
+      process_id: @listing.transaction_type.transaction_process_id
+    )
+    process = process_res.data[:process].to_sym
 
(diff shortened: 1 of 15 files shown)


Gnito:
:+1:

app/controllers/listings_controller.rb
```diff
@@ -110,15 +110,19 @@ def show
...
-    payment_process = select_payment_process(
-      price_field: @listing.transaction_type.price_field?,
-      preauthorize: @listing.transaction_type.preauthorize_payment?,
-      payment_gateway_available: payment_gateway.present?)
+    # TODO Change this so that the path is always the same, but the controller
+    # decides what to do. We don't want to make a API call to TransactionService
+    # just to show a listing details
+    process_res = TransactionService::API::Api.processes.get(
+      community_id: @current_community.id,
+      process_id: @listing.transaction_type.transaction_process_id
```
ovan: Listing must record transaction_process_id reference at create/edit and use that. Is the plan to change that part at later? If so then perhaps add a TODO comment to not miss it later.

rap1ds: Yes, that will be done later.

app/services/transaction_service/process.md
```diff
@@ -0,0 +1,40 @@
...
+Request: Empty
+
+Response:
+
+
+```json
+{
+  id: 123,
+  process: "preauthorize" # or "none" or "postpay"
+}
```
rap1ds: This is an array

PR #1041, opened 2015-03-05
Pick which migrations to run

rap1ds:


gnomet:
Based on looking the code, looks good. Nice that you can choose which migrations to pick. :+1:

lib/tasks/deploy.rake
```diff
@@ -25,6 +25,16 @@ task :deploy_to, [:destination] do |t, args|
   )
 end
 
+# Runs multiple up migrations.
+#
+# This task is mostly run in Heroku during the deploy. No need to run this locally.
+#
+# Usage: rake migrate_up[20150226124214, 20150226124215, 20150226124216]
```
sktoiva: Why do we need an array? Passing just one migrates up to that version (or down to that).

rap1ds: @sktoiva The deploy script let's user to pick which migrations to run, so for example if you have 3 new migrations A, B and C and you want to run only A and C, passing just one migrate up to C would also run B and that would be unexpected result.

Then you might ask, why on earth someone would like to skip a migration in the middle? A common use case is that you want to run A and B, then wait for a while until the new code kicks in after preboot, but why A and C?

Consider a situation where you have added three new migrations A, B and C in your branch. When you deploy the change, you want to run A and B, then wait for the new code and after that run C. However, at the same time, just a little bit before you someone has added another migration D, merged it to and deployed it. When you deploy your changes, the migration status at the remote server is:

down A
down B
down C
up D

Three possibilities (with the Rails default rake tasks):

1. `rake db:migrate`: Result: A, B and C are run, not good
1. `rake db:migrate:up VERSION=B`: Result: Only B is run, not good
1. `rake db:migrate VERSION=B`: Result: A and B are run but also D is rolled back, not good either.

So the only option what is left is to use `rake db:migrate:up VERSION=A` and `rake db:migrate:up VERSION=B` and that's what the migrate_up with an array of versions does under the hood.

PR #1044, opened 2015-03-09
Add mapping for one more danish character

gnomet:


gnomet:
This should fix that Helsingør was not searchable

PR #1045, opened 2015-03-09
Move shipping address to tx entity

ovan:
Shipping address was first implemented as part of the deprecated transaction_with_conversation entity format. Here we move it to current version transaction entity and read the data from the new place consistently.


sktoiva:
:+1:

PR #1050, opened 2015-03-11
Improve previous migration

rap1ds:
- Process should be none, if price_field is NULL
- Process should be none, if type is Inquiry

### DIFF
--- a/db/migrate/20150311073502_fix_transaction_process_values.rb
+++ b/db/migrate/20150311073502_fix_transaction_process_values.rb
@@ -1,0 +1,47 @@
+class FixTransactionProcessValues < ActiveRecord::Migration
+  def up
+    execute("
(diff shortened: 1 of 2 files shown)


db/migrate/20150311073502_fix_transaction_process_values.rb
```diff
@@ -0,0 +1,47 @@
...
+        LEFT JOIN transaction_processes ON (
+          transaction_processes.community_id = transaction_types.community_id AND
+          transaction_processes.process = 'none' AND
+          transaction_processes.author_is_seller = 1)
+
+        WHERE
+         transaction_types.type != 'Request' AND
+         (transaction_types.price_field IS NULL OR transaction_types.type = 'Inquiry') AND
+         transaction_processes.id IS NULL
+      )
```
ovan: Ensure we never insert duplicate none processes for a community

PR #1053, opened 2015-03-11
Remove code from TransactionType subclasses (Offer, Sell, Request, etc.)

rap1ds:
This pull request removes Single Table Inheritance from Transaction Types. All Transaction Type subclasses (Offer, Sell, Request etc.) are removed.

Our payment process is only available for Transaction Types that are "offer" (i.e. listing author is the seller, who gets the money). Thus we can move all the "directions" and simply assume that the transaction type is offer, if we are in the payment flow (preauth or postpay)

It's important to notice that the "discussion_type" is the opposite of "direction", thus the following replace rules should apply everywhere:

```ruby
listing.direction => "offer"
transaction_type.direction => "offer"
transaction.discussion_type => "request"
Listing.opposite_type(listing.transaction_type.direction) => "request"
```


app/actions/backward_compatibility.rb
```diff
@@ -0,0 +1,30 @@
+# All methods is tnis module are DEPRECATED!
+#
+# Do not use them to build anything new.
+#
+# The functions in this module support backward compatibility with parts
+# of the application that we can not just change, e.g. API
+#
+module BackwardCompatibility
```
ovan: This is a bit over generic imho. How about listing_shape_helper under view_utils instead?

app/actions/backward_compatibility.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    community.transaction_types.inject({}) { |direction_map, tt|
+      direction_map.tap { |m|
+        process = process_res.data.find { |p| p[:id] == tt.transaction_process_id }
+
+        direction = process[:author_is_seller] ? "offer" : "request"
+        m[tt.id] = direction
+      }
+    }
+  end
```
ovan: Another option, and maybe a bit more readable, would have been to map over transaction types and build an array of tuples and then call to_hash

app/controllers/free_transactions_controller.rb
```diff
@@ -38,6 +34,16 @@ def create_contact
...
     if contact_form.valid?
+      process_res = TransactionService::API::Api.processes.get(
+        community_id: @current_community.id,
+        process_id: @listing.transaction_type.transaction_process_id
+      )
+
+      unless process_res.data[:process] == "none"
+        flash[:error] = "Sending the message failed. Please try again."
+        return redirect_to root
+      end
```
ovan: Better style would be to use .maybe and explicitly handle the case of missing transaction_process by raising an error with error message. Now we get only "no method [] on Nil".

app/controllers/listings_controller.rb
```diff
@@ -241,6 +248,12 @@ def update
   end
 
   def close
+    process_res = TransactionService::API::Api.processes.get(
+      community_id: @current_community.id,
+      process_id: @listing.transaction_type.transaction_process_id
+    )
+    process = process_res.data[:process].to_sym
```
ovan: No need for .to_sym anymore

app/services/marketplace_service/listing.rb
```diff
@@ -85,8 +58,13 @@ def send_payment_settings_reminder?(listing_id, community_id)
           }
         }
 
+        process = TransactionService::API::Api.processes.get(
+          community_id: community_id,
+          process_id: listing.transaction_type.transaction_process_id
+        ).data[:process].to_sym
```
ovan: No need for .to_sym, handle missing process explicitly

app/views/listings/_listing_actions.haml
```diff
@@ -84,7 +84,7 @@
         .content
           = @listing.transaction_type.action_button_label(I18n.locale)
 
-    - if payment_gateway == :paypal && @listing.transaction_type.status_after_reply != "free"
+    - if payment_gateway == :paypal && process == :preauthorize
```
ovan: ```if payment_gateway == :paypal && process != :none```

config/application.rb
```diff
@@ -31,6 +31,7 @@ class Application < Rails::Application
     config.autoload_paths += Dir[Rails.root.join('app', 'utils')]
     config.autoload_paths += Dir[Rails.root.join('app', 'view_utils')]
     config.autoload_paths += Dir[Rails.root.join('app', 'forms')]
+    config.autoload_paths += Dir[Rails.root.join('app', 'actions')]
```
ovan: Can remove

PR #1054, opened 2015-03-11
Copys and remove links

sktoiva:


ovan:
Looks good. There's another place in posting a new listing / editing an existing listing where the same change is needed.

PR #1058, opened 2015-03-12
populate community_translations with transaction_type_translations 

Gnito:
+ indexes for community_id and translation_key

### DIFF
--- a/app/models/community_translation.rb
+++ b/app/models/community_translation.rb
@@ -10,6 +10,11 @@
 #  created_at      :datetime         not null
 #  updated_at      :datetime         not null
 #
+# Indexes
+#
+#  community_translations_key         (community_id,translation_key)
+#  community_translations_key_locale  (community_id,translation_key,locale)
+#
 
 class CommunityTranslation < ActiveRecord::Base
   validates_presence_of :locale, :translation_key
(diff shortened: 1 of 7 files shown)


app/models/community_translation.rb
```diff
@@ -10,6 +10,11 @@
 #  created_at      :datetime         not null
 #  updated_at      :datetime         not null
 #
+# Indexes
+#
+#  index_community_translations_on_community_id     (community_id)
+#  index_community_translations_on_translation_key  (translation_key)
```
Gnito: Hmm. locale is even more important. 
Maybe the best way is making index for multiple keys (community_id + locale)

db/migrate/20150312155301_insert_transaction_type_translations_to_community_translations.rb
```diff
@@ -0,0 +1,39 @@
+class InsertTransactionTypeTranslationsToCommunityTranslations < ActiveRecord::Migration
+  def up
+    # Index by community_id and translation_key
+    add_index :community_translations, :community_id            unless index_exists?(:community_translations, :community_id)
+    add_index :community_translations, [:community_id, :locale] unless index_exists?(:community_translations, [:community_id, :locale])
+    add_index :community_translations, :translation_key         unless index_exists?(:community_translations, :translation_key)
```
Gnito: I guess, this should be combined index too?

rap1ds:
:+1: Test the migration on staging before deploying to prod

Gnito:
I added write functionality for translation service. (in TransactionTypeCreator)

rap1ds:
:+1:

PR #1060, opened 2015-03-13
Coerce PayPal SDK's Strings into Ruby strings

ovan:
PayPal SDK has a special implementation of String that it retuns from
API calls instead of standard Ruby Strings. This causes YAML
serializaiton and deserialization used by workers to store operation
results to db to convert strings into fixnums in certain cases. To fix
this we coerce to standard strings before returning result from API.


sktoiva:
:+1:

PR #1070, opened 2015-03-17
Add emoji support for messages

sktoiva:


rap1ds:
Looks good to me :+1: 

One question: Should we enable this also for other tables, e.g. comments, listing description, testimonial ?

thomassnielsen:
What happened here? @sktoiva why did you close this? We're seeing the same issue, and it would be nice to know if this was closed due to some issues with the solution :)

rap1ds:
@thomassnielsen Thanks for asking! When this PR was opened, we were still using Rails 3. However, we noticed that we need to upgrade to Rails 4 first in order to support emojis (it had something to do with the MySQL adapter, I can't recall precisely)

Now that we are have upgraded to Rails 4 this has been on the table again. However, this needs some non-trivial database migrations. Let's see when this happens, hopefully soon.

thomassnielsen:
Thanks for your quick and useful answer :+1:

PR #1075, opened 2015-03-17
1/3 Listing price units

rap1ds:
## Move transaction_type 'create' behind ListingShape API

- [X] TransactionTypeCreator
- [X] Helper modules test state initializers
- [X] Factories test state initializers

## Move price_per to units

Move price_per to units table. Nothing new yet. Make sure the existing booking feature works without issues

### Deploy 1 [10/10]

- [X] Add units table
- [X] Add units Model
- [X] Write ListingShape API doc draft
- [X] Add ListingShape API and Store
- [X] Save price per to old transaction types
- [X] Save 'price_per' 'day' value to units table (unit_type: day)
- [X] Save 'price_per' NULL values to units table (unit_type: price) if price enabled
- [X] Validation: must have units if price enabled
- [X] Migrate old data from 'price_per' column to units

### Deploy 2 [0/1]

- [ ] Read from 'units' table

### Deploy 3 [0/2]

- [ ] Don't save new data to 'price_per' column
- [ ] Make TransactionType to reject 'price_per' column

### DIFF
--- a/app/models/listing_unit.rb
+++ b/app/models/listing_unit.rb
@@ -1,0 +1,19 @@
+# == Schema Information
+#
+# Table name: listing_units
+#
+#  id                  :integer          not null, primary key
+#  unit_type           :string(32)       not null
+#  translation_key     :string(64)
+#  transaction_type_id :integer
+#  created_at          :datetime         not null
+#  updated_at          :datetime         not null
+#
+
+class ListingUnit < ActiveRecord::Base
+  attr_accessible(
+    :transaction_type_id,
+    :unit_type,
+    :translation_key
+  )
+end
--- a/app/services/listing_service/listing_shapes_api.md
+++ b/app/services/listing_service/listing_shapes_api.md
@@ -22,8 +22,8 @@ Request body:
 , name_tr_key: TranslationKey.new("listing_shape.1234.123")
 , action_button_tr_key: TranslationKey.new("action_button.1234.123")
 , units:
-  [ { unit\_type: "day" }
-  , { unit\_type: "custom"
+  [ { unit_type: :day }
+  , { unit_type: :custom
     , label: TransactionKey.new("unit.1234.123")
     }
   ]
@@ -40,8 +40,8 @@ Response:
 , name_tr_key: TranslationKey.new("listing_shape.1234.123")
 , action_button_tr_key: TranslationKey.new("action_button.1234.123")
 , units:
-  [ { unit\_type: "day" }
-  , { unit\_type: "custom"
+  [ { unit_type: :day }
+  , { unit_type: :custom
     , label: TransactionKey.new("unit.1234.123")
     }
   ]
@@ -62,8 +62,8 @@ Response:
 , name_tr_key: TranslationKey.new("listing_shape.1234.123")
 , action_button_tr_key: TranslationKey.new("action_button.1234.123")
 , units:
-  [ { unit\_type: "day" }
-  , { unit\_type: "custom"
+  [ { unit_type: :day }
+  , { unit_type: :custom
     , label: TransactionKey.new("unit.1234.123")
     }
   ]
--- a/app/services/listing_service/listing_shapes_api.md
+++ b/app/services/listing_service/listing_shapes_api.md
@@ -7,11 +7,45 @@ Request: empty
 Response:
 
 ```ruby
-[ { id: 123
-  , name: TranslationKey.new("listing_shape.1234.123")
-  }
-, ...
+[
+# Array of listing shapes, see the format below
 ]
+```
+
+## POST /:community_id
+
+Request body:
+
+```ruby
+{ price_enabled: true
+, transaction_process_id: 123
+, name_tr_key: TranslationKey.new("listing_shape.1234.123")
+, action_button_tr_key: TranslationKey.new("action_button.1234.123")
+, units:
+  [ { unit\_type: "day" }
+  , { unit\_type: "custom"
+    , label: TransactionKey.new("unit.1234.123")
+    }
+  ]
+}
+```
+
+Response:
+
+```ruby
+{ id: 12345
+, community_id: 9876
+, price_enabled: true
+, transaction_process_id: 123
+, name_tr_key: TranslationKey.new("listing_shape.1234.123")
+, action_button_tr_key: TranslationKey.new("action_button.1234.123")
+, units:
+  [ { unit\_type: "day" }
+  , { unit\_type: "custom"
+    , label: TransactionKey.new("unit.1234.123")
+    }
+  ]
+}
 ```
 
 ## GET /:community_id/:listing_shape_id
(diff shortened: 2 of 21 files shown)


app/services/listing_service/listing_shapes_api.md
```diff
@@ -0,0 +1,48 @@
...
+## GET /:community_id/:listing_shape_id
+
+Request: empty
+
+Response:
+
+```ruby
+[ { id: 123
+  , name:
+    [ { locale: "en", translation: "Renting" }
```
rap1ds: Return only `translation_key`

app/services/listing_service/listing_shapes_api.md
```diff
@@ -0,0 +1,48 @@
...
+    , { locale: "fi", translation: "Vuokrataan"}
+    ]
+  , custom_fields: [ ... ]
+  , units:
+    [ { unit\_type: "day"}
+    , { unit\_type: "custom"
+      , label:
+        [ { locale: "en", translation: "ski tour trip"}
```
rap1ds: Return only `translation_key`

rap1ds:
```
label: TranslationKey.new("unit_type.1234.1234")

File: app/services/listing_service/listings_api.md [hunk 3]
```diff
@@ -0,0 +1,60 @@
...
+    { value: 50
+    , unit:
+      { unit\_type: "day"
+      }
+    }
+  }
+, { id: 234
+  , title: "Splitboard"
+  , description: "Splitboard snowboard"
+  , price:
```
rap1ds: unit_price (Money) and unit

app/services/listing_service/listings_api.md
```diff
@@ -0,0 +1,60 @@
...
+        , { locale: "fi", translation: "retki" }
+        ]
+      }
+    }
+  }
+]
+```
+
+
+## GET /:comunity_id/:listing_id?limit=30&offset=30&transaction_type=renting&category=snowboards
```
rap1ds: Remove limits and offsets

spec/factories.rb
```diff
@@ -98,20 +98,36 @@ def build_association(association, opts = {})
   end
 
   factory :listing do
-    title "Sledgehammer"
+    title "Sledgeham!!!!!!!!!!!!!!!!!!"
```
rap1ds: Revert

app/models/listing_unit.rb
```diff
@@ -0,0 +1,19 @@
...
+#  translation_key     :string(64)
+#  transaction_type_id :integer
+#  created_at          :datetime         not null
+#  updated_at          :datetime         not null
+#
+
+class ListingUnit < ActiveRecord::Base
+  attr_accessible(
+    :transaction_type_id,
+    :unit_type,
```
Gnito: Could this be renamed this to 'type'
It would remove some key-renames and if Unit-entity doesn't need it to be named unit_type, why ListingUnit need this longer form?

app/services/listing_service/store/shapes.rb
```diff
@@ -0,0 +1,147 @@
...
+
+    tt_model.save!
+
+    from_transaction_type_model(tt_model)
+  end
+
+  # private
+
+  def valid_to_create!(shape)
+    valid = (shape[:price_enabled] == true && !shape[:units].empty?) || (shape[:price_enabled] == false && shape[:units].empty?)
```
Gnito: Is there any reason why customer can't buy 3h worth of massage sessions and start negotiate the price based on that?
or 4 kg of potatoes?

app/services/listing_service/store/shapes.rb
```diff
@@ -0,0 +1,147 @@
...
+
+    valid_to_create!(shape)
+
+    units = shape[:units].map { |unit| Unit.call(unit) }
+    translations = opts[:translations] # Skip data type and validation, because this is temporary
+
+    # TODO We should be able to create transaction_type without community
+    community = Community.find(shape[:community_id])
+
+    create_tt_opts = to_tt_model_attributes(shape).except(:units, :translations)
```
Gnito: n.b. slice would be more readable, imho.

app/models/listing_unit.rb
rap1ds: Good point. `type` was my original idea for the name, but I decided to name it `unit_type`, because by default Rails does some special magic if the table contains column named `type`. It uses `type` for Single Table Inheritance.

At API level the `unit_type` is renamed to `type` so to the outside world the name is `type`, as it should.

app/services/listing_service/store/shapes.rb
rap1ds: Well, `slice` is a bit different thing. Here I want to take all hash keys and values, except units and translations. If there will be new keys and values in the hash in the future, I want to take them, as long as they are not units or translations.

app/services/listing_service/store/shapes.rb
rap1ds: Good point.

Now when I think this, at least the first condition is wrong. If there is a listing e.g. "House in the middle of Helsinki", the listing will have price but NOT units. Even though the price is per piece, I think it doesn't make sense to have the per piece unit, because there's only one house to sell.

About the second condition, and your example, that's also a good question. Probably need to ask Juho, or maybe I just remove this whole validation method and rely that the caller calls the API with correct params.

app/services/listing_service/store/shapes.rb
```diff
@@ -0,0 +1,140 @@
+module ListingService::Store::Shapes
```
ovan: We've seemed to establish a convention of using singular form to name stores, Shape vs Shapes, Store::PaypalPayment, Store::Transaction, etc.

app/models/listing_unit.rb
```diff
@@ -0,0 +1,19 @@
...
+#  unit_type           :string(32)       not null
+#  translation_key     :string(64)
+#  transaction_type_id :integer
+#  created_at          :datetime         not null
+#  updated_at          :datetime         not null
+#
+
+class ListingUnit < ActiveRecord::Base
+  attr_accessible(
+    :transaction_type_id,
```
ovan: transaction_type_id or listing_shape_id ?

app/services/listing_service/listing_shapes_api.md
```diff
@@ -0,0 +1,71 @@
...
+
+Request body:
+
+```ruby
+{ price_enabled: true
+, transaction_process_id: 123
+, name_tr_key: TranslationKey.new("listing_shape.1234.123")
+, action_button_tr_key: TranslationKey.new("action_button.1234.123")
+, units:
+  [ { unit\_type: "day" }
```
ovan: These types should be symbols everywhere, :day, :custom, etc.

app/services/listing_service/store/shapes.rb
```diff
@@ -0,0 +1,140 @@
...
+
+  # TODO Remove transaction_type_id
+  def get(community_id:, transaction_type_id: nil, listing_shape_id: nil)
+    if transaction_type_id
+      model = TransactionTypeModel.where(community_id: community_id, id: transaction_type_id).first
+      from_transaction_type_model(model)
+    elsif listing_shape_id
+      raise NotImplementedError.new("Can not find listing shape by listing_shape_id, yet. Specify transaction_type_id instead.")
+    else
+      rase ArgumentError.new("Can not find listing shape without id.")
```
ovan: raise

app/models/listing_unit.rb
rap1ds: transaction_type_id, for now. We don't have yet listing_shape model or table. Would be very confusing to name it listing_shape_id, when it is actually transaction_type_id

app/services/listing_service/listing_shapes_api.md
rap1ds: Good catch. They are symbols, but the doc is outdated. I'll fix.

PR #1078, opened 2015-03-19
removed acts-as-url without enforcing translation through url-parameter

Gnito:
Affects transaction_type and categories

### DIFF
--- a/app/models/transaction_type.rb
+++ b/app/models/transaction_type.rb
@@ -45,7 +45,8 @@ class TransactionType < ActiveRecord::Base
   validates :price_per, inclusion: { in: %w(day),
     message: "%{value} is not valid" }, allow_nil: true
 
-  acts_as_url :url_source, scope: :community_id, sync_url: true, blacklist: %w{new all}
+  before_save :uniq_url
+
 
   # TODO this can be removed
   def self.columns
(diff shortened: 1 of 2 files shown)


app/models/transaction_type.rb
```diff
@@ -71,6 +72,25 @@ def default_translation_without_cache
...
+  def uniq_url
+    current_url = url_source.to_url
+
+    if new_record? || url != current_url
+      blacklist = ['new', 'all']
+      base_url = current_url
+      transaction_types = TransactionType.where(community_id: community_id)
+
+      i = 1
+      while blacklist.include?(current_url) || transaction_types.find { |tt| tt.url == current_url && tt.id != id }.present? do
```
rap1ds: This works nicely, no need to change anything.

Another super cool super immutable and functional way to do this would be lazy infinite range, something like this:

```ruby
reserved_names = ['domain', 'domain1', 'domain2']
(0..Float::INFINITY).lazy
  .map { |i| i > 0 ? "domain#{i}" : "domain" }
  .find { |name_candidate| !reserved_names.include? name_candidate }
=> "domain3"
```

But as I said, no need to change anything. Just wanted to experiment and try out how lazy infinite sequences work in Ruby. Seems to work ok.

rap1ds:
Good to go :+1:

PR #1080, opened 2015-03-20
Read units

rap1ds:
- [X] Read from 'units' table when posting new listing
- [x] Add 'unit_type' to listing
- [x] Save 'unit_type' to listing
- [x] Migration: Add unit_type to listings

Merge after #1075

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -595,8 +595,8 @@ class ListingsController < ApplicationController
       }
   end
 
-  def shapes_api
-    ListingService::API::Api.shapes
+  def listings_api
+    ListingService::API::Api
   end
 
   def valid_unit_type?(shape:, unit_type:)
@@ -613,7 +613,7 @@ class ListingsController < ApplicationController
       transaction_type_id: transaction_type_id
     }
 
-    shape_res = shapes_api.get(shape_find_opts)
+    shape_res = listings_api.shapes.get(shape_find_opts)
 
     if shape_res.success
       shape_res.data
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -183,8 +183,7 @@ class ListingsController < ApplicationController
     end
 
     params[:listing] = normalize_price_param(params[:listing]);
-
-    shape = get_shape(Maybe(params)[:transaction_type].to_i.or_else(nil))
+    shape = get_shape(Maybe(params)[:listing][:transaction_type_id].to_i.or_else(nil))
     unit_type = Maybe(shape[:units].first)[:type].or_else(nil)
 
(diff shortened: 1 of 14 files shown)


app/controllers/listings_controller.rb
```diff
@@ -581,4 +594,31 @@ def get_transaction_process(community: community, listing: listing)
         raise ArgumentError.new("Can not find transaction process: #{opts}") if process.nil?
       }
   end
+
+  def shapes_api
+    ListingService::API::Api.shapes
+  end
```
ovan: For global searching in dynamic language I noticed this pattern makes it difficult to find where an api method is used in the codebase. But if we instead do:
```ruby
def listings_api
  ListingsService::API::Api
end

listings_api.shapes.get(...)
```
then you can still do global search for shapes.get( to find usages.

app/controllers/listings_controller.rb
```diff
@@ -154,7 +154,8 @@ def new
       @numeric_field_ids = numeric_field_ids(@custom_field_questions)
 
       @listing.transaction_type = @current_community.transaction_types.find(params[:transaction_type])
-      logger.info "Category: #{@listing.category.inspect}"
+
+      shape = get_shape(Maybe(params)[:transaction_type].to_i.or_else(nil))
```
ovan: to_i ?

app/controllers/listings_controller.rb
```diff
@@ -179,8 +183,10 @@ def create
     end
 
     params[:listing] = normalize_price_param(params[:listing]);
+    shape = get_shape(Maybe(params)[:listing][:transaction_type_id].to_i.or_else(nil))
+    unit_type = Maybe(shape[:units].first)[:type].or_else(nil)
```
ovan: So shape can define multiple unit types but we always choose just the first one? What's the logic in this, why do we need :units to be an array?

app/controllers/listings_controller.rb
rap1ds: String to integer.

app/controllers/listings_controller.rb
rap1ds: This is only temporary. Currently we save the unit to `transaction_types` table and to the `units` table. Because `transaction_types` can have only one unit we save only the first one to the units table.

app/controllers/listings_controller.rb
```diff
@@ -581,4 +594,31 @@ def get_transaction_process(community: community, listing: listing)
         raise ArgumentError.new("Can not find transaction process: #{opts}") if process.nil?
       }
   end
+
+  def listings_api
+    ListingService::API::Api
+  end
+
+  def valid_unit_type?(shape:, unit_type:)
```
bingxie: Hi, @rap1ds seems like this method hasn't been used.

PR #1085, opened 2015-03-23
1/2 Read all data from listing (not from TransactionType)

rap1ds:
- [X] listing.transaction_type.price_field? -> listing.price
- [X] Migration: remove price from listings that do not have price_field in their transaction_type
- [X] Check from DB that there are no listing with price but without price_field and vice versa
- [x] Migration: remove quantity if unit_type = day
- [X] listing.transaction_type.shipping_enabled? -> shape[:shipping_enabled]
- [x] listing.transaction_type.name_tr_key -> listing.shape_name_tr_key
- [x] listing.transaction_type.action_button_tr_key -> listing.action_button_tr_key

- [x] Populate all transaction type names to listing
- [x] Populate all transaction type action button label to listing

Merge after #1090

### DIFF
--- a/app/services/payment_registration_guard.rb
+++ b/app/services/payment_registration_guard.rb
@@ -7,25 +7,23 @@ class PaymentRegistrationGuard
   end
 
   def requires_registration_before_accepting?
-    requires_registration?
+    @listing.price && @listing.price > 0 && @community.payments_in_use? && not_registered_already?
   end
 
   def requires_registration_before_posting?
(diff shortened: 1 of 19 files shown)


app/services/payment_registration_guard.rb
```diff
@@ -17,12 +17,19 @@ def requires_registration_before_posting?
   private
 
   def requires_registration?
```
sktoiva: Should query from shape when creating a new listing and from listing when buying.

PR #1086, opened 2015-03-23
Remove read from transaction_type_translations

Gnito:
- "uniq_url" creation happens now in listing service
- "display_name" and "action_button_label" (in transaction_type.rb) read data now from TranslationService

### DIFF
--- a/app/models/transaction_type.rb
+++ b/app/models/transaction_type.rb
@@ -60,11 +60,35 @@ class TransactionType < ActiveRecord::Base
   end
 
   def display_name(locale)
-    TranslationCache.new(self, :translations).translate(locale, :name)
+    result = TranslationService::API::Api.translations
+      .get(community_id, {
+        translation_keys: [name_tr_key],
+        locales: [locale.to_s],
+        fallback_locale: community.default_locale
+      })
+
+    Maybe(result[:data].first)[:translation].or_else(nil).tap { |translation|
+      if translation.nil?
+        error_msg = "translation missing for key #{name_tr_key} and locale #{locale}"
+        raise ArgumentError.new(error_msg)
+      end
+    }
   end
 
   def action_button_label(locale)
-    TranslationCache.new(self, :translations).translate(locale, :action_button_label)
+    result = TranslationService::API::Api.translations
+      .get(community_id, {
+        translation_keys: [action_button_tr_key],
+        locales: [locale.to_s],
+        fallback_locale: community.default_locale
+      })
+
+    Maybe(result[:data].first)[:translation].or_else(nil).tap { |translation|
+      if translation.nil?
+        error_msg = "translation missing for key #{action_button_tr_key} and locale #{locale}"
+        raise ArgumentError.new(error_msg)
+      end
+    }
   end
 
   def self.find_by_url_or_id(url_or_id)
--- a/app/models/transaction_type.rb
+++ b/app/models/transaction_type.rb
@@ -31,7 +31,8 @@ class TransactionType < ActiveRecord::Base
     :transaction_process_id,
     :shipping_enabled,
     :name_tr_key,
-    :action_button_tr_key
+    :action_button_tr_key,
+    :url
   )
 
   belongs_to :community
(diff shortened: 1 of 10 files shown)


spec/controllers/listings_controller_spec.rb
```diff
@@ -16,13 +16,26 @@ def create_transaction_type(community_id, type, process_id, translations = [], c
...
+    created_translations = TranslationService::API::Api.translations.create(community_id, [name_group])
+    name_tr_key = created_translations[:data].map { |translation| translation[:translation_key] }.first
+
     opts = defaults.merge(
       {
         shipping_enabled: false,
         transaction_process_id: process_id,
-        name_tr_key: 'something.here',
+        name_tr_key: name_tr_key,
         action_button_tr_key: 'something.here',
```
Gnito: These tests do not check action_button_label at the moment - so I didn't start to populate those yet.

app/services/listing_service/store/shape.rb
```diff
@@ -116,7 +117,9 @@ def to_tt_model_attributes(hash)
       raise ArgumentError.new("Unknown unit type #{unit_type}")
     end
 
-    model_hash.except(:units)
+    model_hash[:url] = uniq_url(hash[:url_source], hash[:community_id])
```
rap1ds: I would move this up into `create` and `update` methods.

The purpose of `to_tt_model_attributes` and `from_tt_model_attributes` was to map outdated model field names to updated API field names, so that:

`from_tt_model_attributes(to_tt_model_attributes(hash)) == hash`

 My idea is to remove both of these methods in the future. When we rename the `TransactionType` to `ListingShape` we also rename the `price_field` to `price_enabled` and in the end we remove the need for this method altogether.

app/models/transaction_type.rb
```diff
@@ -65,41 +63,36 @@ def to_param
...
+      .get(community_id, {
+        translation_keys: [name_tr_key],
+        locales: [locale.to_s],
+        fallback_locale: community.default_locale
+      })
+
+    Maybe(result[:data].first)[:translation].or_else(nil).tap { |translation|
+      if translation.nil?
+        error_msg = "translation missing for key #{name_tr_key} and locale #{locale}"
+        ApplicationHelper.send_error_notification(error_msg, "TransactionType")
```
rap1ds: Why not just raise?

app/models/transaction_type.rb
```diff
@@ -65,41 +63,36 @@ def to_param
...
+      .get(community_id, {
+        translation_keys: [action_button_tr_key],
+        locales: [locale.to_s],
+        fallback_locale: community.default_locale
+      })
+
+    Maybe(result[:data].first)[:translation].or_else(nil).tap { |translation|
+      if translation.nil?
+        error_msg = "translation missing for key #{action_button_tr_key} and locale #{locale}"
+        ApplicationHelper.send_error_notification(error_msg, "TransactionType")
```
rap1ds: Same here.

spec/services/listing_service/api/shapes_spec.rb
```diff
@@ -26,6 +26,7 @@
               { locale: "en", name: "Selling", action_button_label: "Buy" },
               { locale: "fi", name: "Myydään", action_button_label: "Osta" }
             ],
+            url_source: "Selling",
```
rap1ds: You could also add `url` to Shape entity and test:

`expect(shape[:url]).to eql("selling")`

Reference to `TransactionType` model will be removed from this spec in the future.

rap1ds:
Looks pretty good! A couple comments added.

Tests are not passing.

spec/services/listing_service/api/shapes_spec.rb
Gnito: It's pretty confusing variable name - since it's not url but just a word that can be added to url.
So, I thought that we add this to Shape entity, when we know how to use it over API.

app/services/listing_service/store/shape.rb
Gnito: Ah, ok. I thought this was the place to configure random 'opts' to tt_model_attributes. So, misunderstood the name of the function.

app/models/transaction_type.rb
Gnito: yeah, this changed at some point to a direction where or_else could be nil and I should have changed this error handling too.

PR #1087, opened 2015-03-24
2/2 Read all data from listing (not from TransactionType)

rap1ds:
- [X] Read: listing.transaction_process_id
- [X] Read: listing.shape_name_tr_key
- [X] Read: listing.action_button_tr_key
- [X] Project search for `listing.transaction_type.` does not find anything
- [X] Read price_quantity_placeholder via ListingService API

Merge after #1085

### DIFF
--- a/app/helpers/listings_helper.rb
+++ b/app/helpers/listings_helper.rb
@@ -24,9 +24,7 @@ module ListingsHelper
   end
 
   def listed_listing_title(listing)
-    shape_name = get_translations([listing.shape_name_tr_key], listing.communities.first).first
-
-    shape_name + ": #{listing.title}"
+    shape_name(listing) + ": #{listing.title}"
   end
 
   def transaction_type_url(listing, view)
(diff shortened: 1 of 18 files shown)


features/step_definitions/conversation_steps.rb
```diff
@@ -30,7 +30,7 @@ def create_transaction(community, listing, starter, message, payment_gateway = :
     starter: starter,
     conversation: build_conversation(community, listing, starter, message),
     payment_gateway: payment_gateway,
-    payment_process: TransactionProcess.find(listing.transaction_type.transaction_process_id).process == :preauthorize ? :preauthorize : :postpay,
+    payment_process: TransactionProcess.find(listing.transaction_process_id).process == :preauthorize ? :preauthorize : :postpay,
```
sktoiva: Here, more efficient would be:
```ruby
TransactionProcess.where(id: listing.transaction_process_id).pluck(:process).first == ...
```

sktoiva: Aaaand it's in test, so this is not important.

app/helpers/listings_helper.rb
```diff
@@ -115,4 +117,31 @@ def price_quantity_per_unit(listing)
...
+    result = TranslationService::API::Api.translations
+      .get(community.id, {
+        translation_keys: translation_keys,
+        locales: [locale],
+        fallback_locale: fallback
+           })
+
+    data = result[:data]
+
+    data.select { |t| t[:translation].nil? }.each { |t| raise ArgumentError.new("translation missing: #{t}") }
```
sktoiva: This could be removed.

app/helpers/listings_helper.rb
```diff
@@ -115,4 +117,31 @@ def price_quantity_per_unit(listing)
...
+    get_translations([listing.action_button_tr_key], listing.communities.first).first
+  end
+
+  # private
+  def get_translations(translation_keys, community)
+    locale = I18n.locale.to_s
+    fallback = community.default_locale.to_s
+
+    result = TranslationService::API::Api.translations
+      .get(community.id, {
```
sktoiva: This should probably be discussed through. Is the translations api as it should be?

PR #1090, opened 2015-03-24
Use tr service for tx type

Gnito:
transaction_types do not have translations for default_locale in every situation
I.e. if admin wants to change language from 'fi' to 'en', we don't want to break his/hers website for any moment. 

At some point we probably need to create a tool for changing default locale (and adding new locales) for transaction_type, categories, etc.

c372046 is old reviewed PR #1086 (the origin for this revert-revert)
ec2157c is new code (fix for relying on default_locale)

### DIFF
--- a/app/models/transaction_type.rb
+++ b/app/models/transaction_type.rb
@@ -31,7 +31,8 @@ class TransactionType < ActiveRecord::Base
     :transaction_process_id,
     :shipping_enabled,
     :name_tr_key,
-    :action_button_tr_key
+    :action_button_tr_key,
+    :url
   )
 
   belongs_to :community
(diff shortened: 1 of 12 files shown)


rap1ds:
I find `find_any_translation` a bit complex method and the naming is not the best one (it's not _any_ random translation, there's a clear order of preference). However, this is fine for now as a quick hotfix. Very soon we have to build some kind of a more general solution for using custom translations (i.e. hooking into Rails' I18n system, or something else)

rap1ds:
Actually, we can see that there's a big number of cache requests (translations, probably) when rendering the homepage. We should implement the cache before this goes back to master.

PR #1092, opened 2015-03-25
Improve Uservoice widget

gnomet:
By default the widget is disabled in OS version.
Collect NPS score with autoprompt.

### DIFF
--- a/app/views/admin/_left_hand_navigation.haml
+++ b/app/views/admin/_left_hand_navigation.haml
@@ -3,7 +3,7 @@
 - links_manage = links[2..5]
 - links_configure = links[6..21]
 
-= render :partial => "admin/uservoice_widget"
+= render :partial => "admin/uservoice_widget" if APP_CONFIG.uservoice_widget_url.present?
 
 .left-navi
 
--- a/app/views/admin/_uservoice_widget.html.erb
+++ b/app/views/admin/_uservoice_widget.html.erb
@@ -5,7 +5,7 @@ UserVoice=window.UserVoice||[];
   var uv=document.createElement('script');
   uv.type='text/javascript';
   uv.async=true;
-  uv.src='//widget.uservoice.com/5cNrCs3yhlo0sZ5xd9Btgw.js';
+  uv.src= '<%= APP_CONFIG.uservoice_widget_url %>';
   var s=document.getElementsByTagName('script')[0];
   s.parentNode.insertBefore(uv,s)
 })();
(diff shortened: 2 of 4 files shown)


app/views/admin/_left_hand_navigation.haml
```diff
@@ -3,7 +3,7 @@
 - links_manage = links[2..5]
 - links_configure = links[6..21]
 
-= render :partial => "admin/uservoice_widget"
+= render :partial => "admin/uservoice_widget" if APP_CONFIG.uservoice_widget_url.present?
```
sktoiva: In other parts of the app we pass this as locales from the controller in order to keep views separated from config files.

app/views/admin/_uservoice_widget.html.erb
```diff
@@ -5,7 +5,7 @@ UserVoice=window.UserVoice||[];
   var uv=document.createElement('script');
   uv.type='text/javascript';
   uv.async=true;
-  uv.src='//widget.uservoice.com/5cNrCs3yhlo0sZ5xd9Btgw.js';
+  uv.src= '<%= APP_CONFIG.uservoice_widget_url %>';
```
sktoiva: Same here.

app/views/admin/_uservoice_widget.html.erb
```diff
@@ -25,24 +25,19 @@ UserVoice.push(['set', {
 // Identify the user and pass traits
 // To enable, replace sample data with actual user traits and uncomment the line
 UserVoice.push(['identify', {
```
sktoiva: Actually, nearly all parameters here are new, so pass them as hash from controller. Views shouldn't know anything of services.

app/views/admin/_left_hand_navigation.haml
gnomet: Ok, thanks for mentioning. Had not picked that practice yet, but will modify these to match current best practice.

gnomet: hmm.. @sktoiva 
There are 23 files which have
`
render :partial => "admin/left_hand_navigation"
`
and some locals after that.
So pretty many places to put that uservoice specific stuff.. But I guess that doesn't change the principle and I should still put it everywhere?

And is there something else were doing wrong here.. Feels wrong to copy that everywhere..

Actually the thing we're probably doing wrong is to render the left hand navi separately for each page and not rendering that first and then the page contents..

gnomet:
Concluded in IRL chat that in this case let's keep the APP_CONFIG in the view instead of adding it to 30 different files separately. It's probably that admin UI renewal changes strucutre anyway so that the benefits of having app_config call only in controller would be small.

PR #1093, opened 2015-03-25
do not use community.locales, since we don't know if those exists

Gnito:
Do not use community.locales, since we don't know if those exists for transaction_type (case: locale is 'en', but only existing translation for transaction_type is 'fi')


ovan:
So after this are we back with the old behavior that we fallback to using translations even from locales that are not listed in community locales?

Gnito:
It's sad, but it's true. Though, there is still one addition (I'm not sure should it be removed): 
- we raise errors, if we don't find translations. (I guess, we have returned 'nil' before)

PR #1094, opened 2015-03-25
change indexes after translation cache

Gnito:


db/migrate/20150325164209_add_index_to_community_translations.rb
```diff
@@ -0,0 +1,14 @@
+class AddIndexToCommunityTranslations < ActiveRecord::Migration
+  def up
+    # cache fetches the all translations for given community
+    add_index :community_translations, :community_id
+    # cache makes these obsolete
```
ovan: Does obsolete mean we no longer do SQL queries based upon these columns, or that we do it so seldom that optimizing for it it not a priority? If the first case then go ahead and remove but if it's only the second case then I'd keep the indexes.

Gnito: Yeah, we just fetch all translations for the community - no exceptions atm.

PR #1095, opened 2015-03-25
New optional PayPal onboarding flow

ovan:
Add support for new PayPal onboarding flow under the hood but don't enable in the UI yet. User facing part + per community support is coming later.

### DIFF
--- a/app/services/paypal_service/store/paypal_account.rb
+++ b/app/services/paypal_service/store/paypal_account.rb
@@ -9,10 +9,10 @@ module PaypalService::Store::PaypalAccount
     # Mandatory
     [:community_id, :mandatory, :fixnum],
     [:person_id, :optional, :string],
-    [:order_permission_request_token, :string],
     [:order_permission_paypal_username_to, :mandatory, :string],
 
     # Optional
+    [:order_permission_request_token, :string],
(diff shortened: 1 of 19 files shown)


app/services/paypal_service/store/paypal_account.rb
```diff
@@ -9,7 +9,7 @@ module PaypalService::Store::PaypalAccount
     # Mandatory
     [:community_id, :mandatory, :fixnum],
     [:person_id, :optional, :string],
-    [:order_permission_request_token, :mandatory, :string],
+    [:order_permission_request_token, :string],
```
sktoiva: Under comment 'mandatory', should be moved.

db/migrate/20150324112658_add_onboarding_fields_to_order_permissions.rb
```diff
@@ -0,0 +1,13 @@
+class AddOnboardingFieldsToOrderPermissions < ActiveRecord::Migration
```
sktoiva: We might want to prefer using :after clauses to ensure that our schema.rb stays the same.

ovan: The fields are added at the end so no need in this case.

PR #1097, opened 2015-03-26
View helper and controller method to ease the use of TranslationService

rap1ds:
- [X] Easily translate keys
- [X] Per request cache


ovan:
Looks good. Can you explain what's the mechanism through which the helper, view and the controller end up sharing the same instance variable @community_translations instead of those being two different caches?

PR #1100, opened 2015-03-26
Remove transaction type association from community

sktoiva:
- [X] Remove community has many transaction_types relationship
- [X] Where transaction_type is needed, fetch it via ListingService API
- [X] Add sort_priority to ListingShape
- [X] Populate sort_priority
- [X] Fix test code
- [x] Test new marketplace creation on staging

### DIFF
--- a/app/models/community.rb
+++ b/app/models/community.rb
@@ -545,15 +545,6 @@ class Community < ActiveRecord::Base
     category_hash
   end
 
-  # same as available_categorization_values but returns the models instead of just values
-  def available_categorizations
-    values = {}
-    values["category"] = top_level_categories
-    values["subcategory"] = subcategories
-    values["shape"] = listing_shapes
-    return values
-  end
-
   def main_categories
     top_level_categories
   end
@@ -606,10 +597,6 @@ class Community < ActiveRecord::Base
     payment_gateway.present? && payment_gateway.type == "BraintreePaymentGateway"
   end
 
-  def price_in_use?
-    listing_shapes.any? { |s| s[:price_enabled] }
-  end
-
   # Return either minimum price defined by this community or the absolute
   # platform default minimum price.
   def absolute_minimum_price(currency)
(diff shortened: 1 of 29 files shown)


app/models/community.rb
```diff
@@ -664,4 +645,10 @@ def images_processing?
   def initialize_settings
     update_attribute(:settings,{"locales"=>[APP_CONFIG.default_locale]}) if self.settings.blank?
   end
+
+  def listing_shapes
+    @listing_shapes ||= ListingService::API::Api.shapes.get(community_id: id).maybe.or_else(nil).tap { |shapes|
+      raise ArgumentError.new("Can not find any listing shapes for community #{id}") if shapes.nil?
+    }
+  end
```
ovan: This shouldn't happen inside a model object. Move the logic to controller?

app/models/listing.rb
```diff
@@ -222,10 +222,13 @@ def self.find_with(params, current_user=nil, current_community=nil, per_page=100
     if params[:share_type].present?
       direction = params[:share_type]
       transaction_type_direction_map = ListingShapeHelper.transaction_types_to_direction_map(current_community) # deprecated
+
+      all_shapes = ListingService::API::Api.shapes.get(community_id: current_community.id).maybe.or_else([])
```
ovan: These could be passed in from the controller calling this to avoid calling services inside model.

app/services/listing_service/store/shape.rb
```diff
@@ -28,6 +29,8 @@ module ListingService::Store::Shape
     [:translations, :array, :optional], # TODO Only temporary
     [:units, :array, :mandatory],
     [:shipping_enabled, :bool, :mandatory],
+    [:url, :string, :mandatory],
```
ovan: Wasn't this renamed somewhere to name?

ovan: Looks like it's stored as name so it would make sense to call it name throughout the chain.

app/view_utils/listing_shape_helper.rb
```diff
@@ -19,20 +19,22 @@ def transaction_type_id_to_direction(transaction_type_id, community)
...
+    shapes = ListingService::API::Api.shapes.get(community_id: community.id).maybe.or_else([])
+
+    direction_tuples = shapes.map { |shape|
+        direction = process_res
+          .maybe
+          .map { |processes| processes.find { |p| p[:id] == shape[:transaction_process_id] } }
+          .map { |process| process[:author_is_seller] ? "offer" : "request" }
+          .or_else(nil)
+          .tap { |process|
+            raise ArgumentError.new("Can not find transaction process for community #{community.id}, transaction type #{tt.id}") if process.nil?
```
ovan: tt.id not defined

PR #1103, opened 2015-03-27
Remove TestHelper injection

sktoiva:


rap1ds:
If tests pass and this works locally with zeus, then good to go :+1:

PR #1111, opened 2015-03-31
Read by listing shape id

rap1ds:
- [x] Return listing shape ID via listing shapes API
- [x] Find shape by listing shape id


rap1ds:
Closed, see #1113 instead

PR #1113, opened 2015-04-01
Read listing shape

rap1ds:
- [X] Read everything from listing shapes table
- [X] Change to_bool -> bool because we don't have null values anymore
- [X] Find listing shape by ID
- [x] Make sure the shape url / name naming is correct
- [x] listing_units.transaction_type_id needs index
- [x] listing_units.listing_shape_id needs index
- [x] listing_shapes.transaction_type_id needs index

### DIFF
--- a/app/services/listing_service/store/shape.rb
+++ b/app/services/listing_service/store/shape.rb
@@ -206,7 +206,7 @@ module ListingService::Store::Shape
   end
 
   def find_shape_models(community_id:)
-    ListingShape.where(community_id: community_id)
+    ListingShape.where(community_id: community_id).order(:sort_priority)
   end
 
   def uniq_name(name_source, community_id)
--- a/app/services/listing_service/store/shape.rb
+++ b/app/services/listing_service/store/shape.rb
@@ -18,8 +18,7 @@ module ListingService::Store::Shape
   )
 
   Shape = EntityUtils.define_builder(
-    # TODO Currently we don't have Shape model, i.e. we don't have Shape id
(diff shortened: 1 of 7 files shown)


app/services/listing_service/store/shape.rb
```diff
@@ -77,6 +76,7 @@ def create(community_id:, opts:)
     translations = opts[:translations] # Skip data type and validation, because this is temporary
 
     tt_model = nil
```
rap1ds: Remove old unused `tt_model`

app/services/listing_service/store/shape.rb
```diff
@@ -77,6 +76,7 @@ def create(community_id:, opts:)
...
+    shape_model = nil
 
     ActiveRecord::Base.transaction do
```
rap1ds: Test if we could return `shape_model` from the transaction block.

app/services/listing_service/store/shape.rb
```diff
@@ -97,18 +97,24 @@ def create(community_id:, opts:)
...
+    shape_model = find_shape_model(
       community_id: community_id,
       transaction_type_id: transaction_type_id,
       listing_shape_id: listing_shape_id)
 
-    return nil if tt_model.nil?
+    return nil if shape_model.nil?
 
-    shape_model = ListingShape.where(transaction_type_id: tt_model.id).first
+    transaction_type_id ||= transaction_type_id
```
rap1ds: Should be

```
transaction_type_id ||= shape_model.transaction_type_id
```

PR #1125, opened 2015-04-03
Shape ui create new

ovan:


app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -10,28 +10,88 @@ class Admin::ListingShapesController < ApplicationController
...
+
     render("index",
            locals: {
              selected_left_navi_link: LISTING_SHAPES_NAVI_LINK,
+             templates: templates(processes),
              listing_shapes: all_shapes(@current_community.id)})
   end
 
+  def new
+    processes = TransactionService::API::Api.processes.get(community_id: @current_community.id)[:data]
```
ovan: This should have a helper method that wraps calls to service the same way as we have get_shape, etc.

app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -50,26 +110,142 @@ def update
...
+  def templates(transaction_processes)
+    request_process_available = transaction_processes.any? { |tp| tp[:author_is_seller] == false }
+    preauthorize_process_available = transaction_processes.any? { |tp| tp[:process] == :preauthorize }
+
+    template_defaults.reject { |tmpl|
+      tmpl[:key] == :requesting && !request_process_available
+    }
+  end
+
+  def template_defaults
```
ovan: I think these should go to view_utils instead of controller.

app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -84,8 +260,15 @@ def parse_unit(type)
...
-    trs = TranslationServiceHelper.to_key_locale_hash(translations)
+  def to_form_data(shape, available_locs)
+    action_button_translations = available_locs.map { |(loc_name, loc_key)|
+      [loc_key, t(shape[:action_button_tr_key], locale: loc_key)]
+    }.to_h
+
+
+    name_translations = available_locs.map { |(loc_name, loc_key)|
+      [loc_key, t(shape[:name_tr_key], locale: loc_key)]
+    }.to_h
```
ovan: Helper method would make this read better and avoid dup:
```ruby
action_button_translation = make_translations(shape[:action_button_tr_key], available_locs)
name_translation = make_translations(shape[:name_tr_key], available_locs)
```

app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -10,28 +10,88 @@ class Admin::ListingShapesController < ApplicationController
...
+      redirect_to action: :index
+    else
+      flash[:error] = t("admin.listing_shapes.new.create_failure")
+      # TODO RENDER SOMETHING
+    end
+
   end
 
   def update
+    processes = TransactionService::API::Api.processes.get(community_id: @current_community.id)[:data]
```
ovan: wrapper method

app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -10,28 +10,88 @@ class Admin::ListingShapesController < ApplicationController
...
     :units,
   ).with_validations {
     validates :name, presence: true
     validates :action_button_label, presence: true
     validates :shipping_enabled, inclusion: { in: [true, false] }
+    # TODO Add validations
   }
 
   def index
+    processes = TransactionService::API::Api.processes.get(community_id: @current_community.id)[:data]
```
ovan: wrapper method

app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -42,7 +102,7 @@ def update
     end
 
     update_result =
-      update_translations(shape, shape_form)
+      update_translations(@current_community, shape_form)
```
ovan: Shouldn't this take @current_community.id instead of model object? Also, looking at actual method it seems to expect shape as first parameter?

app/controllers/admin/listing_shapes_controller.rb
ovan: Actually, if this is moved then `templates(transaction_process)` and `valid_template?(template_key, processes)` be moved too. This feels like good abstraction to have template definitions and methods that operate on those defined in a separate place.

PR #1128, opened 2015-04-04
Don't rely on transaction_type_id

rap1ds:
Merge after #1127

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -59,14 +59,16 @@ class HomepageController < ApplicationController
       find_listings(params, listings_per_page, compact_filter_params)
     end
 
+    shape_name_map = all_shapes.map { |s| [s[:id], s[:name]]}.to_h
+
     if request.xhr? # checks if AJAX request
       if @view_type == "grid" then
         render :partial => "grid_item", :collection => @listings, :as => :listing
       else
-        render :partial => "list_item", :collection => @listings, :as => :listing
+        render :partial => "list_item", :collection => @listings, :as => :listing, locals: { shape_name_map: shape_name_map }
       end
     else
-      render locals: { shapes: all_shapes, selected_shape: selected_shape }
+      render locals: { shapes: all_shapes, selected_shape: selected_shape, shape_name_map: shape_name_map }
     end
   end
 
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -37,10 +37,26 @@ class HomepageController < ApplicationController
 
     listings_per_page = APP_CONFIG.grid_listings_limit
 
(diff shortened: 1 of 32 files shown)


app/controllers/homepage_controller.rb
```diff
@@ -37,20 +37,38 @@ def index
 
     listings_per_page = APP_CONFIG.grid_listings_limit
 
+    filter_params = {}
+
+    # :transaction_type was renamed to :order_type
```
ovan: The risk here is we haven't yet locked the order_type term so this might deprecate next week when the design of shape ui progresses and we define the term to go forward with. Order type is the best candidate for now but not yet decided.

app/controllers/homepage_controller.rb
```diff
@@ -37,20 +37,38 @@ def index
...
+    params[:order_type] ||= params[:transaction_type]
+    params[:transaction_type] = nil
+    order_type_params = params[:order_type]
+
+    all_shapes = shapes.get(community_id: @current_community.id)[:data]
+    selected_shape = all_shapes.find { |s| s[:name] == order_type_params }
+
+    filter_params[:listing_shape] = Maybe(selected_shape)[:id].or_else(nil)
+
+    compact_filter_params = HashUtils.compact(filter_params)
```
ovan: This action is overly long already. Looks like there's couple of easy extract methods to parse parameters, to get shapes, and to build filter params from shapes.

app/controllers/listings_controller.rb
```diff
@@ -51,16 +51,17 @@ def index
         page =  params[:page] || 1
         per_page = params[:per_page] || 50
 
+        all_shapes = get_shapes()
+        all_processes = get_processes()
```
ovan: Are these two locals used outside the call to create direction_map? If not, I'd inline to help reader to see their only used for one purpose.

rap1ds: `all_shapes` is used, line 62

app/utils/array_utils.rb
```diff
@@ -46,4 +46,11 @@ def each_slice_columns(xs, columns, &block)
   def trim(xs)
     xs.drop_while { |x| x.blank? }.reverse.drop_while { |x| x.blank? }.reverse
   end
+
+  def zip_by(a, b, &block)
+    a.map { |a_elem|
+      b_found = b.find { |b_elem| block.call(a_elem, b_elem) }
```
ovan: When the zipping condition is arbitrary there's no guarantee that only one element in array b matches with an array a element. I think I would expect this to handle that case too. You could do this by using reduce instead of map and find_all  + concatenation instead of find.

Another case is what if a b element matches with multiple a elements. Should it be duplicated or zipped into the first place where it matches? I'd say put it into first place and have no dups.

PR #1132, opened 2015-04-06
Remove old map redirect

rap1ds:
- It actually causes a double renderer error, since there's no return after the redirect


ovan:
ok

PR #1135, opened 2015-04-07
Improve ListingShape validations:

rap1ds:
- [X] Custom unit has to have translation
- [X] Name must be unique (in community scope)

### DIFF
--- a/app/services/listing_service/store/shape.rb
+++ b/app/services/listing_service/store/shape.rb
@@ -39,11 +39,15 @@ module ListingService::Store::Shape
     [:sort_priority, :fixnum]
   )
 
-  Unit = EntityUtils.define_builder(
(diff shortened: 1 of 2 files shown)


app/models/listing_shape.rb
```diff
@@ -35,6 +35,8 @@ class ListingShape < ActiveRecord::Base
     :price_quantity_placeholder,
   )
 
+  validates_uniqueness_of :name, :scope => :community_id
```
ovan: Is this validated with unique index? This type of model validation if quite slow in my understanding so it might be good enough to leave it to db as long as we're making serious effort to generate a unique name when saving shape.

rap1ds: Good suggestion! This is not validated with unique index. I guess the model selects all shapes by `community_id`.

PR #1136, opened 2015-04-07
Add and save quantity selector type to listing_units and listings

rap1ds:


rap1ds:
Change column name to quantity_selector

PR #1141, opened 2015-04-09
Units ui: (2/2) email receipts

Gnito:
Added quantity to email receipts

### DIFF
--- a/app/mailers/transaction_mailer.rb
+++ b/app/mailers/transaction_mailer.rb
@@ -84,8 +84,8 @@ class TransactionMailer < ActionMailer::Base
     service_fee = payment.total_commission
     you_get = payment.seller_gets
 
-    listing = payment.transaction.listing
-    unit_type = translate_quantity_unit(listing.unit_type)
+    transaction = payment.transaction
+    unit_type = translate_quantity_unit(transaction.unit_type)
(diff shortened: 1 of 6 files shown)


app/mailers/transaction_mailer.rb
```diff
@@ -138,6 +155,9 @@ def paypal_new_payment(transaction, seller_model = nil, buyer_model = nil, commu
 
     you_get = payment_total - service_fee - gateway_fee
 
+    listing = Listing.where(transaction[:listing_id]).first
```
sktoiva: This isn't used anywhere?

Gnito: ah, true - not anymore.

app/mailers/transaction_mailer.rb
```diff
@@ -83,6 +84,10 @@ def braintree_new_payment(payment, community)
     service_fee = payment.total_commission
     you_get = payment.seller_gets
 
+    listing = payment.transaction.listing
+    unit_type = translate_quantity_unit(listing.unit_type)
```
ovan: It's a bad sign that we need to refer to listing here. We must record unit type to transaction because:
1. This mailer is transaction mailer and logically part of "transaction service", so reaching out to other services for data is a no no.
2. The unit type might already be different for listing at this point than when the transaction was started.

PR #1144, opened 2015-04-10
locals as entity

Gnito:


app/controllers/transactions_controller.rb
```diff
@@ -36,22 +36,41 @@ def show
 
     MarketplaceService::Transaction::Command.mark_as_seen_by_current(params[:id], @current_user.id)
 
-    show_total_price = tx[:payment_process] != :none
-    total_price_label = (tx[:payment_process] != :preauthorize) ? t("transactions.price") : t("transactions.total")
-    unit_type = listing[:unit_type].present? ? translate_quantity_unit(tx[:unit_type]) : nil
+    price_break_down_locals =
```
sktoiva: This if-else conditional should be in it's own method. In my opinion, it's difficult to read code where there is a 22 line conditional statement with return value stored in a variable. 

Refactoring this into a own private method would even render the need for this local variable unnecessary.

app/controllers/transactions_controller.rb
```diff
@@ -36,22 +36,41 @@ def show
...
+      if tx[:payment_process] == :none && tx[:listing_price].cents == 0
+        nil
+      else
```
sktoiva: And unless you  want to be explicit:

```ruby
if tx[:payment_process] != :none || tx[:listing_price].cents != 0
```
would remove the else branch.

app/controllers/transactions_controller.rb
Gnito: true dat

app/controllers/transactions_controller.rb
Gnito: Yeah, I wanted to be explicit

PR #1145, opened 2015-04-13
Per community I18n backend

rap1ds:
Add a new translation backend, which makes it possible to scope translations per community.

ApplicationController sets the correct scope and sets the translations, and after that we can use the `t` method to fetch the translations.

Tasks:

- [X] Implement CommunityBackend for translations
- [X] Set correct community for the translation backend in ApplicationController
- [X] Implement `with_locale { ... }` helper method, for temporary changing locale and community in mail views
- [X] Use `with_locale` in each mail view
- [X] Remove per request translation cache, that was stored as a controller instance variable
- [X] Remove some translation helper code and let the translation backend handle that logic
- [x] Periodically clean up translation store (it keeps growing when ever we add new community translations there)

After #1148

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -72,7 +72,7 @@ class ApplicationController < ActionController::Base
     # Load translations from TranslationService
     community_backend = I18n::Backend::CommunityBackend.instance
     m_community.map(&:id).each { |community_id|
-      community_backend.set_community!(community_id)
+      community_backend.community_id = community_id
       community_translations = TranslationService::API::Api.translations.get(community_id)[:data]
       TranslationServiceHelper.community_translations_for_i18n_backend(community_translations).each { |locale, data|
         # Store community translations to I18n backend.
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -68,11 +68,12 @@ class ApplicationController < ActionController::Base
     m_community = Maybe(@current_community)
     community_locales = m_community.locales.or_else([])
     community_default_locale = m_community.default_locale.or_else("en")
+    community_id = m_community[:id].or_else(nil)
+    community_backend = I18n::Backend::CommunityBackend.instance
 
     # Load translations from TranslationService
-    community_backend = I18n::Backend::CommunityBackend.instance
-    m_community.map(&:id).each { |community_id|
-      community_backend.community_id = community_id
+    if community_id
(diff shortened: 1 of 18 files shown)


app/controllers/application_controller.rb
```diff
@@ -68,7 +68,22 @@ def set_locale
...
+    community_backend = I18n::Backend::CommunityBackend.instance
+    m_community.map(&:id).each { |community_id|
+      community_backend.community_id = community_id
+      community_translations = TranslationService::API::Api.translations.get(community_id)[:data]
+      TranslationServiceHelper.community_translations_for_i18n_backend(community_translations).each { |locale, data|
+        # Store community translations to I18n backend.
+        #
+        # Since the data in data hash is already flatten, we don't want to
+        # escape the separators (. dots) in the key
+        community_backend.store_translations(locale, data, escape: false)
```
rap1ds: Storing new translations to the backend per each request will grow the translation hash without ever being cleaned up. Potentially, this may be an issue.

app/utils/mail_utils.rb
```diff
@@ -25,6 +32,42 @@ def premailer(message)
...
   end
 
+  # private
+
+  def set_locale(new_locale, &block)
+    old_locale = I18n.locale
+
+    if old_locale.to_sym != new_locale.to_sym
+      I18n.locale = new_locale
+      # TODO store_translations here
```
ovan: already done

app/utils/mail_utils.rb
```diff
@@ -12,10 +12,17 @@ def set_up_urls(recipient, community, ref="email")
...
       @recipient = recipient
       @url_params[:locale] = @recipient.locale
-      set_locale @recipient.locale
     end
   end
 
+  def with_locale(recipient_locale, community_id = nil, &block)
+    set_locale(recipient_locale) {
+      set_community(community_id) {
+        block.call
```
ovan: `set_locale` and `set_community` should handle cleanup on errors.

app/controllers/application_controller.rb
ovan: The last commit fixes this by only keeping translations for a single community at a time. There's an override exception case for not clearing the translations when we synchronously send email and switch community in the middle of a request.

app/utils/mail_utils.rb
rap1ds: Done

PR #1147, opened 2015-04-13
Feature flags

ovan:
Generic feature flags feature to selectively enable in development features for specific communities.


app/controllers/application_controller.rb
```diff
@@ -388,4 +391,47 @@ def force_ssl
...
+  def fetch_feature_flags
+    # Fetch from service and session
+    flags_from_service = FeatureFlagService::API::Api.features.get(community_id: @current_community.id).maybe[:features].or_else(Set.new)
+    temp_flags_from_session = Maybe(session)[:feature_flags].or_else(Set.new)
+
+    # Fetch from params
+    temp_flags_from_params =
+      Maybe(@current_user)
+      .select { |user| user.is_admin? }
+      .flat_map { Maybe(params)[:enable_feature].map { |feature| [feature.to_sym].to_set } }
```
ovan: What about :disable_feature parameter to undo this?

app/controllers/application_controller.rb
```diff
@@ -388,4 +391,47 @@ def force_ssl
       redirect_to("https://#{request.host_with_port}#{request.fullpath}") unless request.ssl? || ( request.headers["HTTP_VIA"] && request.headers["HTTP_VIA"].include?("sharetribe_proxy")) || request.fullpath == "/robots.txt"
     end
   end
+
+  def feature_flags
+    @feature_flags ||= fetch_feature_flags
+  end
+
+  def fetch_feature_flags
```
ovan: This method could use a rspec test.

app/controllers/application_controller.rb
```diff
@@ -388,4 +391,48 @@ def force_ssl
       redirect_to("https://#{request.host_with_port}#{request.fullpath}") unless request.ssl? || ( request.headers["HTTP_VIA"] && request.headers["HTTP_VIA"].include?("sharetribe_proxy")) || request.fullpath == "/robots.txt"
     end
   end
+
+  def feature_flags
```
bingxie: I am confused why we have the same method in here and FeatureFlagHelper?
We already included FeatureFlagHelper.

PR #1156, opened 2015-04-15
Optmize listing edit/create handling of custom fields

ovan:


bingxie:
Cool, I like these gems!

PR #1161, opened 2015-04-18
EntityUtils.define_builder improvements

rap1ds:
This PR aims to address the following entity builder shortcomings:

### Invalid entities always raise an error

**The problem:** If the data given to the builder is invalid, the entity builder always raises an error. This is ok if we are only using entities to prevent programming errors, but if we use entity builder e.g. for validating user input, then we want to show an informative error message instead of crashing the whole app with an exception.

**The solution:** Add `result: true` option to the entity builder call. That will wrap the entity in a Result object

**Usage:**

```ruby
res = MyEntity.call(my_entity_data, result: true)
```

### Can not nest entitites

**The problem:** There are situations when we would like to nest the entities. For example, the ListingService Shape API `#create` method takes the following hash:

```ruby
{ transaction_process_id: 123
, price_enabled: true
, ...
, units: 
  [ { type: :day }
  , { type: :custom
    , tr_key: "1231-1231-123-abd"
    }
  ]
}
```

Clearly, there are two types of entities: The **listing shape** entity and inside it a collection of **units**

**The solution:** This PR adds two new options for an entity field spec: `collection` and `entity`. `collection` can be used for a collection of entities (i.e. an array of entities, like the units in the above example) and `entity` for a single hash entity.

**Usage:**

```ruby
ListingShape = EntityUtils.define_builder(
  [:transaction_process_id, :fixnum, :mandatory],
  [:price_enabled, :bool, :mandatory],
  [:units, :optional, collection: [
    [:type, one_of: [:day, :hour, ... , :custom]],
    [:tr_key, :string, :optional]
  ]
)
```

You can also define builders for shape and units separately:

```ruby
Unit = EntityUtils.define_builder(
  [:type, one_of: [:day, :hour, ... , :custom]],
  [:tr_key, :string, :optional]
)

ListingShape = EntityUtils.define_builder(
  [:transaction_process_id, :fixnum, :mandatory],
  [:price_enabled, :bool, :mandatory],
  [:units, :optional, collection: Unit]
)
```

For a single hash entity, use `entity`:

```ruby
Name = EntityUtils.define_builder(
  [:first, :string, :mandatory],
  [:middle, :string, :optional],
  [:last, :string, :mandatory]
)

Person = EntityUtils.define_builder(
  [:name, :mandatory, entity: Name]
)
```

### Better error message

**The problem:** Error message doesn't contain the name of the field that failed the validation in a format that is easy to use (the name is in the message string). There's no error code, only the error message. This makes it impossible to translate the error message.

**The solution:** The error message contains now a full path to the items that didn't pass the validation. In addition, the error object contains error code, which makes it possible to show translated error message to the user.

**Usage:**

```ruby
TestBuilder = EntityUtils.define_builder(
  [:root, entity: [
      [:numbers, collection: [
          [:display, collection: [
              [:value, :mandatory]]]]]]])

TestBuilder.call({root: {numbers: [{display: [{value: 1}, {}]}]}}, result: true).data.first
# => {:field=>"root.numbers[0].display[1].value",
#  :code=>:mandatory,
#  :msg=>"Missing mandatory value."}
```

### DIFF
--- a/app/utils/entity_utils.rb
+++ b/app/utils/entity_utils.rb
@@ -254,10 +254,7 @@ module EntityUtils
 
   def transform_and_validate(fields, input)
     output = transform_all(fields, input)
-    errors = validate_all(fields, output).map { |err|
-      err[:msg] = "#{err[:field]}: #{err[:msg]}"
-      err
-    }
+    errors = validate_all(fields, output)
 
     {value: output, errors: errors}
   end
@@ -322,7 +319,10 @@ module EntityUtils
       result = transform_and_validate(fields, data)
 
       if !result[:errors].empty?
-        msg = result[:errors].map { |error| error[:msg] }.join(", ")
+        msg = result[:errors].map { |error|
+          "#{error[:field]}: #{error[:msg]}"
+        }.join(", ")
+
         if opts[:result]
           Result::Error.new(msg, result[:errors])
         else
--- a/app/utils/entity_utils.rb
+++ b/app/utils/entity_utils.rb
@@ -145,7 +145,7 @@ module EntityUtils
     transform_with: -> (transformer, v) { transformer.call(v) }
   }
 
-  def validator_or_transformer(k)
+  def categorice_spec(k)
     if (VALIDATORS.keys.include?(k))
       :validators
     elsif (TRANSFORMERS.keys.include?(k))
@@ -165,7 +165,7 @@ module EntityUtils
     parsed_spec = s.zip([nil].cycle)
       .to_h
       .merge(opts)
-      .group_by { |(name, param)| validator_or_transformer(name) }
+      .group_by { |(name, param)| categorice_spec(name) }
 
     parsed_spec[:validators] =
       (parsed_spec[:validators] || [])
@@ -173,20 +173,19 @@ module EntityUtils
     parsed_spec[:transformers] =
       (parsed_spec[:transformers] || [])
       .map { |(name, param)| TRANSFORMERS[name].curry().call(param) }
-    parsed_spec[:collection] =
-      if opts[:collection].is_a? Proc
-        opts[:collection].call(nil, specs_only: true)
-      else
-        parse_specs(opts[:collection] || [])
-      end
-    parsed_spec[:entity] =
-      if opts[:entity].is_a? Proc
-        opts[:entity].call(nil, specs_only: true)
-      else
-        parse_specs(opts[:entity] || [])
-      end
+
+    parsed_spec[:collection] = parse_nested_specs(opts[:collection])
+    parsed_spec[:entity] = parse_nested_specs(opts[:entity])
 
     parsed_spec
+  end
+
+  def parse_nested_specs(specs)
+    if specs.is_a? Proc
+      specs.call(nil, specs_only: true)
+    else
+      parse_specs(specs || [])
+    end
   end
 
   def parse_specs(specs)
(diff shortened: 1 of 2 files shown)


app/utils/entity_utils.rb
```diff
@@ -36,81 +36,79 @@ def model_to_hash(model)
...
-        "#{field}: Value must be boolean true or false. Was: #{v} (#{v.class.name})."
+        {code: :bool, msg: "Value must be boolean true or false. Was: #{v} (#{v.class.name})." }
       end
     },
-    validate_with: -> (validator, v, field) {
-      unless (validator.call(v))
-        "#{field}: Custom validation failed. Was: #{v}."
-      end
+    validate_with: -> (validator, v, _) {
+      validator.call(v)
```
ovan: The validator return value type change is not backwards compatible. If there's any existing code that uses `validate_with`it will break. Do we have these?

app/utils/entity_utils.rb
```diff
@@ -145,11 +143,15 @@ def model_to_hash(model)
     transform_with: -> (transformer, v) { transformer.call(v) }
   }
 
-  def validator_or_transformer(k)
+  def categorice_spec(k)
```
ovan: My suggestion for renaming is `categorice_spec` -> `spec_category`. You group by spec category and you get the category of a spec by calling this function.

app/utils/entity_utils.rb
```diff
@@ -184,26 +197,64 @@ def parse_specs(specs)
...
+  def validate_all(fields, input)
+    fields.reduce([]) do |errs, (name, spec)|
+      errors = validate(spec[:validators], input[name], name).map { |err|
+        err[:field] = "#{name.to_s}"
+        err
+      }
+
+      nested_errors =
+        if spec[:collection].present?
+          input[name].each_with_index.flat_map { |v, i|
```
ovan: Can you come up with a way to use reduce and mutations instead of flat_map and map that create temporary collections in the way? This code gets run a lot so one of the design goals of the original version was to favor one pass processing and mutations where possible. This is of course much more complex scenario to handle.

app/utils/entity_utils.rb
```diff
@@ -255,17 +306,36 @@ def transform_and_validate(fields, input)
...
+          loc = caller_locations(2, 1).first
+          raise(ArgumentError, "Error(s) in #{loc}: #{msg}")
+        end
+      else
+        if opts[:result]
+          Result::Success.new(result[:value])
+        else
+          result[:value]
+        end
+      end
```
ovan: Now that we have two different ways to run the tranformation and validation keeping the lambda as return value type makes the API a bit convoluted.

Instead, we could return an object which would have an API like:
```ruby
Shape = EntityUtils.define_builder(...);

# This is the default behavior, raise error if validation fails
Shape.build({...})

# This is the new behavior, return Result object
Shape.validate({...})

# This calls build and, afaik, should keep the old call syntax Shape[...] working too.
Shape.call({..})

Shape.specs # attr_reader to get the parsed specs
```

app/utils/entity_utils.rb
rap1ds: We have 2 in entity utils specs and they are changed.

ovan: :+1:

app/utils/entity_utils.rb
rap1ds: Done

ovan:
Yeah, great stuff! Go ahead and merge (check the comment + travis failure). I might like to play with some optimization ideas later to see if anything really affects anything but I'm sure it's fast enough already for production.

PR #1163, opened 2015-04-19
Fix typo?

mparramont:
:)


rap1ds:
The instruction is right. Leave out the RAILS_ENV=production if you want to run it in dev mode.

PR #1165, opened 2015-04-20
Add missing indexes detected in slow query log

ovan:
By analyzing the slow query / no index logs from MySQL these missing indexes came up in top slow queries.

**Note!** It's important to index on community_id when creating new tables because we almost always scope queries in services per community to protect for data leakage.


rap1ds:
:+1:

PR #1167, opened 2015-04-20
Shipping per additional items

Gnito:
If travis likes this, this could be reviewed.

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -188,7 +188,8 @@ class ListingsController < ApplicationController
 
         render :partial => "listings/form/form_content", locals: commission(@current_community, process).merge(
                  shape: shape,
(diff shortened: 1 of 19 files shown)


app/controllers/listings_controller.rb
```diff
@@ -272,10 +273,19 @@ def edit
     shape = get_shape(@listing.listing_shape_id)
     process = get_transaction_process(community_id: @current_community.id, transaction_process_id: shape[:transaction_process_id])
     unit_options = ListingViewUtils.unit_options(shape[:units], unit_from_listing(@listing))
+    shipping_price_additional =
+      if @listing.shipping_price_additional
```
sktoiva: I'd prefer, if these assignments with conditionals would be in their own methods. Makes 'em easier to read.

PR #1168, opened 2015-04-21
Shape UI improvements

rap1ds:
- Use entity instead of form
- Code reorganizing


rap1ds:
Review comments:

- Disabled checkbox label should be gray

app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -5,209 +5,265 @@ class Admin::ListingShapesController < ApplicationController
...
+             templates: templates,
+             category_count: category_count,
              listing_shapes: all_shapes(@current_community.id)})
   end
 
   def new
-    processes = get_processes(@current_community.id)
-    templates = ListingShapeTemplates.new(processes)
+    templates = ListingShapeProcessViewUtils.available_templates(ListingShapeTemplates.all, process_summary)
+    template = ListingShapeProcessViewUtils.find_template(params[:template], templates, process_summary)
```
rap1ds: Find also calls available templates. Not necessary.

app/view_utils/listing_shape_process_view_utils.rb
```diff
@@ -0,0 +1,177 @@
...
+      else
+        Result::Error.new("Process must be none if author is not seller", code: "process_none_if_author_is_not_seller")
+      end
+    }
+
+    module_function
+
+    def sanitize(shape, process_summary)
+      shape.merge(
+        shipping_enabled: process_summary[:preauthorize_available] ? shape[:shipping_enabled] : false,
```
rap1ds: We have to know whether PayPal is available

rap1ds: Also, online_payments should be false, if we don't have appropriate process available.

PR #1173, opened 2015-04-22
Add support for PayPal Authorization process

ovan:
* Support PayPal express checkout process using the authorization flow in addition to our current order flow
* Parameterize transaction starting to allow selecting which flow to use but keep the default as order (enabled separately after migrations in a later branch)


app/services/paypal_service/api/payments.rb
```diff
@@ -128,15 +134,20 @@ def do_create(community_id, token)
 
       # The process either starts by creating a new payment...
       if (payment.nil?)
-        payment_res = create_payment(token)
-        if (payment_res[:success])
-          authorize_payment(community_id, payment_res[:data])
+        create_payment(token)
+          .and_then { |payment_entity|
```
sktoiva: If we convey to ruby conventions it should be 'do ... end`

app/services/paypal_service/data_types/merchant.rb
```diff
@@ -214,6 +256,9 @@ def create_get_express_checkout_details_response(opts); GetExpressCheckoutDetail
       def create_set_express_checkout_order(opts); SetExpressCheckoutOrder.call(opts) end
```
sktoiva: If this is depcrecated, should we add here the deprecation notice to logs?

app/services/paypal_service/data_types/merchant.rb
```diff
@@ -214,6 +256,9 @@ def create_get_express_checkout_details_response(opts); GetExpressCheckoutDetail
...
       def create_set_express_checkout_order_response(opts); SetExpressCheckoutOrderResponse.call(opts) end
```
sktoiva: Same here about deprecation.

app/services/paypal_service/merchant_actions.rb
```diff
@@ -162,6 +174,8 @@ def express_checkout_url(api, token)
         }
       ),
 
+      # Deprecated - Order flow will be removed soon
+      #
       set_express_checkout_order: PaypalAction.def_action(
```
sktoiva: Same here, should we notify about deprecation?

app/services/paypal_service/api/payments.rb
```diff
@@ -50,37 +50,43 @@ def request(community_id, create_payment, async: false)
...
-          Result::Success.new(
-            DataTypes.create_payment_request({
-                transaction_id: create_payment[:transaction_id],
-                token: response[:token],
-                redirect_url: response[:redirect_url]}))
+      create_payment_data = create_payment.merge(
+        { receiver_username: m_acc[:payer_id],
+          invnum: Invnum.create(community_id, create_payment[:transaction_id], :payment)})
+      request =
+        if (create_payment[:payment_action] == :order)
```
sktoiva: I don't like these assignents from conditionals, maybe move to a function?

PR #1176, opened 2015-04-23
Assign new listing shape to all existing categories

rap1ds:
- [X] Create category via Category API
- [x] Assign new listing shape to all existing categories

### DIFF
--- a/app/services/listing_service/store/shape.rb
+++ b/app/services/listing_service/store/shape.rb
@@ -151,7 +151,7 @@ module ListingService::Store::Shape
       hash[:units] = shape_model.listing_units.map { |unit_model|
         to_unit(from_unit_model_attributes(EntityUtils.model_to_hash(unit_model)))
       }
-      hash[:category_ids] = shape_model.categories.map { |category| category.id }
+      hash[:category_ids] = shape_model.categories.pluck(:id)
 
       Shape.call(hash)
     }.or_else(nil)
--- a/app/services/listing_service/store/shape.rb
+++ b/app/services/listing_service/store/shape.rb
@@ -1,6 +1,8 @@
 module ListingService::Store::Shape
 
   ListingUnitModel = ::ListingUnit
+  CategoryModel = ::Category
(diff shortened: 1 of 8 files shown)


rap1ds:
One thing to consider for the review:

CategoryListingShape is a join table, and now for each listing shape get request we query and return data from the join table. However, this is only needed in the admin view. Is this something we should change?

app/services/listing_service/store/shape.rb
```diff
@@ -117,6 +122,15 @@ def update(community_id:, listing_shape_id: nil, opts:)
 
   # private
 
+  # Note: If this method is needed in Category Store, then consider separating this code to
+  # own store, CategoryListingShapeStore
+  def assign_to_categories!(community_id, shape_id)
+    categories = CategoryModel.where(community_id: community_id)
+    categories.pluck(:id).each { |category_id|
+      CategoryListingShapeModel.create!(category_id: category_id, listing_shape_id: shape_id)
```
ovan: Wrap in single transaction

app/services/listing_service/store/shape.rb
```diff
@@ -137,6 +151,7 @@ def from_model(shape_model)
       hash[:units] = shape_model.listing_units.map { |unit_model|
         to_unit(from_unit_model_attributes(EntityUtils.model_to_hash(unit_model)))
       }
+      hash[:category_ids] = shape_model.categories.map { |category| category.id }
```
ovan: `.pluck(:id)`

So we do this now every time to avoid multiple presentations even if in most places we don't need the category IDs?

ovan: Ok, I saw the question now in pull request. :) With plug instead of mapping over full models this might be fast enough to not have to care.

app/services/listing_service/store/shape.rb
rap1ds: It's already in transaction. The transaction is initiated in `create` method.

PR #1179, opened 2015-04-24
Do not use fixture data for RSpec tests

rap1ds:
- Faster RSpec environment load for Zeus
- More isolated RSpec tests
- Now it is possible to add binding.pry breakpoints to the code that is used for creating fixture data. Before this change, the Zeus RSpec environment load stopped if there was a breakpoint. This of course happens still with the cucumber environment, but that's not a big issue I guess.


ovan:
:+1:

PR #1181, opened 2015-04-24
Fix error: Optional entities and collections crash if the value is null

rap1ds:


app/utils/entity_utils.rb
```diff
@@ -241,9 +241,13 @@ def transform_all(fields, input)
       out[name] = transform(spec[:transformers], input[name])
 
       out[name] =
-        if spec[:collection].present?
+        if spec[:collection].present? && out[name]
+          raise "Value for collection '#{name}' must be an Array. Was: #{out[name]} (#{out[name].class.name})" unless out[name].is_a? Array
```
ovan: error type! :) ArgumentError

app/utils/entity_utils.rb
```diff
@@ -241,9 +241,13 @@ def transform_all(fields, input)
...
+
           out[name].map { |v| transform_all(spec[:collection], v) }
-        elsif spec[:entity].present?
+        elsif spec[:entity].present? && out[name]
+          raise "Value for entity '#{name}' must be a Hash. Was: #{out[name]} (#{out[name].class.name})" unless out[name].is_a? Hash
```
ovan: ArgumentError

ovan:
:+1:

PR #1182, opened 2015-04-27
Optimize app controller

sktoiva:


app/controllers/application_controller.rb
```diff
@@ -206,42 +206,29 @@ def needs_redirect?(host, domain)
...
-
-    # Find by username
     app_domain = URLUtils.strip_port_from_host(APP_CONFIG.domain)
     ident = domain.chomp(".#{app_domain}")
-    by_ident = Community.where(ident: ident).first
 
-    if by_ident.present?
-      return by_ident
-    end
+    community = Community.where("domain = ? OR ident = ?", domain, ident).first
```
ovan: The previous version favours domain but I guess it's not specified here if ident or domain match comes first in the result set. If I get this right you'd need to match with "domain = ? OR (domain != ? AND ident = ?)" to get the same effect.

app/controllers/application_controller.rb
```diff
@@ -206,42 +206,29 @@ def needs_redirect?(host, domain)
...
-  # Before filter to check if current user is the member of this community
-  # and if so, find the membership
   def fetch_community_membership
     if @current_user
-      if @current_user.communities.include?(@current_community)
-        @current_community_membership = CommunityMembership.find_by_person_id_and_community_id_and_status(@current_user.id, @current_community.id, "accepted")
-        unless @current_community_membership.last_page_load_date && @current_community_membership.last_page_load_date.to_date.eql?(Date.today)
-          Delayed::Job.enqueue(PageLoadedJob.new(@current_community_membership.id, request.host))
-        end
+      @current_community_membership = CommunityMembership.find_by_person_id_and_community_id_and_status(@current_user.id, @current_community.id, "accepted")
```
ovan: Since we're touching the code better change to .where(... -format

PR #1183, opened 2015-04-27
Optimize homepage query load

sktoiva:


app/controllers/application_controller.rb
```diff
@@ -206,42 +206,29 @@ def needs_redirect?(host, domain)
...
   def fetch_community_membership
     if @current_user
-      if @current_user.communities.include?(@current_community)
-        @current_community_membership = CommunityMembership.find_by_person_id_and_community_id_and_status(@current_user.id, @current_community.id, "accepted")
-        unless @current_community_membership.last_page_load_date && @current_community_membership.last_page_load_date.to_date.eql?(Date.today)
-          Delayed::Job.enqueue(PageLoadedJob.new(@current_community_membership.id, request.host))
-        end
+      @current_community_membership = CommunityMembership.where(person_id: @current_user.id, community_id: @current_community.id, status: "accepted").first
+
+      last_page_load_date = Maybe(@current_community_membership).last_page_load_date.or_else(nil)
```
sktoiva: Maybe is unneccessary here, as the next line guards against current_community_membership being nil.

app/controllers/application_controller.rb
```diff
@@ -206,42 +206,29 @@ def needs_redirect?(host, domain)
...
-
-    # Find by username
     app_domain = URLUtils.strip_port_from_host(APP_CONFIG.domain)
     ident = domain.chomp(".#{app_domain}")
-    by_ident = Community.where(ident: ident).first
 
-    if by_ident.present?
-      return by_ident
-    end
+    community = Community.where("domain = ? OR ident = ?", domain, ident).first
```
sktoiva: If we want to prefer domains:
```
community = Community.where("domain = ? OR (domain != ? AND ident = ?)", domain, domain, ident)
```

PR #1184, opened 2015-04-28
Add admin UI for enabling multiple locales.

mporkola:


app/services/marketplace_service/api/marketplaces.rb
```diff
@@ -70,6 +70,62 @@ def create(params)
...
+      community.settings = settings
+      community.save!
+
+      ensure_translations(community, locales).map { |translations|
+        locale = translations[:locale]
+        translations[:translations].each{ |translation|
+          add_translation(community, locale, translation[:translation_key], translation[:translation])
+        }
+      }
+      Rails.cache.delete("/translation_service/community/#{community.id}")
```
mporkola: Is this the way to clear this cache?

app/assets/stylesheets/partials/_admin.css.scss
```diff
@@ -18,4 +18,22 @@
...
-}
\ No newline at end of file
+}
+
+// Matching Selectize widget to our look
+.locale-selection__input-widget.selectize-control {
+  .selectize-input {
+    border: 2px solid $border;
+    @include border-radius(em(5));
+    font-size: 16px;
```
Gnito: We have two interesting functions to deal with sizes:
em: calculates em value for given pixels
lines: calculates line-heights in relation to our default (24px)
ks. type.scss

app/services/marketplace_service/api/marketplaces.rb
Gnito: Yes! But now that I think about this - translatioService handles this cache. It deletes the cache in two situations: 
1) someone calls create api-call
2) someone calls delete api-call
I guess, the first situation handles this case.

Gnito:
Since loading of selectize.js seems to take a short time, it would be better to create some kind of fade-in for the language selection. But check first if the assumption (i.e. loading) is correct.

app/controllers/admin/community_customizations_controller.rb
```diff
@@ -23,16 +28,24 @@ def update_details
...
       params.require(:community_customizations).require(locale).permit(*permitted_params)
-
       locale_params = params[:community_customizations][locale]
       customizations = find_or_initialize_customizations_for_locale(locale)
       customizations.update_attributes(locale_params)
     end
 
+    enabled_locales = params[:enabled_locales]
+    all_locales = MarketplaceService::API::Marketplaces.all_locales
+    enabled_locales_valid = enabled_locales.map do |locale|
```
Gnito: or something like
```ruby
enabled_locales.map { |locale|  all_locales.include?(locale) }
```

app/controllers/admin/community_customizations_controller.rb
```diff
@@ -23,16 +28,24 @@ def update_details
...
+    end
+    if enabled_locales_valid
+      MarketplaceService::API::Marketplaces.set_locales(@current_community, enabled_locales)
+    end
+
     transaction_agreement_checked = Maybe(params)[:community][:transaction_agreement_checkbox].is_some?
     community_update_successful = @current_community.update_attributes(transaction_agreement_in_use: transaction_agreement_checked)
 
-    if updates_successful.all? && community_update_successful
+    if updates_successful.all? && community_update_successful && enabled_locales_valid.all?
```
Gnito: I guess this doesn't handle the case of sending an empty array
```
pry(main)> [].all?
=> true
```

app/services/marketplace_service/api/marketplaces.rb
```diff
@@ -148,6 +204,10 @@ def create_community_customization!(community, marketplace_name, locale)
         community.community_customizations.create(customization_params(marketplace_name, locale))
       end
 
+      def first_or_create_community_customization!(community, marketplace_name, locale)
+        community.community_customizations.first_or_create(customization_params(marketplace_name, locale))
```
Gnito: Did you test this "first_or_create" function?
Also, there is a mismatch with using "!" -> you should use the same behaviour inside your function as in its definition.

app/services/marketplace_service/api/marketplaces.rb
```diff
@@ -70,6 +70,62 @@ def create(params)
...
+        ensured_translations = translations.map { |translation|
+          key = translation[:translation_key]
+          translated_content = translation[:translation] || I18n.t(key)
+          {
+              translation_key: key,
+              translation: translated_content
+          }
+        }
+        {
+            locale: locale,
```
Gnito: too much indentation

app/services/marketplace_service/api/marketplaces.rb
```diff
@@ -70,6 +70,62 @@ def create(params)
...
+        translations[:translations].each{ |translation|
+          add_translation(community, locale, translation[:translation_key], translation[:translation])
+        }
+      }
+      Rails.cache.delete("/translation_service/community/#{community.id}")
+    end
+
+    def ensure_translations(community, locales)
+      locales.map { |locale|
+        translations_res = TranslationService::API::Api.translations.get(community.id, {locales: [locale]})
```
Gnito: I started to think this a bit different angle: could this work in a bit different way
```
TranslationService::API::Api.translations.get(1, {locales: ['ja'], fallback_locale: 'en'}).data
  .uniq { |t| t[:translation_key] }
  .map {|t| {translation_key: t[:translation_key], translations:[{locale: 'ja', translation: t[:translation]}]}}
```
1st line:  get translations with new locale ('ja'), but use existing locale as fallback_locale
2nd line: useless line, but gives an idea how to remove duplicates if there are more locales per translation_key
3rd line:  map fetched locales in a format that can be passed to TranslationService::API::Api.translations.create(...) using new locale (i.e. 'ja')

app/views/admin/community_customizations/_locale_selection_checkboxes.haml
```diff
@@ -0,0 +1,30 @@
...
+    });
+    ST.loadTranslations(languageKeys);
+    var selectize = $('#locale-selection').selectize({plugins: ['remove_button'], items: #{enabled_locale_keys.to_json.html_safe}})[0].selectize;
+    selectize.on("change", function(value) {
+      defaultLanguage = value ? selectize.options[value[0]].text.trim() : "None";
+      $(".selected-default-language-name").text(defaultLanguage);
+    });
+%div.locale-selection
+  .input-header
+    Enabled languages
```
Gnito: another localization point

app/views/admin/community_customizations/_locale_selection_checkboxes.haml
```diff
@@ -0,0 +1,30 @@
...
+    .info-text-content
+      %p
+        Languages to be used nönönö
+  .row.col-10
+    %select.locale-selection__input-widget{ multiple: true, name: "enabled_locales[]", id: "locale-selection" }
+      - all_locales.each do |locale|
+        %option{ value: locale[:locale_key] }
+          = t("admin.communities.available_languages.#{locale[:locale_key]}")
+  %p#selected-default-language{ style: "clear: both;" }
+    Default language:
```
Gnito: localize this

app/views/admin/community_customizations/_locale_selection_checkboxes.haml
```diff
@@ -0,0 +1,30 @@
+- content_for :extra_javascript do
+  :javascript
+    var languageKeys = _.map(#{all_locales.to_json}, function(locale) {
+      return "admin.communities.available_languages." + locale["locale_key"];
+    });
+    ST.loadTranslations(languageKeys);
```
Gnito: Are these needed anymore?

app/services/marketplace_service/api/marketplaces.rb
Gnito: This way we could handle the creation of new dynamic translations for new languages a bit simpler.

app/assets/stylesheets/partials/_admin.css.scss
rap1ds: :+1: Ems should be used, so instead of `2px`, use `em(2)`

app/controllers/admin/community_customizations_controller.rb
```diff
@@ -5,11 +5,16 @@ def edit_details
     @selected_left_navi_link = "tribe_details"
     # @community_customization is fetched in application_controller
     @community_customizations ||= find_or_initialize_customizations(@current_community.locales)
+    all_locales = MarketplaceService::API::Marketplaces.all_locales
+    enabled_locale_keys = available_locales().map{ |locale| locale[1] }
```
rap1ds: Ruby trick:

There's a method `second`, which can be used to pick the second item from the array

```ruby
available_locales().map{ |locale| locale.second }
```

and then there's a shorthand for that:

```ruby
available_locales().map(&:second)
```

app/controllers/admin/community_customizations_controller.rb
```diff
@@ -23,16 +28,24 @@ def update_details
...
       ]
       params.require(:community_customizations).require(locale).permit(*permitted_params)
-
       locale_params = params[:community_customizations][locale]
       customizations = find_or_initialize_customizations_for_locale(locale)
       customizations.update_attributes(locale_params)
     end
 
+    enabled_locales = params[:enabled_locales]
+    all_locales = MarketplaceService::API::Marketplaces.all_locales
```
rap1ds: Minor perf improvement: With `to_set` you can convert array to Set (http://ruby-doc.org/stdlib-2.2.2/libdoc/set/rdoc/Set.html). Set allows faster constant time look up, which improves the perf of the loop in loop code.

app/controllers/admin/community_customizations_controller.rb
```diff
@@ -23,16 +28,24 @@ def update_details
...
+    enabled_locales_valid = enabled_locales.map do |locale|
+      all_locales.include? locale
+    end
+    if enabled_locales_valid
```
rap1ds: Because `map` always returns an array, the `enabled_locales_valid` will be an array of boolean values (or empty array). In Ruby, everything else than `false` and `nil` is false, so this `if` block will always be executed.

I think what you are looking for is `all?` method. It returns `true` if all the items in array return true, so:

```ruby
enabled_locales_valid = enabled_locales.all? { all_locales.include?(locale) }
```

app/services/marketplace_service/api/marketplaces.rb
```diff
@@ -70,6 +70,62 @@ def create(params)
...
+        translations = translations_res["data"]
```
rap1ds: Prefer symbol, `translations_res[:data]` or dot notation `translations_res.data`

app/services/marketplace_service/api/marketplaces.rb
```diff
@@ -70,6 +70,62 @@ def create(params)
...
+      default_locale = community.locales[0]
+      community_name = community.name(default_locale)
+      locales.each { |locale| Helper.first_or_create_community_customization!(community, community_name, locale) }
+
+      settings = community.settings || {}
+      settings["locales"] = locales
+      community.settings = settings
+      community.save!
+
+      ensure_translations(community, locales).map { |translations|
```
rap1ds: I don't quite understand why do we need this? Why is this necessary?

rap1ds:
Have you tested the selectize with mobile?

app/controllers/admin/community_customizations_controller.rb
Gnito: Uh, didn't know that :+1:

PR #1188, opened 2015-04-29
Add missing processes

rap1ds:
This PR adds two processes, none & author_is_seller: false and none & author_is_seller: true to all marketplaces that do not yet have these processes


mporkola:
:+1:

PR #1196, opened 2015-05-06
removed og:locale meta tag since we don't have that data

Gnito:


gnomet:
Yep, this is what it was about. Feels silly that it wasn't done yet, but been piling things until I get to code mode.. :) Thanks for removal, good to merge.

PR #1199, opened 2015-05-06
handle empty name_display_type as fullname

Gnito:
Better solution might be to make db migration and change the behaviour in 5++ files. 
However, is this good enough?
(This comment is related to first commit)

### DIFF
--- a/app/models/community.rb
+++ b/app/models/community.rb
@@ -284,6 +284,10 @@ class Community < ActiveRecord::Base
     end
   end
 
+  def name_display_type
+    Maybe(read_attribute('name_display_type')).or_else('fullname');
+  end
+
   def full_name(locale)
     name(locale)
   end
--- a/app/models/community.rb
+++ b/app/models/community.rb
@@ -145,8 +145,8 @@ class Community < ActiveRecord::Base
   VALID_BROWSE_TYPES = %{grid map list}
   validates_inclusion_of :default_browse_view, :in => VALID_BROWSE_TYPES
 
-  VALID_NAME_DISPLAY_TYPES = %{first_name_only first_name_with_initial}
-  validates_inclusion_of :default_browse_view, :in => VALID_BROWSE_TYPES, :allow_nil => true, :allow_blank => true
+  VALID_NAME_DISPLAY_TYPES = %{first_name_only first_name_with_initial full_name}
+  validates_inclusion_of :name_display_type, :in => VALID_NAME_DISPLAY_TYPES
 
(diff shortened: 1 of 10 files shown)


app/models/community.rb
```diff
@@ -279,6 +279,10 @@ def name(locale)
     end
   end
 
+  def name_display_type
+    Maybe(read_attribute('name_display_type')).or_else('fullname');
```
ovan: ; :)

app/models/community.rb
```diff
@@ -145,8 +145,8 @@ class Community < ActiveRecord::Base
   VALID_BROWSE_TYPES = %{grid map list}
   validates_inclusion_of :default_browse_view, :in => VALID_BROWSE_TYPES
 
-  VALID_NAME_DISPLAY_TYPES = %{first_name_only first_name_with_initial}
-  validates_inclusion_of :default_browse_view, :in => VALID_BROWSE_TYPES, :allow_nil => true, :allow_blank => true
+  VALID_NAME_DISPLAY_TYPES = %{first_name_only first_name_with_initial full_name}
+  validates_inclusion_of :name_display_type, :in => VALID_NAME_DISPLAY_TYPES
```
Gnito: This looked like a code that didn't work previously...

db/migrate/20150508141500_change_name_display_type_in_communities.rb
```diff
@@ -0,0 +1,15 @@
+class ChangeNameDisplayTypeInCommunities < ActiveRecord::Migration
+  def up
+    execute("UPDATE communities
+      SET name_display_type='full_name'
+      WHERE name_display_type = '';
```
ovan: Do we also need to handle the case where this is NULL, or does this also cover that already?

app/models/community.rb
ovan: Are you intentionally disabling the validation for default_browse_view field and adding validation for name_display_type?

Gnito: :default_browse_view has validations just above these lines. 
(So this is just old copy-paste error, which has not worked correctly ever)
As discussed irl, I started to check if this has real-life effects since there was before allow_nil etc. extra rules ...deciced not to spend time with that since we should never allow nil or blank for this parameter.

This is related to
http://localhost:3000/en/admin/communities/1/edit_look_and_feel#community_default_browse_view
Where there are just 3 options allowed (and unlike previous name-type selection functionality, there is no blank as default value for any of those three options. And after this bugfix, blank option is removed from name-type select too.)

PR #1201, opened 2015-05-07
Close listings and delete listing shape

rap1ds:
Now possible to close listings that belong to the listing shape and delete the listing shape.


app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -40,7 +40,9 @@ def edit
 
     return redirect_to error_not_found_path if shape.nil?
 
-    render_edit_form(params[:id], shape, process_summary, available_locales())
+    count = Listing.where(listing_shape_id: params[:id]).currently_open.count()
```
ovan: What about moving Listing model queries behind a partial listing api?

app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -124,6 +126,37 @@ def order
     render nothing: true, status: 200
   end
 
+  def close_listings
+    shape_res = listing_api.shapes.get(community_id: @current_community.id, listing_shape_id: params[:id])
+
+    if shape_res.success
+      Listing.where(listing_shape_id: params[:id]).update_all(open: false)
```
ovan: listing api?

app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -124,6 +126,37 @@ def order
...
+      return redirect_to action: :index
+    end
+  end
+
+  def destroy
+    shape_res = listing_api.shapes.get(community_id: @current_community.id, listing_shape_id: params[:id])
+
+    if shape_res.success
+      Listing.where(listing_shape_id: params[:id]).update_all(open: false, listing_shape_id: nil)
+      ShapeService.new(processes).delete(
```
ovan: Does this need ShapeService, why not call directly shapes_api ?

app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -139,22 +172,23 @@ def uneditable_fields(process_summary)
...
+  def render_edit_form(id, form, count, process_summary, available_locs)
+    locals = common_locals(form, count, process_summary, available_locs).merge(
       id: id,
       name: pick_translation(form[:name])
     )
     render("edit", locals: locals)
   end
 
-  def common_locals(form, process_summary, available_locs)
+  def common_locals(form, count, process_summary, available_locs)
```
ovan: count isn't really common?

app/models/listing_shape.rb
```diff
@@ -14,10 +14,12 @@
 #  sort_priority              :integer          default(0), not null
 #  created_at                 :datetime         not null
 #  updated_at                 :datetime         not null
+#  deleted                    :boolean          default(FALSE)
 #
 # Indexes
 #
 #  index_listing_shapes_on_community_id  (community_id)
+#  index_listing_shapes_on_deleted       (deleted)
```
ovan: Needs validation if this index is actually used with some query since MySQL cannot combine indices.

ovan: Better index for front page queries would be something like:

```sql
CREATE INDEX bar on listing_shapes(community_id, deleted, sort_priority);
```

app/services/listing_service/listing_shapes_api.md
```diff
@@ -69,3 +69,25 @@ Response:
...
+
+Response:
+
+```ruby
+{ id: 12345
+, community_id: 9876
+, price_enabled: true
+, transaction_process_id: 123
+, name_tr_key: TranslationKey.new("listing_shape.1234.123")
+, action_button_tr_key: TranslationKey.new("action_button.1234.123")
```
ovan: TranslationKey?

app/services/listing_service/store/shape.rb
```diff
@@ -127,6 +127,18 @@ def update(community_id:, listing_shape_id: nil, opts:)
...
+  def delete(community_id:, listing_shape_id:)
+    shape_model = find_shape_model(
+      community_id: community_id,
+      listing_shape_id: listing_shape_id)
+
+    return nil if shape_model.nil?
+
+    shape_model.update_attributes(deleted: true)
+
+    deleted_shape = from_model(shape_model, true)
```
ovan: unused local

app/view_utils/shape_service.rb
```diff
@@ -50,6 +50,13 @@ def create(community_id:, default_locale:, opts:)
     )
   end
 
+  def delete(community_id:, listing_shape_id:)
```
ovan: Feels unnecessary since this just calls through

app/controllers/admin/listing_shapes_controller.rb
rap1ds: actually, it needs to be there since the code would throw undefined local var without this.

PR #1202, opened 2015-05-07
Sort category id array

rap1ds:
Fixes one randomly failing test. The order of category IDs may change.


rap1ds:
This is so simple that I'll just merge it now.

PR #1206, opened 2015-05-08
Fix FB like button text language on few locales

gnomet:
Minor fix to way we send our locale for FB like button.
Better default for future additions and fixes few cases manually.

A better future step to take would be to use full locales internally. No didn't feel useful to try to parse those from current short locales, so I fixed our else branch and added few manual rows. The idea is that this part doesn't need to be touched usually. If clients connect with us (like they did now) to get the button in their language we can add it quickly. But most might work with the default, or be happy with "like".


mporkola:
looks good :tada:

PR #1208, opened 2015-05-08
Fix bug: Send welcome email after FB login

rap1ds:
- [X] Send welcome email after FB login
- [X] Use delayed job to move the email sending away from the request cycle


sktoiva:
:+1: This is good, as this'll fix also the normal user sign up flow email sending if there's a problem with  SES.

PR #1209, opened 2015-05-08
Support author is not seller process

rap1ds:
The previous Order Type UI implementation completely ignored the fact that we have also processes with `author_is_seller: false`. This PR implements support for that, which means:

- Add `author_is_seller` to each template
- Change `filter` method to reject `online_payments` and `shipping` if `!author_is_seller`
- Take `author_is_seller` into account when selecting the correct transaction process.


sktoiva:
:+1:

PR #1211, opened 2015-05-11
tune the feedback email texts a bit better

gnomet:
A customer reported some issues with the texts, so tuned them a bit better.


sktoiva:
:+1:

PR #1212, opened 2015-05-11
bugfix: listing image width & height

Gnito:


app/models/listing_image.rb
```diff
@@ -44,6 +44,24 @@ class ListingImage < ActiveRecord::Base
                                     :unless => Proc.new {|model| model.image.nil? }
 
 
+  def get_dimensions_for_style(style)
+    binding.pry
```
sktoiva: Binding.pry!

Gnito: :D

Gnito:
@sktoiva binding.pry is removed. Do you still want to check this? I would like to get this to production soon.

PR #1213, opened 2015-05-12
Disable order type delete if...

rap1ds:
- [x] The order type is the only order type in that marketplace
- [x] The order type is the only order type in use for some category

### DIFF
--- a/app/controllers/admin/listing_shapes_controller.rb
+++ b/app/controllers/admin/listing_shapes_controller.rb
@@ -42,14 +42,7 @@ class Admin::ListingShapesController < ApplicationController
 
     return redirect_to error_not_found_path if shape.nil?
 
-    count = listing_api.listings.count(
-      community_id: @current_community.id,
-      query: {
-        listing_shape_id: params[:id].to_i,
-        open: true
-      }).data
-
-    render_edit_form(params[:id], shape, count, process_summary, available_locales())
+    render_edit_form(params[:id], shape, process_summary, available_locales())
   end
 
   def create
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -376,6 +376,8 @@ en:
         update_failure: "Could not save changes. Error: %{error_msg}"
         delete: "Delete order type"
         confirm_delete: "There are %{count} listings with this order type. If you delete the order type, those listings will be closed. Are you sure you want to delete the order type?"
+        can_not_delete_last: "You can't delete this order type because it's the only one type in your marketplace."
```
sktoiva: "You can't delete this order type because it's the only one type in your marketplace."
->
"You can't delete this order type because it's the only type in your marketplace."

app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -198,6 +204,53 @@ def common_locals(form, count, process_summary, available_locs)
...
   end
 
+  def can_delete_shape?(current_shape_id, shapes)
+    listing_shapes_categories_map = shapes.map { |shape|
+      [shape[:id], shape[:category_ids]]
+    }
+
+    categories_listing_shapes_map = HashUtils.reverse_key_enum_hash(listing_shapes_categories_map)
+
+    only_one_in_category_ids = categories_listing_shapes_map.select { |category_id, shape_ids|
```
sktoiva: I would rename this to `last_in_category_ids`

PR #1215, opened 2015-05-12
Fix link URL in rejection notification email

gnomet:
Mails telling about notifications had a broken link, this fixes that.

The message_url was used although the ID parameter came from the transaction


rap1ds:
Looks good to me :+1:

gnomet:
Thanks for raving! @rap1ds

PR #1216, opened 2015-05-12
Fine tuning the Shape UI

rap1ds:
- [x] Don't show locale label if there's only one locale
- [x] Consistent styling (labels)
- [x] Pluralize
- [x] Disable sort arrows if only one order type
- [x] Also, disable sort arrows if only one category (that was missing from the category view)


config/locales/en.yml
```diff
@@ -376,13 +376,17 @@ en:
         update_failure: "Could not save changes. Error: %{error_msg}"
         delete: "Delete order type"
         confirm_delete: "There are %{count} listings with this order type. If you delete the order type, those listings will be closed. Are you sure you want to delete the order type?"
+        confirm_delete_singular: "There is %{count} listing with this order type. If you delete the order type, this listings will be closed. Are you sure you want to delete the order type?"
```
sktoiva: Plurals are wrong. Either listing or listings - probably listing in both places.

config/locales/en.yml
```diff
@@ -376,13 +376,17 @@ en:
...
         can_not_delete_last: "You can't delete this order type because it's the only type in your marketplace."
         can_not_delete_only_one_in_categories: "You can't delete this order type because in following categories this order type is the only one in use: %{categories}"
       listing_shape_name: Name
       action_button_label: "Checkout button label"
       open_listings: "There are %{count} open listings with this order type. If you change any of the settings below, those listings will retain the old settings. Those listings can be changed to the new settings by editing them manually. If you don't want to have any listings with the old settings visible on your site, you can close them by clicking the button below."
+      open_listings_singular: "There is %{count} open listing with this order type. If you change any of the settings below, that listings will retain the old settings. The listing can be changed to the new settings by editing it manually. If you don't want to have the listing with the old settings visible on your site, you can close it by clicking the button below."
```
sktoiva: Same here.

rap1ds: Damn, good catch!

PR #1217, opened 2015-05-13
Use shape name in the URL instead of ID

rap1ds:
-[x] Add `id` to Shape form shape
-[x] Listing shape API can `get`, `update`, and `delete` shape by name
-[x] Use name instead of ID

### DIFF
--- a/app/services/listing_service/api/shapes.rb
+++ b/app/services/listing_service/api/shapes.rb
@@ -34,10 +34,19 @@ module ListingService::API
       ))
     end
 
-    def update(community_id:, listing_shape_id: nil, opts:)
+    def update(community_id:, listing_shape_id: nil, name: nil, opts:)
+      if listing_shape_id.present? && name.present?
+        return Result::Error.new("Can not have both listing shape id (#{listing_shape_id}) and name (#{name}) present.")
+      end
+
+      unless listing_shape_id.present? || name.present?
+        return Result::Error.new("Must have either id or name present.")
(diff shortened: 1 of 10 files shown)


app/services/listing_service/store/shape.rb
```diff
@@ -193,8 +196,19 @@ def from_unit_model_attributes(hash)
...
+    q =
+      if listing_shape_id.present?
+        q.where(id: listing_shape_id)
+      elsif name.present?
+        q.where(name: name)
+      else
+        raise ArgumentError.new("Must have either name or listing shape id present")
+      end
+
+    q.first
```
ovan: no need for local variable q, just two branches that both do the whole chain of ops.

app/services/listing_service/api/shapes.rb
```diff
@@ -42,10 +56,19 @@ def update(community_id:, listing_shape_id: nil, opts:)
...
 
-    def delete(community_id:, listing_shape_id:)
+    def delete(community_id:, listing_shape_id: nil, name: nil)
+      if listing_shape_id.present? && name.present?
+        return Result::Error.new("Can not have both listing shape id (#{listing_shape_id}) and name (#{name}) present.")
+      end
+
+      unless listing_shape_id.present? || name.present?
+        return Result::Error.new("Must have either id or name present.")
+      end
```
ovan: These two checks could be in private method because they are used from two places.

PR #1223, opened 2015-05-15
Custom units

mporkola:


app/assets/javascripts/admin/listing_shapes.js
```diff
@@ -93,6 +86,21 @@ window.ST.initializeListingShapeForm = function(formId) {
     toggleLabel($(".js-unit-label"), enabled);
   };
 
+  var addCustomUnitForm = function() {
+    var $form = $(".js-listing-shape-add-custom-unit-form").last().clone();
+    $form.find('input').prop('disabled', false);
+    var randomId = Math.floor(Math.random() * (Number.MAX_VALUE - 1)) + 1;
```
rap1ds: If I understand right, this doesn't have to be random, instead, it has to be unique, right?

There's a handy function in lodash `uniq` https://lodash.com/docs#uniq, maybe that could be used here?

config/locales/fi.yml
```diff
@@ -398,6 +398,11 @@ fi:
         week: ~
         month: ~
       can_not_find: "Tyyppiä ID:llä %{id} ei löytynyt"
+      add_custom_unit: "+ Lisää uusi yksikkö…"
+      custom_unit_form:
+        title: "Uusi hinnoittelutyyppi"
+        label_heading: "Nimi"
+        selector_label_heading: "Yksikkö"
```
rap1ds: Placeholder translations are missing

spec/view_utils/shape_service_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+require 'spec_helper'
+
+describe ShapeService do
+
+  let(:shape_service) {ShapeService.new([{:id=>1, :community_id=>1, :author_is_seller=>true, :process=>:none}])}
+
+  def create_shape(opts = {})
+    defaults = {
+      action_button_label: {"en" => "ACTION LABEL"},
+      name: {"en" => "ERIKOIS SHAPE"},
```
rap1ds: I guess "ERIKOIS" is not en :)

db/migrate/20150512083842_remove_old_translation_from_listing_unit.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveOldTranslationFromListingUnit < ActiveRecord::Migration
+  def up
+    remove_column :listing_units, :translation_key
+  end
+
+  def down
+    add_column :listing_units, :translation_key, :string, limit: 64, after: :quantity_selector
+  end
+end
```
rap1ds: Because of the 2 min preboot delay, there's a change that removing a column may break the system for 2 minutes. The reason is that when you run the migration we remove the column, but the old code which still thinks that there is the `translation_key` may break.

Here's a good blog post about the different things you have to consider when doing zero downtime deployments in Rails. http://blog.codeship.com/rails-migrations-zero-downtime/ See the chapter "Removing columns".

I'd suggest that you add the `self.columns` method to the `ListingUnit` model as described in the article.

spec/view_utils/shape_service_spec.rb
```diff
@@ -0,0 +1,47 @@
+# coding: utf-8
+require 'spec_helper'
+
+describe ShapeService do
```
rap1ds: Very good that you added tests to `ShapeService`! :+1:

rap1ds:
There are a couple places where translations are probably still using the old `translation_key`:

screen shot 2015-05-25 at 14 36 35
screen shot 2015-05-25 at 14 36 08
screen shot 2015-05-25 at 14 30 08
screen shot 2015-05-25 at 14 30 31

rap1ds:
A couple ideas about the design:

- If we added a disabled and checked checkbox in front of the custom unit, it would be nicely aligned with the others. What do you think?
- We could add a pointer cursor for the "Add new custom unit" link

screen shot 2015-05-25 at 14 28 35

- Also, if there's only one language in use, we haven't usually displayed the language name

screen shot 2015-05-25 at 14 13 20

db/migrate/20150512083842_remove_old_translation_from_listing_unit.rb
mporkola: If I understood the blog post correctly, this https://github.com/sharetribe/sharetribe/commit/f7f3e68bf4617ffc054b76938a406f13a2348a22 should do it?

app/assets/javascripts/admin/listing_shapes.js
```diff
@@ -106,6 +109,16 @@ window.ST.initializeListingShapeForm = function(formId) {
     el.toggleClass("listing-shape-label-disabled", !state);
   };
 
+  $('.js-price-enabled').change(function() {
+    priceChanged($(this));
```
rap1ds: We should also disable the custom units when price is disabled.

db/migrate/20150512083842_remove_old_translation_from_listing_unit.rb
rap1ds: Yep, that will do it.

Also, when you deploy the code, the deploy script asks which migrations to run. DO NOT RUN THIS MIGRATION. Wait for the 2 minutes for the new code to kick in, and after that run this migration manually.

rap1ds:
screen shot 2015-05-26 at 15 57 35

rap1ds:
Also, seller and buyer email receipts are missing the unit translations:

screen shot 2015-05-26 at 16 09 56

screen shot 2015-05-26 at 16 09 27

rap1ds:
Saving a listing with custom units does not work properly if you have more than 1 custom unit. The reason is that for all custom units, the dropdown item value is `custom`.

Steps to reproduce:

1. Create listing shape with more than one custom unit
2. Add a new listing with price and use the last custom unit

Actual result: The first custom unit is used.

mporkola:
This should allow multiple custom units to be used: https://github.com/sharetribe/sharetribe/commit/8bf7934150b3a4961e2bf0cd4f8617ed359fbb89

PR #1227, opened 2015-05-19
Handle custom translations.

mporkola:
Don't break the locale selector in cases where marketplace is using a language we don't fully support.

Without a custom language:
Without a custom language

With a custom language:
With a custom language


app/controllers/admin/community_customizations_controller.rb
```diff
@@ -5,15 +5,23 @@ def edit_details
...
     @community_customizations ||= find_or_initialize_customizations(@current_community.locales)
-    all_locales = MarketplaceService::API::Marketplaces.all_locales
+    all_locales = MarketplaceService::API::Marketplaces.all_locales.map{|l| l[:locale_key]}
     enabled_locale_keys = available_locales.map(&:second)
 
+    custom_translation_in_use = !all_locales.include?(@current_community.locales.first)
+    if custom_translation_in_use
+      custom_locale_key = @current_community.locales.first
+      custom_locale_name = Kassi::Application.config.AVAILABLE_LOCALES.select {|k, v| v == custom_locale_key}.map(&:first).first
+      binding.pry
```
rap1ds: Err!

app/views/admin/community_customizations/_locale_selection.haml
```diff
@@ -2,9 +2,12 @@
   :javascript
     var selectize = $('#locale-selection').selectize({plugins: ['remove_button'], items: #{enabled_locale_keys.to_json.html_safe}})[0].selectize;
     selectize.on("change", function(value) {
-      defaultLanguage = value ? selectize.options[value[0]].text.trim() : "None";
-      $(".selected-default-language-name").text(defaultLanguage);
-      $('#edit_community').valid();
+      var customLocaleExists = #{custom_locale.present?}
+      if (!customLocaleExists) {
+        defaultLanguage = value ? selectize.options[value[0]].text.trim() : "None";
```
rap1ds: "None" is not translated

app/controllers/admin/community_customizations_controller.rb
```diff
@@ -5,15 +5,23 @@ def edit_details
     @selected_left_navi_link = "tribe_details"
     # @community_customization is fetched in application_controller
     @community_customizations ||= find_or_initialize_customizations(@current_community.locales)
-    all_locales = MarketplaceService::API::Marketplaces.all_locales
+    all_locales = MarketplaceService::API::Marketplaces.all_locales.map{|l| l[:locale_key]}
     enabled_locale_keys = available_locales.map(&:second)
 
+    custom_translation_in_use = !all_locales.include?(@current_community.locales.first)
```
rap1ds: Why first?

app/views/admin/community_customizations/_locale_selection.haml
rap1ds: Maybe it could be just `-`

app/controllers/admin/community_customizations_controller.rb
```diff
@@ -54,6 +56,25 @@ def update_details
 
   private
 
+  def validate_and_set_locales(params)
```
rap1ds: The `and` in the name is a strong indicator that this method should be splitted :)

app/controllers/admin/community_customizations_controller.rb
rap1ds: `custom` is not very good name here, since it strongly refers to per marketplace custom languages, which will be removed. I vote for `unofficial` or something like that.

(not sure if that needs to be changed right now)

PR #1235, opened 2015-05-21
Improve locale selection and redirect

rap1ds:
- Use fallback of user's language, if available
- Use fallback of URL language, if available
- Redirect, if needed

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -48,12 +48,7 @@ class ApplicationController < ActionController::Base
   def redirect_removed_locale
     if params[:locale] && Kassi::Application.config.REMOVED_LOCALES.include?(params[:locale])
       fallback = Kassi::Application.config.REMOVED_LOCALE_FALLBACKS[params[:locale]]
-
-      if @current_community.default_locale == fallback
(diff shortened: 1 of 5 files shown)


config/available_locales.rb
```diff
@@ -1,41 +1,42 @@
...
 module Sharetribe
 
-  # Format: [name, identifier, language, region]
+  # Format: [name, identifier, language, region, fallback identifier]
   #
   # language format: ISO 639-1, two letters, lowercase
   # region format: ISO 3166, two letters, uppercase
+  # fallbacks: should not include US English, which is a last default fallback for each language
 
   AVAILABLE_LOCALES = [
```
ovan: What do you think about making this a hash? :p

app/controllers/application_controller.rb
```diff
@@ -124,6 +112,22 @@ def set_locale
     }
   end
 
+  # If URL contains locale parameter that doesn't match with the selected locale,
+  # redirect to the selected locale
+  def redirect_locale_param
+    needs_redirect = params[:locale].present? && params[:locale] != I18n.locale.to_s
+
+    redirect_to_locale(I18n.locale, :temporary_redirect) if needs_redirect
```
ovan: `redirect_to_locale(...) if param_locale_not_selected`

PR #1236, opened 2015-05-22
Change the class attr selector

rap1ds:
We have some rows that are not correctly aligned:

screen shot 2015-05-22 at 11 50 33

The reason for this is that we use other classes with the `.col-*` `div` and the current class attribute selector does not match.

For example, instead of this:

```haml
.row
  .myclass.col-12
```

We should use:

```haml
.row
  .col-12
    .myclass
```

This pull request changes the class attribute selector so that it's perfectly ok to use the `.myclass.col-12` version.

Fixed version:

screen shot 2015-05-22 at 11 47 40


ovan:
Good context description! :)

PR #1241, opened 2015-05-26
Create community customization if one is not found for the requested locale.

mporkola:


rap1ds:
There's one comment but the current implementation is also good to go :+1:

PR #1243, opened 2015-05-26
Save nil quantity, if the params do not include it

rap1ds:
Why: There are already some customers that would like to use the new units, but they are also using old "free form units".

This pull request lets them to use units so that first we set manually the `price_quantity_placeholder` database value to `nil` and after that they can use the Shape UI. When new listings are created, they will use the new units. When old listings are edited, they will use the new units and (here comes the beef) **old quantity is removed from the listing**.


Gnito:
:+1:

PR #1247, opened 2015-05-28
Fix floating point error

rap1ds:


mporkola:
go ahead :moneybag: :+1: :

PR #1249, opened 2015-05-29
Entity field key must be a Symbol

rap1ds:
Why: `deserialize` method relies on the fact that the field key is always symbol, even though that's not enforced by the code. This PR adds a check that ensures the key is always a symbol.


mporkola:
:cool:

PR #1252, opened 2015-05-29
More robust time-related tests

rap1ds:
- [X] Make `time_to` tests use specific time, which is not affected by the DST changes
- [X] Set Timecop to safe mode, which enforces the use of blocks


mporkola:
good stuff

PR #1253, opened 2015-06-01
Custom unit kind

rap1ds:
- [X] Add unit "kind" to custom units


app/services/listing_service/store/shape.rb
```diff
@@ -44,11 +44,13 @@ module ListingService::Store::Shape
 
   BuiltInUnit = EntityUtils.define_builder(
     [:type, :to_symbol, one_of: [:piece, :hour, :day, :night, :week, :month]],
+    [:kind, :to_symbol, const_value: :time],
```
mporkola: `:piece` should probably be removed?

PR #1255, opened 2015-06-01
Cache listing item partials.

mporkola:
This should increase front page performance pretty easily. Performance testing with siege on `testing` gave an increase of 1,08 -> 2,02 trans/sec, details below:

Before:
```
Transactions:		         194 hits
Availability:		      100.00 %
Elapsed time:		      179.16 secs
Data transferred:	        1.86 MB
Response time:		        4.03 secs
Transaction rate:	        1.08 trans/sec
Throughput:		        0.01 MB/sec
Concurrency:		        4.36
Successful transactions:         194
Failed transactions:	           0
Longest transaction:	        9.96
Shortest transaction:	        1.43
```

After:
```
Transactions:		         364 hits
Availability:		      100.00 %
Elapsed time:		      179.93 secs
Data transferred:	        3.38 MB
Response time:		        1.94 secs
Transaction rate:	        2.02 trans/sec
Throughput:		        0.02 MB/sec
Concurrency:		        3.92
Successful transactions:         364
Failed transactions:	           0
Longest transaction:	       10.77
Shortest transaction:	        0.74
```

### DIFF
--- a/app/views/homepage/_grid_item.haml
+++ b/app/views/homepage/_grid_item.haml
@@ -1,4 +1,4 @@
-- cache(listing) do
+- cache(["grid_item", listing, listing.author, @current_community]) do
   .home-fluid-thumbnail-grid-item
     %div
       -# Listing image
--- a/app/views/homepage/_grid_item.haml
+++ b/app/views/homepage/_grid_item.haml
@@ -1,11 +1,12 @@
-.home-fluid-thumbnail-grid-item
-  %div
-    -# Listing image
-    = render :partial => "layouts/grid_item_listing_image", :locals => {:listing => listing, :modifier_class => ""}
+- cache(listing) do
+  .home-fluid-thumbnail-grid-item
+    %div
+      -# Listing image
+      = render :partial => "layouts/grid_item_listing_image", :locals => {:listing => listing, :modifier_class => ""}
 
-    -# Listing author details
-    .home-fluid-thumbnail-grid-author
-      .home-fluid-thumbnail-grid-author-avatar
-        = small_avatar_thumb(listing.author, :class => "home-fluid-thumbnail-grid-author-avatar-image")
-      = link_to(listing.author, :class => "home-fluid-thumbnail-grid-author-name") do
-        = listing.author.name(@current_community)
+      -# Listing author details
+      .home-fluid-thumbnail-grid-author
+        .home-fluid-thumbnail-grid-author-avatar
+          = small_avatar_thumb(listing.author, :class => "home-fluid-thumbnail-grid-author-avatar-image")
+        = link_to(listing.author, :class => "home-fluid-thumbnail-grid-author-name") do
+          = listing.author.name(@current_community)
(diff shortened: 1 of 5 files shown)


app/views/homepage/_grid_item.haml
```diff
@@ -1,11 +1,12 @@
-.home-fluid-thumbnail-grid-item
-  %div
-    -# Listing image
-    = render :partial => "layouts/grid_item_listing_image", :locals => {:listing => listing, :modifier_class => ""}
+- cache(listing) do
```
mporkola: This way the cache key is generated by combining the model name, id and `updated_at` timestamp, as described here: [Fragment caching](http://guides.rubyonrails.org/caching_with_rails.html#fragment-caching)

> Behind the scenes, a method called cache_key will be invoked on the model and it returns a string like products/23-20130109142513. The cache key includes the model name, the id and finally the updated_at timestamp. Thus it will automatically generate a new fragment when the product is updated because the key changes.

app/views/homepage/_grid_item.haml
```diff
@@ -1,11 +1,12 @@
...
+  .home-fluid-thumbnail-grid-item
+    %div
+      -# Listing image
+      = render :partial => "layouts/grid_item_listing_image", :locals => {:listing => listing, :modifier_class => ""}
 
-    -# Listing author details
-    .home-fluid-thumbnail-grid-author
-      .home-fluid-thumbnail-grid-author-avatar
-        = small_avatar_thumb(listing.author, :class => "home-fluid-thumbnail-grid-author-avatar-image")
```
sktoiva: Hm, what happens if avatar is updated and listing is not touched?

app/views/homepage/_grid_item.haml
```diff
@@ -1,11 +1,12 @@
...
-      = link_to(listing.author, :class => "home-fluid-thumbnail-grid-author-name") do
-        = listing.author.name(@current_community)
+      -# Listing author details
+      .home-fluid-thumbnail-grid-author
+        .home-fluid-thumbnail-grid-author-avatar
+          = small_avatar_thumb(listing.author, :class => "home-fluid-thumbnail-grid-author-avatar-image")
+        = link_to(listing.author, :class => "home-fluid-thumbnail-grid-author-name") do
```
sktoiva: After this change, changing usernames is not possible anymore, or it'll break these links. Or at least without invalidating cache.

app/views/homepage/_grid_item.haml
ovan: I would say we cannot release with a possibility of that happening. We need to make sure we add touching from users to listings. Maybe active record model touching is enough but if not then we just need to explicitly cover all update paths to relevant users data (avatar icon, given name, family name, username, anything else?).

All in all I feel this is a place where the added complexity of handling the cache invalidation pays off because the frontpage is still quite heavy even after the recent query and db optimizations and majority of the traffic comes from frontpage. Great stuff @mporkola!

mporkola:
The cache key is now formed from `listing`, `listing.author` and `@current_community`, which should cover all invalidation cases. Performance didn't take a hit:

```
Transactions:		         399 hits
Availability:		      100.00 %
Elapsed time:		      179.68 secs
Data transferred:	        3.82 MB
Response time:		        1.74 secs
Transaction rate:	        2.22 trans/sec
Throughput:		        0.02 MB/sec
Concurrency:		        3.86
Successful transactions:         399
Failed transactions:	           0
Longest transaction:	        7.93
Shortest transaction:	        0.80
```

mporkola:
Still one caveat: expiring the fragment when template is updated. `cache-digest` gem should help

PR #1257, opened 2015-06-02
Close listing updated at

rap1ds:
- Add `HashUtils#flatten`
- Add `SQLUtils#hash_to_query`
- Use `hash_to_query` for `update_all`
- Fixes bug: Closing listings doesn't update the `updated_at` timestamp

### DIFF
--- a/app/utils/hash_utils.rb
+++ b/app/utils/hash_utils.rb
@@ -118,7 +118,7 @@ module HashUtils
         key_s = k.to_s
 
         if !k.is_a?(Symbol) || key_s.include?(".")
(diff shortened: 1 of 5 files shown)


app/utils/hash_utils.rb
```diff
@@ -109,4 +109,28 @@ def transpose(x)
...
   end
+
+  # { a: b: 1 } -> { :"a.b" => 1 }
+  def flatten(h)
+    # use helper lambda
+    acc = ->(prefix, hash) {
+      hash.inject({}) { |memo, (k, v)|
+        key_s = k.to_s
+
+        if !k.is_a?(Symbol) || key_s.include?(".")
```
sktoiva: Text says key must be a String even though the test is that key must be a symbol.

PR #1267, opened 2015-06-03
Ignore empty order_id in auth ipn msg parsing

ovan:
When we switched from Order based paypal flow to auth based flow we no longer have orders and order ids in payments. One change that went unnoticed was an IPN message informing about authorization no longer has parent transaction id. This was incorrectly parsed as an empty string instead of nil. Because of this the payment update contained extra data and tried to insert a new payment with already existing authorization id. DB rejected this update because of unique index, and was correct in doing so. This parsing change now recognized nil order id and removes incorrect extra payment update.

https://sharetribe.airbrake.io/projects/15448/groups/1389590260555809569/notices/1434356405166769802

### DIFF
--- a/app/services/paypal_service/data_types/ipn.rb
+++ b/app/services/paypal_service/data_types/ipn.rb
@@ -248,7 +248,7 @@ module PaypalService
           },
           params)
 
-        p[:order_id] = p[:order_id].strip.empty? ? nil : p[:order_id]
+        p[:order_id] = Maybe(p)[:order_id].strip.or_else(nil)
 
         create_authorization_created(
           p.merge({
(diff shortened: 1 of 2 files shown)


app/services/paypal_service/data_types/ipn.rb
```diff
@@ -248,6 +248,8 @@ def to_authorization_created(params)
           },
           params)
 
+        p[:order_id] = p[:order_id].strip.empty? ? nil : p[:order_id]
```
rap1ds: Empty string maps to None, so you could also use:

`Maybe(p)[:order_id].strip.or_else(nil)`

rap1ds:
:+1:

PR #1270, opened 2015-06-03
Enable quantity selectors for free transactions

rap1ds:
Merge after #1269 

All new transactions now go through the same TransactionsController#new action. In theory, listing show page doesn't now need to know details (process, payment gateways, etc.).

- [X] All transactions go through TransactionsController#new
- [X] Free transactions are rendered by TransactionsController
- [X] Other transactions are redirected to correct path
- [X] Selectors work with free transactions

### DIFF
--- a/app/assets/javascripts/transaction.js
+++ b/app/assets/javascripts/transaction.js
@@ -1,6 +1,8 @@
 window.ST = window.ST || {};
 
-window.ST.transaction = (function(_) {
+window.ST.transaction = window.ST.transaction || {};
+
+(function(module, _) {
 
   function toOpResult(submitResponse) {
     if (submitResponse.op_status_url) {
@@ -104,10 +106,8 @@ window.ST.transaction = (function(_) {
 
   }
 
-  return {
-    initializePayPalBuyForm: initializePayPalBuyForm,
-    initializeCreatePaymentPoller: initializeCreatePaymentPoller,
-    initializeFreeTransactionForm: initializeFreeTransactionForm
-  };
+  module.initializePayPalBuyForm = initializePayPalBuyForm;
+  module.initializeCreatePaymentPoller = initializeCreatePaymentPoller;
+  module.initializeFreeTransactionForm = initializeFreeTransactionForm;
 
-})(_);
+})(window.ST.transaction, _);
--- a/app/assets/javascripts/transaction.js
+++ b/app/assets/javascripts/transaction.js
@@ -1,6 +1,6 @@
 window.ST = window.ST || {};
 
-ST.transaction = (function(_) {
+window.ST.transaction = (function(_) {
 
   function toOpResult(submitResponse) {
     if (submitResponse.op_status_url) {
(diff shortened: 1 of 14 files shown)


app/controllers/transactions_controller.rb
```diff
@@ -10,6 +10,59 @@ class TransactionsController < ApplicationController
 
   MessageForm = Form::Message
 
+  def new
+    Result.all(
+      ->() {
+        binding.pry
```
ovan: binding.pry

app/controllers/transactions_controller.rb
```diff
@@ -10,6 +10,59 @@ class TransactionsController < ApplicationController
...
+        redirect_to post_pay_listing_path(:listing_id => listing_model.id)
+      else
+        params = "listing_id: #{listing_id}, payment_gateway: #{gateway}, payment_process: #{process}, booking: #{booking}"
+        raise ArgumentError.new("Can not find new transaction path to #{params}")
+      end
+    }.on_error { |error_msg|
+      flash[:error] = "Could not start transaction, error message: #{error_msg}"
+      redirect_to root_path
+    }
+  end
```
ovan: Can you use helper methods instead of having all this in one controller action. Now it's hard to quickly get the high level idea what's done here because of all the details. How about e.g. extracting the lambda function contents to helper methods?

app/services/result.rb
```diff
@@ -106,4 +106,25 @@ def maybe()
...
 
+  module_function
+
+  def all(*operations)
+    operations.inject(Result::Success.new([])) { |res, op|
+      if res.success
+        res_data = res.data
+        op_res = op.call(*res_data)
+
+        raise ArgumentError.new("Lambda must return Result") unless (res.is_a?(Result::Success) || res.is_a?(Result::Error))
```
rap1ds: must be `op_res`

app/assets/javascripts/transaction.js
```diff
@@ -1,6 +1,6 @@
 window.ST = window.ST || {};
 
-ST.transaction = (function(_) {
+window.ST.transaction = (function(_) {
```
Gnito: I usually prefer:
```
    (function(module) {
      //...
    })(window.ST)
```

app/assets/javascripts/transaction.js
```diff
@@ -89,9 +89,25 @@ ST.transaction = (function(_) {
     ).onValue(function () { window.location = redirectUrl; });
   }
 
+  function initializeFreeTransactionForm(locale) {
+    window.auto_resize_text_areas("text_area");
+    $('textarea').focus();
```
Gnito: "textarea" vs. "text_area"?
Could this function get the element as parameter instead of class search from the whole dom?

app/controllers/transactions_controller.rb
```diff
@@ -1,15 +1,104 @@
...
+      ->((listing_id, listing_model)) {
+        ensure_can_start_transactions(listing_model: listing_model, current_user: @current_user, current_community: @current_community)
+      }
+    ).on_success { |((listing_id, listing_model, author_model, process, gateway))|
+      booking = listing_model.unit_type == :day
+
+      transaction_params = {listing_id: listing_model.id}.merge(params.slice(:start_on, :end_on, :quantity))
+
+      case [process[:process], gateway, booking]
+      when matches([:none])
```
Gnito: This was just moved from one place to another, so it's ok as it is.
(But I find it a bit too heavy to follow)

Gnito:
couple of comments, but otherwise I guess it looks ok at the moment.

PR #1271, opened 2015-06-03
Fixed to order type copys

Kusti:
Minor edits & improvements to order type copys.

### DIFF
--- a/app/views/admin/listing_shapes/index.haml
+++ b/app/views/admin/listing_shapes/index.haml
@@ -38,7 +38,7 @@
         - elsif shape[:category_ids].size == category_count
           = t("admin.listing_shapes.index.all_categories")
         - else
-          = t("admin.listing_shapes.index.category_count", :count => "#{shape[:category_ids].size}/#{category_count}")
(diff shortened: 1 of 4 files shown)


app/views/admin/listing_shapes/index.haml
```diff
@@ -38,7 +38,7 @@
         - elsif shape[:category_ids].size == category_count
           = t("admin.listing_shapes.index.all_categories")
         - else
-          = "#{shape[:category_ids].size}/#{category_count}"
+          = t("admin.listing_shapes.index.category_count", :count => "#{shape[:category_ids].size}/#{category_count}")
```
rap1ds: I guess this works ok, but I'd be a careful using the variable name `count` here. `count` has a special meaning for pluralizations:

```yml
apples:
  one: %{count} apple
  other: %{count} apples
```

```ruby
t('apples', count: 5) #=> 5 apples
```

http://guides.rubyonrails.org/i18n.html#pluralization

rap1ds:
Didn't test this but code looks ok. I'd change the `count`.

PR #1275, opened 2015-06-08
Add lograge and change logs to json

sktoiva:
For Loggly to be more effective, change logging to json.


config/environments/staging.rb
```diff
@@ -26,14 +26,26 @@
...
+  end
+  config.lograge.formatter = Lograge::Formatters::Json.new
+
   # Use a different cache store in production
-  config.cache_store = :dalli_store, (ENV["MEMCACHIER_SERVERS"] || "").split(","), {
-    username: ENV["MEMCACHIER_USERNAME"],
-    password: ENV["MEMCACHIER_PASSWORD"],
+  config.cache_store = :dalli_store, (ENV["MEMCACHIER_GREEN_SERVERS"] || "").split(","), {
+    username: ENV["MEMCACHIER_GREEN_USERNAME"],
+    password: ENV["MEMCACHIER_GREEN_PASSWORD"],
```
ovan: plop

config/environments/staging.rb
```diff
@@ -26,14 +26,26 @@
...
     failover:  true,
     socket_timeout: 1.5,
     socket_failure_delay:  0.2,
-    namespace: "sharetribe-staging",
+    namespace: "sharetribe-production",
```
ovan: plop 2

app/services/paypal_service/logger.rb
```diff
@@ -2,29 +2,28 @@ module PaypalService
   class Logger
 
     def info(msg)
-      #TODO: Fix logging totally
-      Rails.logger.warn "[Paypal] #{msg}"
+      Rails.logger.info ({ paypal: msg }.to_json)
```
rap1ds: Is there a reason why this is a bit inconsistent compared to other log levels? Shouldn't it be `Rails.logger.warn ({ paypal: { type: :info, message: msg } }.to_json)`

config/environments/development.rb
```diff
@@ -15,6 +15,37 @@
   # Log error messages when you accidentally call methods on nil.
   config.whiny_nils = true
 
+  config.log_level = :debug
+
+  # Basic log config, for calls to Rails.logger.<level> { <message> }
+  config.logger = ::Logger.new(STDOUT)
+  config.logger.formatter = ->(severity, datetime, progname, msg) { "#{severity[0]} #{msg}\n" }
```
rap1ds: Do we always have `severity` present? No need to worry about null pointer errors?

config/environments/development.rb
```diff
@@ -15,6 +15,37 @@
...
+
+  # Lograge config, overrides default instrumentation for logging ActionController and ActionView logging
+  config.lograge.enabled = true
+  config.lograge.custom_options = lambda do |event|
```
rap1ds: I'd prefer the new lambda syntax. You are already using the new syntax in line 22, so maybe this could be changed for the sake of consistency?

app/services/paypal_service/logger.rb
sktoiva: Because this is used for the requests and responses from paypal, and they contain json, not a string message.

We could of course create more consistent schema with e.g.:
```
 { paypal: { type: :info, message: nil, content: <request hash> }
 { paypal: { type: :error, message: "error message", content: nil }
```

config/environments/development.rb
```diff
@@ -15,6 +15,37 @@
...
+    params = event.payload[:params].reject do |k|
```
rap1ds: This is ok, no need to change, but also `expect` could be used here I guess.

config/environments/development.rb
```diff
@@ -15,6 +15,37 @@
...
+      ['controller', 'action'].include? k
+    end
+
+    { params: params,
```
rap1ds: How about filtering the parameter, like passwords?

app/services/paypal_service/logger.rb
rap1ds: Maybe this could be renamed to something else, something more general, like `log_json`, `log_paypal_message` e.g.

rap1ds:
Since we want to encourage JSON logging when ever possible, it would be cool to have a couple of helpers to make it easier, e.g. something like this:

```ruby
# /app/some_place_where_we_put_helpers/json_logger.rb

class JsonLogger
  def initialize(community_id:, person_id:)
    @community_and_person = {community_id: community_id, person_id: person_id}
  end

  def debug
    ...
  end

  def info(obj)
    obb.merge(@community_and_person).to_json
  end
end

# /app/application_controller.rb

  ...

  def json_log(obj)
    @json_logger ||= JsonLogger.new(community_id: Maybe(@current_community).id, person_id: Maybe(@current_user).id)
    @json_logger.log(obj)
  end

  def json_warn(obj)
    ...
  end
end

config/environments/development.rb
sktoiva: Shouldn't be worries with this, as the logger is always invoked with `log.<debug|info|warn|error|fatal>` which add's the severity.

PR #1278, opened 2015-06-09
Handle missing category translation

rap1ds:
Steps to reproduce

1. Add a new category, which uses only one order type
1. Add a new language and use that language
1. Go to order type admin and try to edit the order type which is the only one for the newly added category

Actual: Error, because the newly added category doesn't have translation for the newly added language

Solution: Use the first translation if nothing else is not available


sktoiva:
:+1:

PR #1279, opened 2015-06-09
Increase limits on customization page content

ovan:
* Change column types in db text -> medium text to hold at most 16777215
chars.
* Use rails validations to set lower limits for users to disallow excess
data usage. The limits can be hit e.g. when users incorrectly paste
content from Word documents.


rap1ds:
:+1:

PR #1280, opened 2015-06-09
Remove units from shapes without price migrations

rap1ds:
- We shouldn't have units for shapes that do not have price enabled. However, probably due to earlier migrations, this has happened. This PR fixes the issue by removing units from shapes that do not have price enabled.


rap1ds:
Most like the broken data in DB is due to our manual changes to the DB before Order Types UI. Thus there's no need for a migration, instead the data is fixed manually straight to the DB.

PR #1283, opened 2015-06-10
Fix failed TimeUtils test

bingxie:
expect(TimeUtils.time_to(30.days.since(now), now)).to eq({unit: :days, count: 30})

This one will fail at my timezone. So I just get rid of the timezone from 'now' to make it passed.


mporkola:
Thanks for reporting this issue! We want to keep timezones in the test, so we fixed it a bit differently: https://github.com/sharetribe/sharetribe/pull/1367

PR #1284, opened 2015-06-11
Fix pluralization error

rap1ds:
- This error occures because these keys were already created, but later we
added the pluralization there (one, other etc.)
- For some reason, WTI can not handle this case. If we add pluralization to
existing key, we need to rename the key.
- The issue can be reproduced in production: Choose German and go to Order
Type UI


ovan:
Ok. Good to know that we can't pluralize an existing key. Can you shout this couple more times in internal chat to make sure everyone is aware?

PR #1285, opened 2015-06-11
Make community or display type mandatory

rap1ds:
When we display users name, we should always consider communities setting for the name display format, which can be eg. full_name or first_name_with_initials.

However, that's easy to miss, since Person#name method makes it optional to pass the community or display_name_format to the method.

This pull request changes it so that it's mandatory to pass the community of display_name_format

### DIFF
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -25,7 +25,7 @@ class Feedback < ActiveRecord::Base
   def author_name_and_email
     email_in_brackets = email.present? ? "(#{email})" : "No email provided"
     if author
-      "#{author.name} #{email_in_brackets}"
+      "#{author.name('first_name_with_initial')} #{email_in_brackets}"
     elsif !email.blank?
       "Unlogged user #{email_in_brackets}"
     else
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -288,6 +288,10 @@ class Person < ActiveRecord::Base
   # Deprecated: This is view logic (how to display name) and thus should not be in model layer
   # Consider using PersonViewUtils
   def name(community_or_display_type=nil)
+    # TODO Make `community_or_display_type` param mandatory.
+    # The custom error is here in order to get better error message while, but it can be removed in the future.
+    raise ArgumentError.new("Missing community or display type. Can not know how to format the name") if community_or_display_type.nil?
+
     return name_or_username(community_or_display_type)
   end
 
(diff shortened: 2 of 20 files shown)


app/models/person.rb
```diff
@@ -288,6 +288,10 @@ def first_name_with_initial
   # Deprecated: This is view logic (how to display name) and thus should not be in model layer
   # Consider using PersonViewUtils
   def name(community_or_display_type=nil)
+    # TODO Make `community_or_display_type` param mandatory.
+    # The custom error is here in order to get better error message while, but it can be removed in the future.
+    raise ArgumentError.new("Missing community or display type. Can not know how to format the name") if community_or_display_type.nil?
+
```
ovan: What's the idea behind defaulting the parameter to nil but at the same time making it mandatory with explicit check?

app/models/feedback.rb
```diff
@@ -25,7 +25,7 @@ class Feedback < ActiveRecord::Base
   def author_name_and_email
     email_in_brackets = email.present? ? "(#{email})" : "No email provided"
     if author
-      "#{author.name} #{email_in_brackets}"
+      "#{author.name('first_name_with_initial')} #{email_in_brackets}"
```
ovan: This type of "selecting one of enumerated options" usecase should always use symbols and never plain strings. It documents the case better, makes comparison operations more efficient and uses less memory.

app/models/person.rb
rap1ds: Just to get better error messages. I have a hunch that when we deploy this to production, we will get a couple of those errors and that's why it's nicer to have better and more explicit error message than "Wrong number of arguments 0 for 1)

app/models/feedback.rb
rap1ds: I definitely agree, but I'm not changing it right now.

app/models/person.rb
ovan: I can see your point but it leaves us with method signature that lies to user. From "code that documents itself" point of view method signature is very important. I would guess I look at that before looking at any comments or the method body, and probably often times only look at the signature and only study the body if I don't feel like I can guess what the method does from name, context and parameters.

rap1ds: The idea was to remove it a bit later. However, I can definitely see myself forgetting to remove this, and it will stay there forever. So, I guess the safest bet is just remove it immediately. I'll do that.

PR #1287, opened 2015-06-11
Fix frontpage cache

mporkola:
Changes in .yml translations and user locale now affect cache key. Fragments have an expiration time of 4 hours to ensure all changes are at least eventually visible.

Remaining issues:
* Routes changes still break stuff if the cached fragments link to an old route.
* In Memcached you can only delete a single item or the entire cache, so invalidating things manually is hard.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -889,12 +889,4 @@ module ApplicationHelper
       content_for :extra_javascript do js end
     end
   end
-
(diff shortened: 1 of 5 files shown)


rap1ds:
I took a very quick look and this looks ok. :+1:

mporkola:
Great. Let's still not merge this before holidays :fearful:

PR #1288, opened 2015-06-12
Allow trial marketplace admins to delete their marketplace

rap1ds:
- [ ] Add admin page UI under "Settings" where admin can delete the marketplace
- [ ] Allow only if free plan

screen shot 2015-06-12 at 14 08 32

### DIFF
--- a/app/controllers/admin/communities_controller.rb
+++ b/app/controllers/admin/communities_controller.rb
@@ -106,13 +106,11 @@ class Admin::CommunitiesController < ApplicationController
   def settings
     @selected_left_navi_link = "admin_settings"
 
-    can_delete_marketplace = MarketplaceService::Community::Query.current_plan(@current_community.id).plan_level == CommunityPlan::FREE_PLAN
-
     render :settings, locals: {
              supports_escrow: escrow_payments?(@current_community),
              delete_redirect_url: delete_redirect_url(APP_CONFIG),
              delete_confirmation: @current_community.ident,
-             can_delete_marketplace: can_delete_marketplace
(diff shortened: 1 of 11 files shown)


config/locales/en.yml
```diff
@@ -203,6 +203,15 @@ en:
         automatic_newsletter_frequency: "Send automatic newsletter: %{frequency_dropdown}"
         newsletter_daily: Daily
         newsletter_weekly: Weekly
+        delete_marketplace_title: "Delete marketplace"
+        type_marketplace_domain: "Type your marketplace domain (%{domain}) to the text field below:"
+        type_marketplace_domain_placeholder: "Type your marketplace domain here"
+        once_you_delete: "Once you delete the marketplace, you will not be able to access it anymore. Be careful."
+        are_you_sure: "Are you sure?"
+        i_understand_button: "I understand that by clicking this button my marketplace will be deleted"
+        last_community_updates: "Please note that after deletion, you may still receive the last marketplace updates email."
```
sktoiva: Should this be:

Please note that after deletion, you and your customers may still receive the last marketplace update emails.

app/controllers/admin/communities_controller.rb
```diff
@@ -184,6 +192,18 @@ def update_settings
             :settings)
   end
 
+  def delete_marketplace
+    if params[:delete_confirmation] == @current_community.ident
```
ovan: Can you double check the "can_delete_marketplace" condition here

config/locales/en.yml
rap1ds: :+1:

PR #1290, opened 2015-06-12
Recognize nil order_ids for auth flow events

ovan:
When we switched from the order flow to authorization flow authorization
related IPN messages no longer have parent_txn_id set because they don't
have a parent transaction. The parent transaction used to be the id of
the order. IPN sends this field non the less but with empty string
value. We didn't recognize the empty field to mean nil which caused an
attempt to update an incorrect payment at later stage of handling the
IPN message. Database prevents this with unique id constraint but this
causes an unnecessary error to pop up in airbrake. Also this would mean
we lose data if the web transaction flow would fail in the middle of
receiving success response from PayPal server. Properly handled IPN
message will reconcilate the state of the payment.

### DIFF
--- a/app/services/paypal_service/data_types/ipn.rb
+++ b/app/services/paypal_service/data_types/ipn.rb
@@ -322,6 +322,8 @@ module PaypalService
           params
         )
 
+        p[:order_id] = Maybe(p)[:order_id].strip.or_else(nil)
+
         create_payment_voided(p)
       end
       private_class_method :to_payment_voided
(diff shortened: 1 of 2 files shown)


app/services/paypal_service/data_types/ipn.rb
```diff
@@ -322,6 +322,8 @@ def to_payment_voided(params)
           params
         )
 
+        p[:order_id] = Maybe(p)[:order_id].strip.or_else(nil)
```
Gnito: Is it possible that order_id is "" after strip? Does it matter?

Gnito:
:+1:

app/services/paypal_service/data_types/ipn.rb
ovan: It is "" after strip but maybe takes care of that can goes to else-branch

PR #1296, opened 2015-06-16
Allow admins to view conversations

rap1ds:


app/controllers/transactions_controller.rb
```diff
@@ -100,10 +100,28 @@ def create
...
+      community_id: @current_community.id)
+
+    admin_transaction_conversation =
+      if is_admin && feature_enabled?(:admin_conversations) && participant_transaction_conversation.nil?
+        MarketplaceService::Transaction::Query.transaction_with_conversation(
+          transaction_id: params[:id],
+          community_id: @current_community.id)
+      else
+        nil
+      end
```
ovan: Could this be written more succinctly by chaining Maybe when the if has participant_transaction_conversation.nil? check anyway?

app/services/marketplace_service/transaction.rb
```diff
@@ -253,13 +253,20 @@ def transaction(transaction_id)
...
           .includes(:booking)
-          .where("starter_id = ? OR listings.author_id = ?", person_id, person_id)
-          .first)
+
+        with_person =
+          if person_id
+            rel.where("starter_id = ? OR listings.author_id = ?", person_id, person_id)
+          else
+            rel
+          end
```
ovan:
```
Maybe(person_id)
  .map { |person_id| rel.where("starter_id = ? OR listings.author_id = ?", person_id, person_id) }
  .or_else(rel)
```

rap1ds: Actually, this didn't quite work. 

The fact that the `rel` is a magical ActiveRecord relationship made this a bit tricky. After mapping, Maybe tries to define whether the value is a non-value (None) or a value (Some). The way it tests it is to see if the value is null, or if the length of the value is 0. Asking the length from the `rel` runs the query, and that's not actually what we wanted. Running the query returns an empty array and that maps to None, and then the `or_else` is run.

The solution was to wrap the `rel` in another array.

PR #1300, opened 2015-06-25
Two-way market support

jeffstrickler:
This branch adds the ability to request a service and pay someone upon fulfillment of the request.

### DIFF
--- a/app/models/listing.rb
+++ b/app/models/listing.rb
@@ -66,6 +66,8 @@ class Listing < ActiveRecord::Base
   include ApplicationHelper
   include ActionView::Helpers::TranslationHelper
   include Rails.application.routes.url_helpers
+  
+  belongs_to :listing_shape
 
   belongs_to :author, :class_name => "Person", :foreign_key => "author_id"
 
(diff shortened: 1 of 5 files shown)


rap1ds:
Hi! 

Thanks for the pull request!

Two-way marketplaces (or reverse marketplaces) are definitely interesting! We plan to support those in the future for sure. 

The way how we have planned to implement it requires a bit more work. For example, I think the way how users would like to use a reverse marketplace is the TaskRabbit style "I need someone to do task X, give me an offer", and after that the requester receives a number of offers and picks one based on reviews and price. Implementing that kind of bidding system with payments requires some additional work, which unfortunately is not on the road map for near future.

Because of that I'm closing this Pull Request without merging it.

In addition, we are planning to do some heavy changes to the Sharetribe codebase and architecture. That's also a reason why we kind of avoid merging pull requests that are not direct bug fixes, but add a new features. 

Anyway, thanks for you pull request! It's exising to see that people are interested our platform and developing it further! And sorry about the fact that we can not merge your contribution to the master branch right now.

PR #1303, opened 2015-06-29
Fix find paypal payment

sktoiva:
Payment should be fetched directly from store and not use lookup. Lookup will inject merchant and therefore cause unwanted functionality.


rap1ds:
:+1:

PR #1304, opened 2015-06-29
Always show sum and status

rap1ds:
**Background:** In the past we made a design decision that we do not show the "Status" and "Sum" columns in the admin transaction view (see the screenshot) if the marketplace doesn't use payments (i.e. doesn't have any payment gateway configured). However, we lately added a feature for marketplace admins to see the conversation happening in the marketplace. The "Status" column link is how the admins can access the conversations. The problem is that the marketplaces without payments can't access the conversation, because the link is not visible.

screen shot 2015-06-29 at 11 21 30

**This Pull Request** removes the `if` clause that hided the "Status" and "Sum". I see three benefits:

- Now marketplaces without payments can access conversations
- Reduces "product" complexity (hard to explain, but I mean that the product it self is less complex, when all admins see the same thing, regardless of their settings.)
- Currently, with Order Types, it's not so easy to define if the marketplace has payments enabled or not. The marketplace may have PayPal configured, but no order type is using online payments. In addition, there might be transactions that were using payments before admin decided to remove online payments.

### DIFF
--- a/app/controllers/admin/community_transactions_controller.rb
+++ b/app/controllers/admin/community_transactions_controller.rb
@@ -54,7 +54,6 @@ class Admin::CommunityTransactionsController < ApplicationController
 
     render("index",
       { locals: {
-        show_status_and_sum: @current_community.payments_in_use?,
         community: @current_community,
         conversations: conversations
       }}
--- a/app/views/admin/community_transactions/index.haml
+++ b/app/views/admin/community_transactions/index.haml
@@ -15,9 +15,8 @@
     %thead
       %tr
         %th= render partial: "layouts/sort_link", locals: { column: "listing", direction: sort_link_direction("listing"), title: t("admin.communities.transactions.headers.listing") }
-        - if show_status_and_sum
-          %th=t("admin.communities.transactions.headers.status")
-          %th=t("admin.communities.transactions.headers.sum")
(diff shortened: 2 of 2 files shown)


app/controllers/admin/community_transactions_controller.rb
```diff
@@ -54,7 +54,6 @@ def index
 
     render("index",
       { locals: {
-        show_status_and_sum: @current_community.payments_in_use?,
```
ovan: Did you check if the payments_in_use? method is needed somewhere else? It feels like legacy logic that shouldn't exist anymore (I can't even guess what's the deciding factor with listing shapes / transaction processes / transaction settings).

ovan:
I think the design decision is very sound. +1

app/controllers/admin/community_transactions_controller.rb
rap1ds: It is still used in a few places. 

The logic is definitely wrong for asking whether payments are in use in this marketplace. RIght way to define that would be to see all the Order Types if any of them uses process != free.

The logic is right, if the question is "does this community have any payment gateways configured". And that question is also a "legacy" question, since all our new trials will have PayPal enabled even if they don't have any Order Type with Online payments. The only reason why we remove PayPal from a community is that they want to use currency that is not supported by PayPal.

PR #1311, opened 2015-07-01
Mark transactions deleted instead of removing them

ovan:
We stop permanently deleting transactions and instead introduce a flag, deleted, to differentiate between initiated and deleted transactions.

This PR stops us from losing data permanently. This has bitten us already several times when there's some unexpected behavior in the middle of the payment process. This won't solve those issues but leaves the data in place so we can manually detect what was going on and have a way to bring back the transaction in case of failure.

This affects multiple places and I've tried to cover them all. Luckily we have pretty good tests to cover the payment flow and messages. I'm still slightly concerned about touching the huge inbox SQL queries from perf POV. Locally everything seemed to be in order so need to just test in production and be ready to rollback in case of any issues.

### DIFF
--- a/app/services/transaction_service/paypal_events.rb
+++ b/app/services/transaction_service/paypal_events.rb
@@ -1,6 +1,5 @@
 module TransactionService::PaypalEvents
 
-  TransactionModel = ::Transaction
   TransactionStore = TransactionService::Store::Transaction
   module_function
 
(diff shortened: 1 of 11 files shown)


app/services/transaction_service/paypal_events.rb
```diff
@@ -122,11 +121,8 @@ def pending_ext_to_paid(tx)
   end
 
   def delete_transaction(cid:, tx_id:)
-    tx = TransactionModel.where(community_id: cid, id: tx_id, current_state: "initiated").first
-
-    if tx
-      tx.conversation.destroy
```
sktoiva: What happens now to conversations with transactions marked as deleted?

ovan: Nothing, they stay as they are so that we if decide to resurrect the transaction we haven't lost them. B conversations with transactions are only ever shown in the context of the transaction so they are filtered out from inbox etc. by looking at the transaction deleted flag.

PR #1316, opened 2015-07-03
Don't make transaction agreement mandatory on disabled locales.

mporkola:
When removing locales with the new locale admin UI, custom transaction agreement texts were still mandatory to fill. This caused user issues.


Gnito:
Looks good! (I didn't test though)

PR #1317, opened 2015-07-03
Handle paypal payment review state for payments

ovan:
We detected a new payment state coming from PayPal, payment-review. This happens before an authorization for a payment goes through. So far it seems this state if very rare so we only implement minimal handling for it by:

1. We detect the state from PayPal response and show a warning message to seller saying that payment has been submitted but it only will go through once PayPal approves it.
2. We keep the transaction in initiated state. This means it will not be visible in buyers or sellers inbox and no emails are sent about it at this stage.
3. When PayPal approves the payment we get an IPN message that payment has now been authorized. This PR also makes sure we handle it properly and allow payment to transition from payment-review to authorized. This causes transaction to transition from initiated to preauthorized and from there on the flow continues normally.

What the buyer sees when this happens:

screen shot 2015-07-03 at 15 46 43


mporkola:
Looks nice!

PR #1320, opened 2015-07-08
Don't remove token/tx while waiting payment review

ovan:
When payment goes to payment-review state we cannot authorize the payment before the review has been passed by PayPal. Our process deletes paypal retry tokens only after the associated payment has been authorized. Before that it keeps retrying. However, we have a time limit of 1 hour after which we stop retrying and decide that the payment has failed forever. This happens when e.g. user goes to paypal to pay but closes browser window and never enters payment details. The new payment-review state in the process is an exception case that this pull request now detects and ignores the one hour token delete limit to make sure we can wait for more than 1 hour for the payment review process to complete.

### DIFF
--- a/app/services/paypal_service/api/payments.rb
+++ b/app/services/paypal_service/api/payments.rb
@@ -253,7 +253,7 @@ module PaypalService::API
       TokenStore.get_all.each do |token|
         response = create(token.community_id, token.token)
 
-        if(!response[:success] && token.created_at < clean_time_limit)
+        if(!response[:success] && stop_retrying_token?(response, token.created_at, clean_time_limit))
           request_cancel(token.community_id, token.token)
         end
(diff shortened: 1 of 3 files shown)


app/services/paypal_service/api/payments.rb
```diff
@@ -266,6 +266,14 @@ def retry_and_clean_tokens(clean_time_limit)
     # Reusable bits of the API operations
     #
 
+    def stop_retrying_token?(response, token_created_at, clean_time_limit)
+      if (token_created_at < clean_time_limit && (response[:data] == nil || response[:data][:error_code] != :"payment-review"))
```
Gnito: Would it be better to return just the result of the comparison? (or use variable for the sake of readability and return that)

PR #1321, opened 2015-07-09
Don't expire OS installation marketplaces

ovan:
When the DB is empty (fresh OS installation) we trigger a form for creating the first marketplace. Currently filling the form creates a marketplace plan with expires_at time set to 31 days from now. This is the default behavior for trial marketplaces but not desired with OS installations. This PR adds a possibility to override the default behavior and never expire a marketplace and then uses this new behavior in OS marketplace creation form.


mporkola:
nice and clean :+1:

PR #1326, opened 2015-07-14
Prevent people from getting directed to sharetribe.com

sktoiva:
Now people are redirected to sharetribe.com if they mistakenly enter an address like `www.ident.sharetribe.com` instead of `ident.sharetribe.com`. They will also receive a certificate error in that case. This is a simple fix for that.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,6 +17,7 @@ class ApplicationController < ActionController::Base
   layout 'application'
 
   before_filter :force_ssl,
+    :redirect_to_marketplace_ident,
     :check_auth_token,
     :fetch_logged_in_user,
     :fetch_community,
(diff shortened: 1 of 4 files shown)


mporkola:
great!

app/controllers/application_controller.rb
```diff
@@ -253,6 +254,15 @@ def needs_redirect?(host, domain)
     domain.present? && host != domain
   end
 
+  def redirect_to_marketplace_ident
+    host = request.host
+    app_domain = URLUtils.strip_port_from_host(APP_CONFIG.domain)
+    if host.start_with?("www.") && host.end_with?(app_domain)
+      ident_without_www = host.sub(/www\./, '').chomp(".#{app_domain}")
+      return redirect_to "#{request.protocol}#{ident_without_www}.#{APP_CONFIG.domain}"
```
mporkola: [Google recommends](https://support.google.com/webmasters/answer/139066?hl=en#4) using a 301 redirect for non-canonical urls, `redirect_to` uses 302 as default: http://guides.rubyonrails.org/layouts_and_rendering.html#using-redirect-to

PR #1332, opened 2015-07-23
Check for recipient when user redirects to contact page after login

Gnito:


gnomet:
looks good, as long as 
´´´´´redirect_to (session[:return_to_content]
doesn't cause eternal loop?
as I understood that this is the case when we have already gone to :return_to_content?
but if it was tested in practice, probably works. (might be that :return_to_content is already cleared at this point.

Gnito:
yeah, it works - but maybe it's better to just use root (or profile page)

PR #1334, opened 2015-07-24
Fix updating comments count

bingxie:
Add the missing #id and use counter_cache for better performance.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -8,6 +8,7 @@ class CommentsController < ApplicationController
 
   def create
     if @comment.save
+      @comment.reload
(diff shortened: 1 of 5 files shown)


app/controllers/comments_controller.rb
```diff
@@ -8,6 +8,7 @@ class CommentsController < ApplicationController
 
   def create
     if @comment.save
+      @comment.reload
```
mporkola: This is not required, since `CommentCreatedJob` isn't dependent on the saved object, and will fetch the object on perform.

mporkola:
Please see the comment and rebase to current master, so we can merge this.

mporkola:
Actually we did it ourselves in https://github.com/sharetribe/sharetribe/pull/1368, thanks!

PR #1335, opened 2015-07-25
Delete a comment

bingxie:
The comment's author and admin can delete the comment.

### DIFF
--- a/config/locales/zh.yml
+++ b/config/locales/zh.yml
@@ -1353,6 +1353,8 @@ zh:
     comment:
       wrote: 写到
       send_private_message: "发送私人消息给%{person}"
+      delete: 删除
+      are_you_sure: 你确定要删除吗？
     comment_form:
       ask_a_question: 评论该条信息或者询问更多细节。所有其他用户都能看到你的评论。
       log_in: 登录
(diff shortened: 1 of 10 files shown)


mporkola:
Thanks for the PR! We'll have to think about the UI changes, and we'll get back to you. At least the alert dialog will have to be changed, as it's something we're not currently using elsewhere.

config/locales/zh.yml
```diff
@@ -1353,6 +1353,8 @@ zh:
     comment:
       wrote: 写到
       send_private_message: "发送私人消息给%{person}"
+      delete: 删除
+      are_you_sure: 你确定要删除吗？
```
rap1ds: All language changes and additions (expect English) should be done with WebTranslateIt. Otherwise, the WTI tool may override the local changes.

So what you should do is to translate these two keys from WTI, after the feature has been merged to master.

bingxie: yes, i see.

ovan:
Hi @bingxie. This got merged in another branch where I just rebased this version and removed the chinese translation because it has to come through WTI. Thanks for the contribution again and really sorry for the long delay!

PR #1336, opened 2015-07-27
Do not delete CommunityMembership when user is deleted

rap1ds:
**Background:** The data export script uses community_memberships table to define which users to export. When a user deletes his or her account also the membership is deleted. Thus, the export script doesn't know that this user once belonged to the community.

Now when another user who has had interaction with the deleted user goes to transaction view, the view breaks because the export script wasn't able to copy the deleted user.

**Steps:**

Prerequisite: Two existing users (Joe and Paul)

1. Log in as Joe
1. Send a message to Paul
1. Delete own (Joe's) account
1. Run the data export script and setup new marketplace from the data dump (i.e. run the script locally and change your database.yml to use `dump` database)
1. Log in as Paul (after resetting the password)
1. Go to inbox and transaction view to see the message from Paul

Expected: Paul sees conversation with "Removed user"

Actual: Site breaks

**Solution:** This pull request changes the behaviour so that the memberhips is not deleted when user deletes his or her account. Instead, the state is changed to "deleted_user".

### DIFF
--- a/app/models/community_membership.rb
+++ b/app/models/community_membership.rb
@@ -22,7 +22,7 @@
 
 class CommunityMembership < ActiveRecord::Base
 
-  VALID_STATUSES = ["accepted", "pending_email_confirmation", "banned"]
+  VALID_STATUSES = ["accepted", "pending_email_confirmation", "banned", "deleted_user"]
 
   belongs_to :person
   belongs_to :community, :counter_cache => :members_count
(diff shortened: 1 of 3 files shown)


gnomet:
Code changes and logic look good. Also assuming that member list doesn't show "deleted_user" memberships?

Also, see why travis goes to red, but otherwise this is good to go in my opinion.

PR #1340, opened 2015-07-28
Quantity label for too long labels.

Gnito:


Gnito:
Using wrapper for label, so that we can style the label element correctly with paddings or what ever is needed.

PR #1341, opened 2015-07-29
Add a new 'redirect-to-domain' flag

rap1ds:
This Pull Request adds a new boolean column `redirect_to_domain` to communities table. When the column value is `true`, the user accessing the marketplace with marketplace ident in the URL (*.sharetribe.com subdomain) will be redirected to the full marketplace domain, if present. Before this change, the users were redirected to full domain always if it was present.

**Why?** Domain validation: This change is needed for SSL file verification. The SSL provider needs to be able to access the application with the full domain, thus we have to add the full domain to the database before we have a valid SSL cert. So we want to permit the access to the marketplace with the full domain, but we do not want to redirect all the traffic to the full domain because there's no valid cert yet.

**Why?** Code structure and performance: The application controller has now a big list of filters, that are dealing with redirection:

- redirect_to_marketplace_ident
- force_ssl
- redirect_deleted_marketplace
- redirect_to_marketplace_domain
- redirect_removed_locale
- redirect_locale_param

The structure of the code is pretty difficult to follow. It's hard to get a clear picture what redirects are happening and when. In addition, all this code is untested, even though it's quite critical code for our app.

In addition, multiple redirects may affect to the persieved performance. Let's take an example:

There's a marketplace with ident "boardmarket" and domain "www.boardmarket.com". A user accessed that marketplace with URL "http://www.boardmarket.sharetribe.com". How many redirects will happen (at least)?

1. Redirect http -> https
2. Redirect the ident with www to ident without www (www.boardmarket.sharetribe.com -> boardmarket.sharetribe.com)
3. Redirect to the full marketplace domain boardmarket.sharetribe.com -> www.boardmarket.com
4. And maybe some other redirects because of the selected locales etc.

So my idea was to extract the code from ApplicationController to MarketplaceRedirectUtils. It has a single method `needs_redirect` which takes all the required information to decide where the user should be redirect. That method will be fully tested. In addition, the idea is that the function returns the final redirection target in the current before filter chain and thus eliminates number of redirects happening.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -244,8 +244,9 @@ class ApplicationController < ActionController::Base
 
     host = request.host
     domain = @current_community.domain
+    redirect_to_domain = @current_community.redirect_to_domain
 
-    redirect_opts = request_hash.slice(:host, :protocol, :fullpath).merge(community_domain: domain, domain_ready: domain.present?)
+    redirect_opts = request_hash.merge(community_domain: domain, redirect_to_domain: redirect_to_domain)
 
     MarketplaceRedirectUtils.needs_redirect(redirect_opts) { |redirect_url, redirect_status|
       redirect_to(redirect_url, status: redirect_status)
@@ -256,7 +257,8 @@ class ApplicationController < ActionController::Base
     @request_hash ||= {
       host: request.host,
       protocol: request.protocol,
-      fullpath: request.fullpath
+      fullpath: request.fullpath,
+      port_string: request.port_string
     }
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -245,13 +245,19 @@ class ApplicationController < ActionController::Base
     host = request.host
     domain = @current_community.domain
 
-    if needs_redirect?(host, domain)
-      redirect_to "#{request.protocol}#{domain}#{request.fullpath}", status: :moved_permanently
-    end
+    redirect_opts = request_hash.slice(:host, :protocol, :fullpath).merge(community_domain: domain, domain_ready: domain.present?)
+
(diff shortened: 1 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -244,14 +244,22 @@ def redirect_to_marketplace_domain
...
     domain = @current_community.domain
+    redirect_to_domain = @current_community.redirect_to_domain
 
-    if needs_redirect?(host, domain)
-      redirect_to "#{request.protocol}#{domain}#{request.fullpath}", status: :moved_permanently
-    end
+    redirect_opts = request_hash.merge(community_domain: domain, redirect_to_domain: redirect_to_domain)
+
+    MarketplaceRedirectUtils.needs_redirect(redirect_opts) { |redirect_url, redirect_status|
+      redirect_to(redirect_url, status: redirect_status)
```
gnomet: I assume calling redirect_to with nil parameter just doesn nothing and causes no issues?

gnomet:
So is the plan to get rid of even more separate filters and merge the functionality to now-created utils?
Sounds good to me, and despite the answer, this part looks like a good step with nice tests, so good to go imo! :+1:

app/controllers/application_controller.rb
rap1ds: The block doesn't get called if there's no need to redirect

rap1ds:
Yep, that's the plan. I started with this one since it's needed for the domain file validation. I'm planning to add a few more if they are easy to add. If not, I'll adjust the plan.

app/controllers/application_controller.rb
gnomet: Oh yea, didn't look carefully enough.
:+1:

gnomet:
Sounds good, the plan and going in small steps!

PR #1342, opened 2015-07-29
Add invitation patch limit

Gnito:


app/controllers/invitations_controller.rb
```diff
@@ -9,12 +9,17 @@ class InvitationsController < ApplicationController
...
     @selected_tribe_navi_tab = "members"
     @invitation = Invitation.new
+    render locals: { invitation_limit: Invitation.invitation_limit }
   end
 
   def create
     invitation_emails = params[:invitation][:email].split(",")
-    sending_problems = nil
 
+    if invitation_emails.size >= Invitation.invitation_limit
```
Gnito: This will be changed in the next PR from patch limit to daily limit.

config/locales/en.yml
```diff
@@ -1133,9 +1133,13 @@ en:
       this_is_private_community: "You need to sign up before you can view the content."
     invitation_form:
       email: "Email address(es)"
+      email_with_limit: "Email addresses (max. %{invitation_limit})"
       message: "A personal message"
       send_invitation: "Send invitation"
       add_as_many_emails_as_you_want: "Add the email addresses of the people you are inviting to the field below. If you add multiple email addresses, separate them with comma."
+      add_email_addresses_description: "Add the email addresses of the people you are inviting to the field below. To add multiple email addresses, separate them with comma."
+      add_lots_of_email_addresses: "If you plan to send lots of invitations, you should probably use an emailing tools. Check out %{this_article_link} to find out resources about suitable email services."
```
rap1ds: - Should we remove the "maybe". I think the case is pretty clear that if they are sending a lot of email, then the emailing tool should be used.
- "...an emailing tool_", there should not be "s"

config/locales/en.yml
```diff
@@ -1133,9 +1133,13 @@ en:
       this_is_private_community: "You need to sign up before you can view the content."
     invitation_form:
       email: "Email address(es)"
+      email_with_limit: "Email addresses (max. %{invitation_limit})"
       message: "A personal message"
       send_invitation: "Send invitation"
       add_as_many_emails_as_you_want: "Add the email addresses of the people you are inviting to the field below. If you add multiple email addresses, separate them with comma."
```
rap1ds: I guess this is not used anymore? If so, then it should be removed.

rap1ds:
A couple small comments about copy texts. Otherwise looks good to me :+1: 

Didn't test the code, should I?

config/locales/en.yml
Gnito: removed extra 's' & 'probably'

config/locales/en.yml
Gnito: True. However, I kind of left this because I would prefer to just rename the key. (The texts are almost the same - and therefore it would have been ok to just rephrase the old translation.)

PR #1344, opened 2015-07-30
Refactor marketplace redirects

rap1ds:
Add a `MarketplaceRedirectionUtils`, which takes a care of the redirections needed to perform.

Following redirections are done:

1. If community is not found and there are no communities in the whole platform, redirect to new community form (this is mostly for new open source installations and local development)
2. If community is not found, redirect to not found page, or to url from configurations (sharetribe.com)
3. If community is deleted, redirect to not found page, or to url from configurations (sharetribe.com)
4. If community is accessed with ident, but the community has domain which is ready, redirect to domain
5. If community is accessed with subdomain but it includes "www.", redirect to subdomain without "www"
6. If http is used and `always_use_ssl` is set to true, redirect to https (except if it's robots.txt, or domain validation file, or if the request comes from proxy)

One goal of this Pull Request is to reduce the amount of redirects by doing all the necessary redirects at once. Here's a good SO answer how to debug redirects with Chrome: http://stackoverflow.com/questions/10987453/how-to-use-chromes-network-debugger-with-redirects

For example, if I try to access "Larun tori" marketplace (a marketplace previously run by Sharetribe, but currently deleted) we can see quite a many redirects happening:

screen shot 2015-07-31 at 10 01 45

1. Redirect from http://www.lauttasaari.sharetribe.com to http://lauttasaari.sharetribe.com
2. Redirect from http://lauttasaari.sharetribe.com to https://lauttasaari.sharetribe.com
3. Redirect from https://lauttasaari.sharetribe.com to www.sharetribe.com (because the marketplace is deleted)

And as you can see from the waterfall chart, there redirects do take some time.

### DIFF
--- a/app/controllers/braintree_webhooks_controller.rb
+++ b/app/controllers/braintree_webhooks_controller.rb
@@ -1,7 +1,7 @@
 class BraintreeWebhooksController < ApplicationController
 
   skip_before_filter :verify_authenticity_token
-  skip_filter :redirect_to_marketplace_ident, :redirect_to_marketplace_domain, :check_email_confirmation
+  skip_filter :check_email_confirmation
 
   before_filter do
     unless @current_community.braintree_in_use?
--- a/app/controllers/braintree_webhooks_controller.rb
+++ b/app/controllers/braintree_webhooks_controller.rb
@@ -1,9 +1,7 @@
 class BraintreeWebhooksController < ApplicationController
 
   skip_before_filter :verify_authenticity_token
-  skip_filter :fetch_community, :redirect_to_marketplace_ident, :redirect_to_marketplace_domain, :check_email_confirmation
-
-  before_filter :fetch_community_by_params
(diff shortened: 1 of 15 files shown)


app/controllers/braintree_webhooks_controller.rb
```diff
@@ -1,9 +1,7 @@
 class BraintreeWebhooksController < ApplicationController
 
   skip_before_filter :verify_authenticity_token
-  skip_filter :fetch_community, :redirect_to_marketplace_ident, :redirect_to_marketplace_domain, :check_email_confirmation
-
-  before_filter :fetch_community_by_params
+  skip_filter :redirect_to_marketplace_ident, :redirect_to_marketplace_domain, :check_email_confirmation
```
rap1ds: Remove :redirect_to_marketplace_ident

Also, search for other removed filters

PR #1345, opened 2015-07-30
change from patch limit to daily limit

Gnito:


config/locales/en.yml
```diff
@@ -1133,7 +1133,7 @@ en:
       this_is_private_community: "You need to sign up before you can view the content."
     invitation_form:
       email: "Email address(es)"
-      email_with_limit: "Email addresses (max. %{invitation_limit})"
+      email_with_limit: "Email addresses (max. %{invitation_limit} per day)"
```
rap1ds: Should we say "per day per sender", because that's what the real limit is?

Gnito: True, but do you think that user's expects that someone else's invitations affect to their daily amount?

Gnito: After discussing with Thomas & Antti, we decided to remove all explicit info about daily limits since only few admins can see it (and they see also this linked KB article)

PR #1351, opened 2015-08-04
Don't delete tx from 13116 error (payment ongoing)

ovan:
When PayPal responds to DoExpressCheckoutPayment call with 13116 it means that we already called this API method once and the payment processing is in progress. We end up calling the API twice in case the token retry logic is in race condition with the synchronous payment process. This is rare but possible and has happened in production at least once. This change prevents the worker process from incorrectly terminating the transaction.


rap1ds:
:+1:

PR #1352, opened 2015-08-04
Identify PayPal req/resp pairs in log with uuid

ovan:


app/services/paypal_service/logger.rb
```diff
@@ -14,23 +14,29 @@ def error(msg, type = :other)
     end
 
     def log_request_input(request, input)
+      request_id = SecureRandom.uuid
```
rap1ds: No need to change, but this is a example case where `tap` could be used.

rap1ds:
:+1:

PR #1354, opened 2015-08-05
Add community_sender_emails table and migrate data

rap1ds:
- [x] Add community_sender_emails table
- [x] Copy data from community table to the new community_sender_emails table

The community_sender_emails table data is not used, yet.


ovan:
Looks good but let's rename the table to marketplace_sender_emails

PR #1355, opened 2015-08-05
Upgrade AWS SDK to version 2

ovan:
Add dependency to AWS Ruby SDK version 2. We have old code that is built with v1 so we keep the v1 around. AWS SDK is designed to support having both version in parallel and has different root-namespace for classes and modules in v1 (AWS) and v2 (Aws).


rap1ds:
Isn't it so that minor version updates will never contain breaking changes to public API? I guess so. In that case this is ok to go :+1:

PR #1356, opened 2015-08-05
Use data from marketplace_sender_emails table

rap1ds:
The previous PR #1354 added a new marketplace_sender_emails table and moved data there. This pull requests changes the code so the the new table and the data in it is used.

Migrate after #1354

### DIFF
--- a/app/services/email_service/api/addresses.rb
+++ b/app/services/email_service/api/addresses.rb
@@ -14,7 +14,7 @@ module EmailService::API
         to_smtp_format(name: address[:name], email: address[:email])
       }.or_else(@default_sender)
 
-      Result::Success.new(sender)
+      Result::Success.new(formatted: sender)
     end
 
     # TODO get_user_defined
@@ -30,7 +30,7 @@ module EmailService::API
     private
 
     def with_smtp_format(address)
-      address.merge(smtp_format: to_smtp_format(name: address[:name], email: address[:email]))
+      address.merge(formatted: to_smtp_format(name: address[:name], email: address[:email]))
     end
 
     def to_smtp_format(name: nil, email:)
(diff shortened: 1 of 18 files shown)


app/utils/mail_utils.rb
```diff
@@ -74,10 +74,11 @@ def set_community(new_community_id, community_locales, &block)
...
 
   def community_specific_sender(community)
-    if community && community.custom_email_from_address
-      community.custom_email_from_address
-    else
-      APP_CONFIG.sharetribe_mail_from_address
-    end
+    Maybe(community).id.map { |cid|
+      # TODO Use API
+      MarketplaceSenderEmail.where(community_id: cid).first
```
ovan: Why leave a TODO here instead of implementing this part of the API and store straight away?

app/services/email_service/email_service_injector.rb
```diff
@@ -0,0 +1,7 @@
+module EmailService::EmailServiceInjector
+  def addresses_api
+    @addresses = EmailService::API::Addresses.new(
```
rap1ds: ||=

app/services/email_service/api/addresses.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def get_sender(community_id:)
+      sender = Maybe(community_id).map {
+        AddressStore.get(community_id: community_id)
+      }.map { |address|
+        to_smtp_format(name: address[:name], email: address[:email])
+      }.or_else(@default_sender)
+
+      Result::Success.new(sender)
+    end
```
ovan: My suggestion is to change this to be a hash with single key 'formatted'. This way it's ~the same format (with some keys missing) as what create and get_user_defined would use.

PR #1360, opened 2015-08-06
Sender address ui

rap1ds:


app/assets/javascripts/admin/emails.js
```diff
@@ -0,0 +1,58 @@
...
+    var validEmailStream = emailStream.filter(function() { return $form.valid(); });
+    validEmailStream.take(1).onValue(function() {
+      $previewContainer.show();
+    });
+
+    nameEmailStream.onValue(function(values) {
+      $preview.text(formatSender(values));
+    });
+
+    nameEmailStream.log("Name and email: ");
```
ovan: Leftover from debugging?

app/services/email_service/api/addresses.rb
```diff
@@ -29,17 +36,27 @@ def create(community_id:, opts:)
 
     private
 
-    def with_smtp_format(address)
-      address.merge(formatted: to_smtp_format(name: address[:name], email: address[:email]))
+    def with_formatted(address)
+      address.merge(
+        formatted: to_format(name: address[:name], email: address[:email], quotes: false),
```
ovan: Suggestion to change names to indicate the intended use case better:
formatted -> display_format
smtp_formatted -> smtp_format

app/services/email_service/store/address.rb
```diff
@@ -17,13 +17,28 @@ def get(community_id:)
...
     from_model(MarketplaceSenderEmail.create!(address))
   end
 
   def find_model(community_id:)
-    MarketplaceSenderEmail.where(community_id: community_id).first
+    find_models(community_id: community_id).first
+  end
+
+  def find_models(community_id:)
+    MarketplaceSenderEmail.where(community_id: community_id)
```
ovan: My feeling about readability is that find_models and find_model add very little in terms of code reuse but by adding one level of indirection make it harder to read this code, i.e. to quickly see what exactly is queried and by which attributes.

app/controllers/admin/communities_controller.rb
```diff
@@ -30,21 +31,40 @@ def edit_welcome_email
...
+             sender_address: sender_address,
+             user_defined_address: user_defined_address,
+             post_sender_address_url: create_sender_address_admin_community_path,
+           }
+  end
+
+  def create_sender_address
+    # TODO validate email
+
+    EmailService::API::Api.addresses.create(community_id: @current_community.id, opts: {name: params[:name], email: params[:email]})
```
ovan: opts is a quite generic and when I see it I always assume something like flags to change method behavior. How about:
```ruby
EmailService::API::Api.addresses.create(community_id: @current_community.id, address: {name: params[:name], email: params[:email]})
```

PR #1361, opened 2015-08-06
Fix minor UI bug in admin Paypal settings page

vsaarinen:
From:
screenshot 2015-08-06 12 16 04
To: 
screenshot 2015-08-06 12 16 09


ovan:
+1

PR #1364, opened 2015-08-07
SES Client wrapper

ovan:


app/services/email_service/ses/client.rb
```diff
@@ -0,0 +1,52 @@
...
+  class Client
+
+    def initialize(region:, stubs: nil)
+      if region.blank?
+        raise ArgumentError.new("Missing mandatory value for region parameter.")
+      end
+
+      # The SDK v2 automatically gets the AWS_ACCESS_KEY and
+      # AWS_SECRET_ACCESS_KEY from the environment. We only need to
+      # supply region here.
```
rap1ds: I guess we could and should use the APP_CONFIG, so that local development settings would be also available

spec/services/email_service/ses/client_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'spec_helper'
+
+describe EmailService::SES::Client do
+  context "ctor" do
```
rap1ds: I'm not sure if everyone knows what "ctor" means...

rap1ds: Could use also `describe` here

rap1ds:
Looks good! :+1: I think the APP_CONFIG thing should be changes

PR #1365, opened 2015-08-07
Synchronize verification state between SES and local

ovan:


app/services/email_service/ses/synchronize.rb
```diff
@@ -0,0 +1,74 @@
...
+        update_statuses(build_sync_updates(addresses, vaddrs))
+
+        offset += BATCH_SIZE
+        addresses = AddressStore.load_all(limit: BATCH_SIZE, offset: offset)
+      end
+    }
+  end
+
+  def build_sync_updates(addresses, verified_addresses)
+    vaddrs = verified_addresses.to_set
```
rap1ds: I don't know how time consuming operation `to_set` is, but since `build_sync_updates` may be in theory called multiple times, you could also convert `to_set` outside this method and pass the set as parameter.

app/services/email_service/ses/synchronize.rb
```diff
@@ -0,0 +1,74 @@
...
+    updates = { verified: [], expired: [], touch: [], none: [] }
+
+    addresses.each do |a|
```
rap1ds: reduce?

app/services/email_service/ses/synchronize.rb
ovan: True, I'll change it (or leave it here still but already pass a set from batch syncing since calling to_set to a set should be O(1))

app/services/email_service/ses/synchronize.rb
ovan: Yeah, that's cleaner put contained mutability chosen on purpose as an easy optimization.

spec/services/email_service/ses/synchronize_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+
+        stubs = {list_verified_email_addresses: {verified_email_addresses: verified_addresses}}
+        ses_client = EmailService::SES::Client.new(config: {region: "fake-region", access_key_id: "access_key", secret_access_key: "secret_access_key"},
+                                                   stubs: {list_verified_email_addresses: {verified_email_addresses: verified_addresses}})
+
+
+        Timecop.travel(now + 5.seconds) do
+          Synchronize.run_synchronization!(ses_client)
+
+          verified_addresses = MarketplaceSenderEmail.where(verification_status: "verified").pluck(:id)
```
rap1ds: I feel that we instead of accessing the Model object, we could use some higher abstraction, like Store or even API.

PR #1369, opened 2015-08-10
Fix other party author

Gnito:
Admin sees wrong status message in transaction

fix for:
- admin sees wrong person when payment is waiting for author to accept payment
- 'paid' status message was changed to 'payment authorized'

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -597,6 +597,7 @@ en:
       canceled_request: "canceled the request"
       canceled_offer: "canceled the offer"
       paid: "paid %{sum}"
+      payment_preauthorized: "Payment authorized: %{sum}"
     new:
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -597,6 +597,7 @@ en:
       canceled_request: "canceled the request"
       canceled_offer: "canceled the offer"
       paid: "paid %{sum}"
+      payment_preauthorized: "Payment authorized: %{sum}"
```
rap1ds: This is not used.

rap1ds:
One comment, otherwise ok :+1:

config/locales/en.yml
Gnito: Finally, saved.

PR #1371, opened 2015-08-10
Integrate SES calls with addresses API

ovan:


app/controllers/admin/communities_controller.rb
```diff
@@ -35,8 +35,12 @@ def edit_welcome_email
     sender_address = EmailService::API::Api.addresses.get_sender(community_id: @current_community.id).data
     user_defined_address = EmailService::API::Api.addresses.get_user_defined(community_id: @current_community.id).data
 
-    Maybe(user_defined_address)[:verification_status].reject { |status| status == :verified }.each {
-      EmailService::API::Api.addresses.enque_status_sync
+    Maybe(user_defined_address)
+      .reject { |address| address[:verification_status] == :verified }
```
rap1ds: Should also reject `expired`.

app/services/email_service/api/addresses.rb
```diff
@@ -35,15 +37,38 @@ def get_user_defined(community_id:)
     end
 
     def create(community_id:, address:)
-      Result::Success.new(
-        with_formats(
-          AddressStore.create(
-          community_id: community_id,
-          address: {verification_status: :none}.merge(address))))
+      create_in_status = @ses_client ? :none : :verified
```
rap1ds: `get_sender` needs to be changed so that it allows status `:none` if ses client is `nil`

ovan: Why? Without ses client we have everything created in status :verified?

rap1ds: Sorry. Never mind.

app/services/email_service/jobs/request_email_verification.rb
```diff
@@ -0,0 +1,25 @@
+module EmailService::Jobs
+  class RequestEmailVerification < Struct.new(:community_id, :id)
+
+    AddressStore = EmailService::Store::Address
+
+    include DelayedAirbrakeNotification
+
+    def perform
+      if ses_client
```
rap1ds: At this point, can we just assume that we have the `ses_client` configured?

ovan: Yeah. It's an error to schedule this job without ses_client being configured. It's a different process though so it might not be configured here but in that case it would be better to get an error raised to detect it than let if fail silently. I'll remove the check.

PR #1372, opened 2015-08-11
Marketplace Plan API

rap1ds:

PR #1376, opened 2015-08-11
Hide emails from free

rap1ds:
Merge after #1372


app/controllers/admin/communities_controller.rb
```diff
@@ -307,4 +310,25 @@ def can_delete_marketplace?(community_id)
     PlanService::API::Api.plans.get_current(community_id: community_id).data[:plan_level] == PlanService::Levels::FREE
   end
 
+  def can_set_sender_address(plan)
+    Maybe(plan).map { |p|
+      p[:expired] == false && p[:plan_level] > 0
+    }.or_else(false)
+  end
```
ovan: Reading through this it looks like checking the plan level is a little cumbersome (have to make two comparisons yourself) and also this doesn't communicate to the person reading the code very well what the idea here really is. Is this a problem with the API? Should it do something other than just return a plan for given community so you could directly ask things like "Is this community pro or above?"

PR #1377, opened 2015-08-12
Addresses API validates the email address

rap1ds:
Addresses API validates the email address and returns Result::Error if the address is invalid. Proper error message is included. The error message is shown to user as a error flash notification.

### DIFF
--- a/app/controllers/admin/communities_controller.rb
+++ b/app/controllers/admin/communities_controller.rb
@@ -54,17 +54,25 @@ class Admin::CommunitiesController < ApplicationController
   end
 
   def create_sender_address
-    # TODO validate email
-
-    EmailService::API::Api.addresses.create(
+    res = EmailService::API::Api.addresses.create(
       community_id: @current_community.id,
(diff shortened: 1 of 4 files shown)


app/controllers/admin/communities_controller.rb
```diff
@@ -54,17 +54,25 @@ def edit_welcome_email
...
-    flash[:notice] = t("admin.communities.outgoing_email.successfully_saved")
-    redirect_to action: :edit_welcome_email
+    if res.success
+      flash[:notice] =
+        t("admin.communities.outgoing_email.successfully_saved")
+
+      redirect_to action: :edit_welcome_email
+    else
+      flash[:error] =
+        "An error occured while saving address: #{res.error_msg}" # Shouldn't come here.
```
ovan: My feeling is that we should just always use translatable strings even if it's unlikely that these things happen. That way we know that we at least can translate these if needed and the extra work using localized str is nothing.

PR #1378, opened 2015-08-12
SES integration fine tuning

rap1ds:
- Show notification about sender adderess, i.e. "The verification email was sent from Amazon Web Services"
- Remove some extra whitespaces
- Fix the verification sent "5 seconds ago". We need to use JS to set the correct value
- Add knowledge base article link


ovan:
:+1:

PR #1379, opened 2015-08-12
Setup bounces&complaints SNS for SES identitiies

ovan:


app/services/email_service/ses/client.rb
```diff
@@ -38,17 +49,52 @@ def verify_address(email: )
...
         end
-      rescue StandardError => e
-        Result::Error.new(e)
       end
     end
 
+    def set_notification_topic(email:, type:)
+      nt = DataTypes::NotificationTopic.call({email: email, type: type})
+      method_params = {identity: nt[:email],
+                       notification_type: nt[:type].to_s.capitalize,
```
rap1ds: I guess you could also use Entity transformer here?

app/services/email_service/ses/client.rb
```diff
@@ -38,17 +49,52 @@ def verify_address(email: )
...
+                       sns_topic: @sns_topic}
```
rap1ds: What happens, if SNS topic is not configured and thus `@sns_topic` is `nil`? Should we check and make sure that it is configured, before making the API call?

ovan: initialize method validates that it exists so the only way to mess this up is to mess with instance variables after construction. I might still add a validation because that's a valid call (nil sns topic) and it clears the sns topics from the email.

PR #1381, opened 2015-08-13
Disable email forwarding for SES sender addresses

ovan:
To prevent marketplace admins from unnecessarily getting all the bounces
and complaints to their sender address we instead redirect them to a SNS
topic that we handle on their behalf.

### DIFF
--- a/app/services/email_service/ses/client.rb
+++ b/app/services/email_service/ses/client.rb
@@ -88,6 +88,28 @@ module EmailService::SES
       end
     end
 
+    def disable_email_forwarding(email:)
+      if email.blank?
+        raise ArgumentError.new("Missing mandatory value for email parameter.")
+      end
+
+      method_params = {identity: email, forwarding_enabled: false}
+
+      log_request_response(:set_identity_feedback_forwarding_enabled, method_params) do
+        begin
+          response = @ses.set_identity_feedback_forwarding_enabled(method_params)
+          if response.successful?
+            Result::Success.new()
+          else
+            Result::Error.new(response.error)
+          end
+        rescue StandardError => e
+          Result::Error.new(e)
+        end
+      end
+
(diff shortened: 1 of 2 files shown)


app/services/email_service/ses/client.rb
```diff
@@ -88,6 +88,28 @@ def set_notification_topic(email:, type:)
...
+        begin
+          response = @ses.set_identity_feedback_forwarding_enabled(method_params)
+          if response.successful?
+            Result::Success.new()
+          else
+            Result::Error.new(response.error)
+          end
+        rescue StandardError => e
+          Result::Error.new(e)
+        end
```
rap1ds: There's quite a lot of repetition. This is the 4th time this pattern is present in this file: log, do request, if success, if error, catch exception.

I don't know, maybe we could add a helper or something. On the other it might be that adding a helper would actually make the code harder to read and understand.

ovan: I thought about this. Because we handle response differently with list_verified_addresses the only repetition accross all calls is begin ... rescue StandardError => e -thingy. That could go to a helper that takes a block but then with logging block we would have two nested blocks that each "advice" the innermost block:

```ruby
      log_request_response(:set_identity_feedback_forwarding_enabled, method_params) do
        run_request do
          response = @ses.set_identity_feedback_forwarding_enabled(method_params)
          if response.successful?
            Result::Success.new()
          else
            Result::Error.new(response.error)
          end
      end
  end
end
```

This is more brief and to the point but also to understand the code you have to read through two block wrappers and understand the order in which things expand. I feel like single block wrapper that does one coherent thing is the amount of nesting I can manage easily in my head. That single block wrapper here is logging. So in this case I want to opt for little copy&paste to keep this readable instead of going super DRY.

PR #1383, opened 2015-08-13
Add guards for payment not being in correct state

sktoiva:
This change guards against users navigating back in accept transaction
flow and then trying to accept again.


ovan:
Looks good. Can you also add a check to accept_or_reject action so that double post doesn't result in displaying error message? So check that if we're already past the preauthorized state then just redirect forward with no flash messages.

PR #1385, opened 2015-08-13
Yahoo is not allowed

rap1ds:


app/services/email_service/api/addresses.rb
```diff
@@ -99,15 +103,26 @@ def quote(str, quotes)
...
+        email_regexp.match(email).present? ? Result::Success.new() : Result::Error.new("invalid email format")
+      else
+        Result::Error.new()
+      end
+    end
+
+    def valid_email_domain?(email)
+      if @ses_client
+        fulldomain = email.split("@").second
+        main_domain = fulldomain.split(".")[-2]
```
ovan: What about top levels like co.uk. Trusting the -2 doesn't work there. Now we only want to test for yahoo so this is not a problem. It just looks like this code tries to be a bit more generic than what it is. How about just testing if the address ends with yahoo\..+ ?

rap1ds: Oh, right. I didn't think about co.uk etc...

PR #1397, opened 2015-08-17
README content tweaks

vsaarinen:
Suggest using a dedicated HTTP server in production.


rap1ds:
Looks good to me :+1:

PR #1398, opened 2015-08-17
Add Hungarian to the list of unsupported languages

rap1ds:
- The unsupported list contains languages that exist in the application but
are not translated completely

### DIFF
--- a/config/available_locales.rb
+++ b/config/available_locales.rb
@@ -5,8 +5,9 @@ module Sharetribe
   #
   # language format: ISO 639-1, two letters, lowercase
   # region format: ISO 3166, two letters, uppercase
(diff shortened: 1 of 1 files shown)


config/available_locales.rb
```diff
@@ -5,8 +5,9 @@ module Sharetribe
   #
   # language format: ISO 639-1, two letters, lowercase
   # region format: ISO 3166, two letters, uppercase
-  # fallbacks: should not include US English, which is a last default fallback for each language
+  # fallbacks: should not include US English, which is a last default fallback for each langua
```
sktoiva: Language, and actually, while you are at it, it should probably *the* last default fallback - not *a*.

PR #1403, opened 2015-08-19
Issue template

rap1ds:
**Why?**

In order to provide useful answers to Issue Reports, we need to know the basic information about the environment and the steps that were done before the error occurred. The sheer stack trace is almost never enough information in order to give any meaningful answer to the issue, other than "please tell me more?".

The goal is to save time from both parties, the reporter and the reviewer by clearly defining, what do we expect from the issue report.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,11 +1,10 @@
 # Contributing to Sharetribe
 
-
 First of all, thank you so much! We greatly appreciate all contributions. However, before you spend time working on Sharetribe, please read over our guidelines:
 
-## Issue reports
+## Reporting issues
 
-Please use [ISSUE_TEMPLATE.md](https://raw.githubusercontent.com/sharetribe/sharetribe/master/ISSUE_TEMPLATE.md) for the issue report.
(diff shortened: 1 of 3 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
...
+
+What's your environment? MacOSX, Heroku, or some other?
+
+##### Server mode:
+
+Are you running development more or production mode?
+
+##### Description:
+
+In our own words, briefly describe the issue. Please add enough details, yet be consice.
```
mporkola: it's "concise"!

CONTRIBUTING.md
```diff
@@ -3,6 +3,10 @@
 
 First of all, thank you so much! We greatly appreciate all contributions. However, before you spend time working on Sharetribe, please read over our guidelines:
 
+## Issue reports
+
+Please use [ISSUE_TEMPLATE.md](https://raw.githubusercontent.com/sharetribe/sharetribe/master/ISSUE_TEMPLATE.md) for the issue report.
+
```
mporkola: I might also add "Check if the issue is already reported", even though it may seem obvious. From Github's example:

> Check existing issues for your issue. Duplicating an issue is slower for both parties so search through open and closed issues to see if what you’re running into has been addressed already.

PR #1404, opened 2015-08-19
Initial Search API integration

rap1ds:
The purpose of this PR is to test the Search API integration. It's not meant to be merged, at least not right now.

To test it, open console and hit:

```
ListingService::API::Api.listings.search(community_id: 1, search: {keywords: "lorem"})
```

### DIFF
--- a/app/services/listing_index_service/api/listings.rb
+++ b/app/services/listing_index_service/api/listings.rb
@@ -28,6 +28,7 @@ module ListingIndexService::API
           listings: res[:listings].map { |search_res|
             search_res.merge(url: "#{search_res[:id]}-#{search_res[:title].to_url}")}))
       }.or_else {
+        raise search_result.data if Rails.env.development?
         log_error(search_result, community_id)
         search_result
       }
--- a/app/services/listing_index_service/api/listings.rb
+++ b/app/services/listing_index_service/api/listings.rb
@@ -3,7 +3,8 @@ module ListingIndexService::API
   RELATED_RESOURCES = [:listing_images, :author, :num_of_reviews, :location].to_set
 
   # TODO Maybe conf+injector?
-  ENGINE = :sphinx
+  # ENGINE = :sphinx
+  ENGINE = :zappy
 
   ListingIndexResult = ListingIndexService::DataTypes::ListingIndexResult
 
@@ -42,6 +43,8 @@ module ListingIndexService::API
       case ENGINE
       when :sphinx
         ListingIndexService::Search::SphinxAdapter.new
+      when :zappy
+        ListingIndexService::Search::ZappyAdapter.new
       else
         raise NotImplementedError.new("Adapter for search engine #{ENGINE} not implemented")
       end
--- a/app/services/listing_index_service/api/listings.rb
+++ b/app/services/listing_index_service/api/listings.rb
@@ -2,10 +2,6 @@ module ListingIndexService::API
 
   RELATED_RESOURCES = [:listing_images, :author, :num_of_reviews, :location].to_set
 
-  # TODO Maybe conf+injector?
-  # ENGINE = :sphinx
-  ENGINE = :zappy
-
   ListingIndexResult = ListingIndexService::DataTypes::ListingIndexResult
 
   class Listings
@@ -14,12 +10,12 @@ module ListingIndexService::API
       @logger_target = logger_target
     end
 
-    def search(community_id:, search:, includes: [])
+    def search(community_id:, search:, includes: [], engine: :sphinx)
       unless includes.to_set <= RELATED_RESOURCES
         return Result::Error.new("Unknown included resources: #{(includes.to_set - RELATED_RESOURCES).to_a}")
       end
 
-      search_result = search_engine.search(
+      search_result = search_engine(engine).search(
         community_id: community_id,
         search: ListingIndexService::DataTypes.create_search_params(search),
         includes: includes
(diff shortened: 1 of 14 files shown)


rap1ds:
@ovan You have committed one new commit here. Is it useful? Can we close this branch?

app/services/listing_index_service/search/zappy_adapter.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    def format_params(original)
+      defaults = {
+        include_closed: false
+      }
+      params = {}
+      params[:'page[number]'] = original[:page] if original[:page]
+      params[:'page[size]'] = original[:per_page] if original[:per_page]
+      params[:keywords] = original[:keywords] if original[:keywords]
+      params[:include_closed] = original[:include_closed] if original[:include_closed]
```
rap1ds: This pattern is used somewhere in our codebase, however, I find it a bit complex. It has a lot of mutation going on and a lot of if-branches.

How about using `compact?`

```ruby
{
  :include_closed => false,
  :'page[number]' => original[:page],
  :'page[size]' => original[:per_page],
  ...
}.compact
```

(of course, this doesn't work if `nil` actually means something)

app/services/listing_index_service/search/zappy_adapter.rb
```diff
@@ -0,0 +1,84 @@
...
+      defaults.merge(params)
+    end
+
+    def listings_from_ids(id_obs, includes)
+      # use pluck for much faster query after updating to Rails >4.1.6
```
rap1ds: TODO: Rebase and use `pluck` :)

app/services/listing_index_service/search/zappy_adapter.rb
```diff
@@ -0,0 +1,84 @@
...
+      # http://collectiveidea.com/blog/archives/2015/03/05/optimizing-rails-for-memory-usage-part-3-pluck-and-database-laziness/
+      # https://github.com/rails/rails/issues/17049
+
+      ids = id_obs.map { |r| r['id'] }
+
+      Listing
+        .where(id: ids) # use find_each for more efficient batch processing after updating to Rails 4.1
```
rap1ds: TODO: Rebase and use `find_each`

app/services/listing_index_service/search/zappy_adapter.rb
```diff
@@ -0,0 +1,84 @@
...
+      if needs_db_query?(search) && needs_search?(search)
+        return Result::Error.new(ArgumentError.new("Both DB query and search engine would be needed to fulfill the search"))
+      end
+
+      if needs_search?(search)
+        # TODO: is out-of-bounds check necessary here?
+        begin
+          res = @conn.get do |req|
+            req.url("/api/v1/marketplace/#{community_id}/listings", search_params)
+            req.headers['Authorization'] = 'apikey key=asdfasdf'
```
rap1ds: The API key is probably only for testing?

app/services/listing_index_service/search/zappy_adapter.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    INCLUDE_MAP = {
+      listing_images: :listing_images,
+      author: :author,
+      num_of_reviews: {author: :received_testimonials},
+      location: :location
+    }
+
+    def initialize
+      @conn = Faraday.new(url: "http://127.0.0.1:8080") do |c|
```
rap1ds: This needs to go to environment variable

app/services/listing_index_service/search/zappy_adapter.rb
```diff
@@ -0,0 +1,84 @@
...
+         c.request  :url_encoded             # form-encode POST params
+         c.response :logger                  # log requests to STDOUT
+         c.response :json, :content_type => /\bjson$/
+         c.adapter  Faraday.default_adapter  # make requests with Net::HTTP
+         c.use Faraday::Response::RaiseError if Rails.env.development?
```
rap1ds: Hmm... I'm just wondering that this would maybe be a bit cleaner if you'd pass `raise_errors` as on argument to the `initialize` method instead of being dependant on the outside world state.

app/services/listing_index_service/search/commons.rb
```diff
@@ -0,0 +1,85 @@
+module ListingIndexService::Search::Commons
```
rap1ds: I guess this module could have a more descriptive name. All the methods seem to have something to do about converting models to hashes, so could the module be `Converters` or something like that?

app/services/listing_index_service/api/listings.rb
```diff
@@ -31,6 +28,7 @@ def search(community_id:, search:, includes: [])
           listings: res[:listings].map { |search_res|
             search_res.merge(url: "#{search_res[:id]}-#{search_res[:title].to_url}")}))
       }.or_else {
+        raise search_result.data if Rails.env.development?
```
rap1ds: Same comment here as below: This probably would be cleaner if you'd pass `raise_errors` as a variable and not depend on the outside world state.

app/services/listing_index_service/search/zappy_adapter.rb
mporkola: Yeah, I was quite certain there's a better way. Thanks!

app/services/listing_index_service/search/zappy_adapter.rb
mporkola: That's true, also API URI and stuff like that. Will fix.

rap1ds:
Using concrete inheritance (i.e. inheriting concrete method implementations, not only an interface) for the `SearchAdapter` is an interesting choice. I think that the fact that SphinxAdapter and ZappyAdapter are using some same functions is a temporary thing. In the ideal world, ZappyAdapter will not touch the database anymore. When this happens, you need to move `needs_db?` and `fetch_from_db` away from the `SearchAdapter` back to `SphinxAdapter`.

That made me thinking, instead of inheriting the methods, should you extract them to a separate module (something like `DatabaseSearchHelper`), and both ZappyAdapter and SphinxAdapter would use methods like `DatabaseSearchHelper.needs_db?( ... )`.

rap1ds:
Looks good!

PR #1406, opened 2015-08-20
Use custom struct instead of Listing object to populate browsing views.

mporkola:
This allows us to get the data from any datasource containing the necessary fields, instead of requiring ActiveRecord access.

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -6,6 +6,12 @@ class HomepageController < ApplicationController
   APP_DEFAULT_VIEW_TYPE = "grid"
   VIEW_TYPES = ["grid", "list", "map"]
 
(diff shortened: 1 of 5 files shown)


app/controllers/homepage_controller.rb
```diff
@@ -105,7 +111,17 @@ def find_listings(params, listings_per_page, filter_params)
     if numeric_search_needed && filter_params[:listing_id].empty?
       Listing.none.paginate(:per_page => listings_per_page, :page => params[:page])
     else
-      Listing.find_with(filter_params, @current_user, @current_community, listings_per_page, params[:page])
+      listings = Listing.find_with(filter_params, @current_user, @current_community, listings_per_page, params[:page]).map { |e|
```
sktoiva: Should this whole block be now abstracted behind some service? Is this going to be the next phase?

mporkola: Yes, that's happening next

mporkola:
This branch in itself causes performance issues (~50 % drop in throughput) without the underlying data source changes, so let's not merge until they're done.

PR #1408, opened 2015-08-20
Fix :mandatory validator performance issue with large hashes

rap1ds:
- Remove to_s


mporkola:
Can't think where this would work differently, looks good to me

PR #1410, opened 2015-08-20
Remove unused privacy

rap1ds:
- [X] Listing contains a field called `privacy`. However, in practise, this field is not in use. We use community information to define whether the community is private or not and that info is used to define whether we should show the listing or not.

- [X] Remove `Listing.visible_to`, since it was using the `privacy` column

The `visible_to` did two things: Ensured that only listings with privacy "public" were returned if user was not logged in (useless) and that listing belonged to current community. When `visible_to` was removed it's important to make sure that the community is still the right one.

- [X] Remove unused `@listing_count`
- [X] Remove unused `@private_listing_count`
- [X] Remove tests that were testing the privacy

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -101,7 +101,13 @@ class ListingsController < ApplicationController
   # Used to show multiple listings in one bubble
   def listing_bubble_multiple
     ids = numbers_str_to_array(params[:ids])
-    @listings = Listing.visible_to(@current_user, @current_community, ids).order("id DESC")
+
+    if @current_user || !@current_community.private?
+      @listings = @current_community.listings.where(listings: {id: ids}).order("listings.created_at DESC")
+    else
+      @listings = []
+    end
+
     if @listings.size > 0
       render :partial => "homepage/listing_bubble_multiple"
     else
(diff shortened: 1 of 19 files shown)


app/controllers/listings_controller.rb
```diff
@@ -190,6 +196,10 @@ def create
       params[:listing].delete("origin_loc_attributes")
     end
 
+    # TODO REMOVE ME!
+    params[:listing][:privacy] = @current_community.private? ? "private" : "public"
```
rap1ds: Remove

app/controllers/listings_controller.rb
```diff
@@ -292,6 +302,10 @@ def update
       end
     end
 
+    # TODO REMOVE ME!
+    params[:listing][:privacy] = @current_community.private? ? "private" : "public"
```
rap1ds: Remove

app/controllers/listings_controller.rb
```diff
@@ -101,7 +101,13 @@ def numbers_str_to_array(str)
   # Used to show multiple listings in one bubble
   def listing_bubble_multiple
     ids = numbers_str_to_array(params[:ids])
-    @listings = Listing.visible_to(@current_user, @current_community, ids).order("id DESC")
+
+    if @current_user || !@current_community.private?
```
sktoiva: This could be extracted to own function. Now `@listings` is initialized in two places. Something like:

```ruby
@listings = listings_for_community(@current_community)
```

PR #1411, opened 2015-08-21
Add community_id to listings

rap1ds:
- [X] Add column
- [X] Copy the community_id from communities_listings
- [X] Listings#create now adds the community_id to new Listings
- [x] Add index


mporkola:
go!

PR #1412, opened 2015-08-21
Remove communities from listings

rap1ds:
- Use `listing.community_id`
- Do not use `listing.communities`
- However, save entries to `communities_listings` table, in case this doesn't work and we need to rollback. (So ignore the `TODO Remove this soon` comments. They are there on purpose.)

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -2,6 +2,11 @@
 class ListingsController < ApplicationController
   class ListingDeleted < StandardError; end
 
+  # TODO Remove me soon
+  # Gives us access to communities_listings join table
+  class CommunitiesListing < ActiveRecord::Base; end
+  # TODO Remove me soon
+
   include PeopleHelper
 
   # Skip auth token check as current jQuery doesn't provide it automatically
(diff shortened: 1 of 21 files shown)


app/controllers/listings_controller.rb
```diff
@@ -225,6 +230,9 @@ def create
     @listing.author = @current_user
 
     if @listing.save
+      # TODO Remove this soon
+      CommunitiesListing.create!(community_id: @current_community.id, listing_id: @listing.id)
```
ovan: This is not in the same TX so we can potentially lose data here still

app/models/listing.rb
```diff
@@ -132,17 +130,6 @@ def set_updates_email_at_to_now
   validates_inclusion_of :valid_until, :allow_nil => :true, :in => DateTime.now..DateTime.now + 7.months
   validates_numericality_of :price_cents, :only_integer => true, :greater_than_or_equal_to => 0, :message => "price must be numeric", :allow_nil => true
 
-  def set_community_visibilities
-    if current_community_id
```
ovan: What is this and do we still need it? Does some logic now work differently because of removing this before save hook?

rap1ds: Everything should work as before. In practise this piece of code was use to add the `@current_community` to the `communities` relation. Now we achieve the same by setting the id of the `@current_community` to `listing.community_id`.

PR #1413, opened 2015-08-25
Change Listing search index

rap1ds:
This pull request changes the listing search index so that it uses a single `community_id` instead of multiple community ids.

In addition, test code has been changed so that the tests pass.

All the other code is still using the `listing.communities` relation.

These are the same changes that are in PR #1412 , but I want to deploy the search index changes separately and first make sure that they work. After that, I'll continue with #1412


ovan:
good!

PR #1418, opened 2015-08-28
Remove unused `only_public_listings` column.

rap1ds:
- `communities.public` is the column that defines whether the community is public or not


ovan:
ok

PR #1419, opened 2015-08-28
Make sender email addresses lower-case

rap1ds:


app/services/email_service/store/address.rb
```diff
@@ -3,7 +3,7 @@ module EmailService::Store::Address
   NewAddress = EntityUtils.define_builder(
     [:community_id, :fixnum, :mandatory],
     [:name, :string, :optional],
-    [:email, :string, :mandatory],
+    [:email, :string, :mandatory, transform_with: ->(v) { v.downcase }],
```
ovan: I'd say this functionality belongs to service API instead of store because it's a service level concern, not a storing concern.

PR #1421, opened 2015-09-01
Explicitly say to jQuery that we expect the response to be HTML

rap1ds:
- Otherwise jQuery tries to guess the response type and in some corner cases it will guess it wrong


mporkola:
go ahead!

PR #1422, opened 2015-09-01
Speed up testimonials in use check

rap1ds:
`@current_community.testimonials_in_use` create a SQL query (or at least a cache hit) everytime when it's called. In the list view it is called once per listing.

Save the `@current_community.testimonials_in_use` result in a variable.


rap1ds:
The WTI pull commit is already in master, but for some reason Github doesn't understand it.

PR #1424, opened 2015-09-02
Fix bug in PayPal payment-review state

ovan:
Fixes two bugs in PayPal logic:
1. Add missing handling for PayPal payment-review IPN message. This is in practise a no-op message since we already know the payment state from synchronous response but previously it was handled incorrectly as pending ext message causing state to be overridden in DB.
2. Ensure that throughout the chain we use canonical storing for pending reason "payment-review", (with the dash). IPN messages send the information without dash but synchronous responses return it with dash.

Together these two bugs led to transactions been marked deleted when they should have been in payment-review state that normally prevents deleting the tokens. It also led to state change to authorized been ignored, i.e. when payment review passes we didn't notice.

### DIFF
--- a/app/services/paypal_service/data_types/ipn.rb
+++ b/app/services/paypal_service/data_types/ipn.rb
@@ -269,7 +269,7 @@ module PaypalService
 
         create_payment_review(
           p.merge({
-              authorization_total: to_money(p[:mc_gross_1], p[:mc_currency]),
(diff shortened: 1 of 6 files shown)


sktoiva:
Your comment to this pr is fantastic, but why not write it directly into some of the commits? Would give beneficial information later on in searchable format.

PR #1433, opened 2015-09-10
Fix bug: Shipping fee is not calculated

rap1ds:
Fixes bug:

screen shot 2015-09-10 at 09 54 27


ovan:
This fixes the issue by duplicating the payment calculation that would already happen correctly in transaction_service/transactions.rb. So I foresee this getting out of sync again. What is we just pull the transaction entity from the transaction service and use that here?

rap1ds:
As discussion in Flowdock, we'll focus on the duplication later on. Now this fix can go, even though it's not the cleanest one.

PR #1438, opened 2015-09-14
Use ListingIndexService on People#show and add SEO urls

rap1ds:
- Previous changes dropped the SEO friendly URLs (id + title). This PR will bring them back.
- People#show now uses the ListingIndexService. Because of this change now both homepage and profile page are using the same format for listings (structs). This way we can continue reusing the `grid_item.haml` that is also used in Homepage. 
- Now that ListingIndexService is used in many places, add a new ListingIndexViewUtils which takes care of transforming the ListingIndexService results (hash) to structs.
- Clean up `with_first_listing_image` helper. Now that it receives only one form of data (structs only, no models anymore) we can get rid of the smelly `respond_to`

### DIFF
--- a/app/services/listing_index_service/search/sphinx_adapter.rb
+++ b/app/services/listing_index_service/search/sphinx_adapter.rb
@@ -16,7 +16,7 @@ module ListingIndexService::Search
       included_models = includes.map { |m| INCLUDE_MAP[m] }
 
       if needs_db_query?(search) && needs_search?(search)
-        raise ArgumentError.new("Both DB query and search engine would be needed to full fill the search")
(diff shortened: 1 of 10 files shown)


app/services/listing_index_service/search/sphinx_adapter.rb
```diff
@@ -69,6 +85,7 @@ def search_with_sphinx(community_id:, search:, included_models:)
             listing_shape_id: search[:listing_shape_id],
             price_cents: search[:price_cents],
             listing_id: numeric_search_match_listing_ids,
+            author_id: search[:author_id]
```
mporkola: No `author_id` in sphinx

PR #1440, opened 2015-09-15
New language en-NZ

Gnito:
Added new language: English (New Zealand)

### DIFF
--- a/app/models/listing.rb
+++ b/app/models/listing.rb
@@ -23,6 +23,7 @@
 #  delta                           :boolean          default(TRUE), not null
 #  open                            :boolean          default(TRUE)
 #  share_type_old                  :string(255)
+#  privacy                         :string(255)      default("private")
 #  comments_count                  :integer          default(0)
 #  subcategory_old                 :string(255)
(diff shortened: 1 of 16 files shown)


app/models/listing.rb
```diff
@@ -23,6 +23,7 @@
 #  delta                           :boolean          default(TRUE), not null
 #  open                            :boolean          default(TRUE)
 #  share_type_old                  :string(255)
+#  privacy                         :string(255)      default("private")
```
rap1ds: This is not related to this PR. Should be removed.

rap1ds:
Good to go :+1:  There are some stuff in this PR that are unrelated, so a good practise in general is not to include any unrelated stuff to PR, but it doesn't matter that much. This can be merged.

Gnito:
Yeah, there was only 1 line in this whole PR that was worth a PR process (addition to available locales). So, I decided to add those generated privacy changes into this one too - along with wti pull stuff that normally goes to master directly. Maybe those generated stuff should be but to master directly too? (if they are not added on their original branch)

PR #1443, opened 2015-09-18
Usability improvement test: Separate link for PayPal help popup

rap1ds:
Some people have reported that it's confusing that pressing the credit card icons open a new PayPal popup, which actually is only a help popup, not for the payment.

This Pull Request adds a separate link text under the credit card icons that will open the popup. The link is removed from the credit card icons.

The feature is behind feature flag, so that we can test it first in one marketplace.

Before:

screen shot 2015-09-18 at 09 17 46

After:

screen shot 2015-09-18 at 09 16 12

### DIFF
--- a/app/assets/stylesheets/views/_listings.css.scss
+++ b/app/assets/stylesheets/views/_listings.css.scss
@@ -198,3 +198,7 @@ $listing-author-avatar-height: 108;
   width: 20px;
   color: $link;
 }
+
+.listings-how-paypal-works-link {
+  font-size: em(14);
+}
--- a/app/services/feature_flag_service/store.rb
+++ b/app/services/feature_flag_service/store.rb
@@ -8,7 +8,8 @@ module FeatureFlagService::Store
       [:features, :mandatory, :set])
 
     FLAGS = [
-      :location_search
+      :location_search,
+      :how_paypal_works_text_link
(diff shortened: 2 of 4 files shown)


vsaarinen:
Suggestion: what if we had a "Instructions for payment methods:" label between the Request button and the logos?

vsaarinen:
Ah, I thought the logos were still links to instructions. If they are no longer links, then having the label below makes more sense.

vsaarinen:
The current way of having multiple `if feature_enabled(...)` blocks made it a bit difficult to figure out the end result. Maybe have something like:

```
- if feature_enabled?(:how_paypal_works_text_link)
  = content_for(:paypal_credit_card_icons)
  %br/
  %a#how-paypal-works-popup-link{title: t("listings.listing_actions.how_paypal_works"), href: how_paypal_works_link}
    %span.listings-how-paypal-works-link
      = t("listings.listing_actions.learn_more_how_paypal_works")
- else
  %a#how-paypal-works-popup-link{title: t("listings.listing_actions.how_paypal_works"), href: how_paypal_works_link}
    = content_for(:paypal_credit_card_icons)
```

rap1ds:
@vsaarinen That's a good suggestion. Lot cleaner that way. I did the change.

PR #1446, opened 2015-09-22
Add "Export to CSV" option for marketplace users and transactions

vsaarinen:
Give admins the ability to export some of your marketplace's data: members and transactions.

For transactions, export the following data:

```
transaction_id, listing_id, listing_name, status, sum, commission, started_at, last_activity_at, starter_username, other_party_username
```

For users, export the following data:

```
first_name, last_name, username, email_address, email_address_confirmed, joined_at, status, is_admin, email_from_admins_allowed
```

TODO:
- [x] Export for Manage users page
- [x] Tests for users export
- [x] Export for Transactions page
- [x] Test for transactions export
- [x] Put it behind a feature flag?

### DIFF
--- a/app/views/admin/community_transactions/index.haml
+++ b/app/views/admin/community_transactions/index.haml
@@ -9,6 +9,8 @@
 .left-navi-section
   %h2= t("admin.communities.transactions.transactions", community_name: community.name(I18n.locale))
 
+  #export-as-csv.right= link_to(" " + t("admin.communities.transactions.export_as_csv"), {per_page: 99999, format: :csv}, {class: icon_class("download")})
+
   %span#admin_transactions_count= page_entries_info(conversations, :model => "Transaction")
 
   %table#admin_transactions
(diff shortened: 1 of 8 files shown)


config/locales/en.yml
```diff
@@ -253,6 +253,7 @@ en:
         ban_user: "Ban user"
         saving_user_status: Saving...
         save_user_status_successful: Saved
+        export_as_csv: "Export as CSV"
```
ovan: This link doesn't export exactly the same content that is shown on the screen because we export all and ignore pagination and sorting. How about saying "Export all as CSV" here to make it clear?

config/locales/en.yml
```diff
@@ -291,6 +292,7 @@ en:
         url: URL
         empty: "You don't have any additional menu links"
       transactions:
+        export_as_csv: "Export as CSV"
```
ovan: Same comment here, export all

app/views/admin/community_transactions/index.haml
```diff
@@ -9,6 +9,9 @@
 .left-navi-section
   %h2= t("admin.communities.transactions.transactions", community_name: community.name(I18n.locale))
 
+  - if feature_enabled?(:export_as_csv)
+    #export-as-csv.right= link_to(" " + t("admin.communities.transactions.export_all_as_csv"), {per_page: 99999, format: :csv}, {class: icon_class("download")})
```
rap1ds: I think this could be changed so that the `per_page` is not mandatory, and giving `nil` would mean that it should return everything.

rap1ds:
Very nice! Added one comment.

PR #1447, opened 2015-09-23
Limit the max zoom level

rap1ds:
- Nicer map view if there's only one listings
- This also limits the max zoom level, but imo it zooms close enough

Before:

screen shot 2015-09-23 at 16 24 45

After:

screen shot 2015-09-23 at 16 25 01


mporkola:
This makes sense :+1:

PR #1450, opened 2015-09-27
fix condition on listing shapes controller

artembiserov:
'is_admin' is an integer attribute, so we have to use #is_admin? method


rap1ds:
@artembiserov Thanks for the fix!

PR #1451, opened 2015-09-28
Add de and pt-BR translations for PayPal popup

rap1ds:


app/views/listing_conversations/_paypal_payment_methods.haml
```diff
@@ -1,8 +1,6 @@
-- how_paypal_works_link = "https://www.paypal.com/webapps/mpp/paypal-popup"
-
 = image_tag "https://www.paypalobjects.com/webstatic/en_US/i/buttons/cc-badges-ppmcvdam.png", style: "max-width: 100%"
 %br/
-%a#how-paypal-works-popup-link{title: t("listings.listing_actions.how_paypal_works"), href: how_paypal_works_link}
+%a#how-paypal-works-popup-link{title: t("listings.listing_actions.how_paypal_works"), href: t("paypal.how_paypal_works_link")}
   %span.listings-how-paypal-works-link
```
ovan: Is there any way to do this generically by adding locale into URL at correct spot instead of doing via translation files where every language has to be manually added and kept up-to-date? I think we do something like that already with other PayPal urls

PR #1454, opened 2015-09-28
Update PayPal signup link

Gnito:
Old link doesn't work and user ends up to 
https://www.paypal.com/webapps/mpp/get-started 
(with US locale)


ovan:
ok!

PR #1461, opened 2015-10-01
Better CSV export

mporkola:
- enclose values in quotes
- handle corrupt data better
- skip deleted users
- export also currency


ovan:
ok!

PR #1463, opened 2015-10-02
remove attributes from white list for mass assignment

artembiserov:
this code is unused, because we have protected attribute 'is_admin'
I know that it's better to determine white list attributes and I tried, but specs broke


rap1ds:
@artembiserov Thanks for the fix!

PR #1464, opened 2015-10-02
remove extra verification in listings controller

artembiserov:


rap1ds:
Thanks for fixing this!

PR #1467, opened 2015-10-05
Paypal fee link

rap1ds:
1. Add country specific links to PayPal fee page. Whitelist countries that should use the new fee page. Use the old page for others.
2) Fix the PayPal spinner padding:

Before:
screen shot 2015-10-05 at 11 35 02

After:
screen shot 2015-10-05 at 11 35 21


rap1ds:
Np: 💋🌹

PR #1469, opened 2015-10-06
Change to use word "someone" everywhere (instead of "somebody")

gnomet:
Also fix a missing preposition

Not sure if PR even made sense, but wanted to make sure that word changes won't break tests etc..


mporkola:
I think the test `features/settings/user_changes_notification_settings.feature` depends on the translation using "somebody", at least that what `git grep somebody` seems to suggest

PR #1470, opened 2015-10-06
Add user locale to CSV export data

mporkola:


rap1ds:
Looks good! The onclick handler could go to external JS file.

PR #1472, opened 2015-10-07
Handle missing listing_image_ids array gracefully

sktoiva:


app/controllers/listings_controller.rb
```diff
@@ -252,7 +252,19 @@ def create
...
+        listing_image_ids =
+          if params[:listing_images]
+            params[:listing_images].collect { |h| h[:id] }.select { |id| id.present? }
+          else
+            Rails.logger.error({tag: :error,
+                                free: "Listing images array is missing",
+                                structured: {params: params,
+                                             community_id: @current_community.id,
+                                             current_user: @current_user.id},
+                                request_id: request.env["HTTP_X_REQUEST_ID"]}.to_json)
```
rap1ds: I'm not sure if it's a good idea to add Heroku specific code here.

PR #1475, opened 2015-10-08
Add librato-rails and start reporting delayed job queue length

sktoiva:


app/controllers/application_controller.rb
```diff
@@ -396,6 +397,10 @@ def warn_about_missing_payment_info
     end
   end
 
+  def report_queue_length
+    Librato.increment 'delayed_job_queue', source: Delayed::Job.where("attempts < ?", 3).count
```
rap1ds: Add index to attempts

PR #1481, opened 2015-10-09
Expiration lightbox points to Charm

mporkola:


app/controllers/application_controller.rb
```diff
@@ -385,11 +385,23 @@ def fetch_community_plan_expiration_status
     }
   end
 
+  def charm_login_link(marketplace_id)
```
mporkola: this should probably be somewhere else

config/config.defaults.yml
```diff
@@ -238,6 +238,11 @@ default: &default_settings
   # This is completely optional and probably not needed in most cases.
   uservoice_widget_url:
 
+  # Charm connection
+  # The connection details hosted Sharetribe uses to connect to subscription management system Charm.
+  charm_secret: 
+  charm_url_base:
```
mporkola: There are specific to the hosted service, I wonder how we should control these and enable Charm only there

app/helpers/application_helper.rb
```diff
@@ -766,6 +766,23 @@ def payment_settings_url(gateway_type, person, url_params)
...
+    payload = {user_id: marketplace_id}
+    secret = APP_CONFIG.external_plan_service_jwt_secret
+    external_plan_service_url = APP_CONFIG.external_plan_service_url + "login"
+    token = JWTUtils.encode(payload, secret)
+    URLUtils.append_query_param(external_plan_service_url, "token", token)
+  end
+
+  def display_expiration_notice?
+    if feature_enabled?(:new_plan_page)
+      display = APP_CONFIG.external_plan_service_jwt_secret && @current_user && @current_user.has_admin_rights_in?(@current_community) && PlanUtils.expired?(@current_plan)
```
Gnito: This is a bit long line to read

app/helpers/application_helper.rb
```diff
@@ -766,6 +766,23 @@ def payment_settings_url(gateway_type, person, url_params)
...
+      @external_plan_service_link = external_plan_service_login_url(@current_community.id) if display
```
Gnito: Should this be passed through controller to haml instead of globals (i.e. a bit by bit remove all the globals away)

app/views/layouts/_expiration_notice.html.erb
```diff
@@ -5,17 +5,29 @@
   <div class="expiration-content">
     <div class="expiration-head-panel">
       <h1 class="expiration-title"><%= t("expiration.title") %></h1>
-      <h2 class="expiration-subtitle" ><%= t("expiration.sub_title") %></h2>
+      <h2 class="expiration-subtitle" >
+        <% if feature_enabled?(:new_plan_page) %><%= t("expiration.sub_title_new") %>
+        <% else %><%= t("expiration.sub_title") %>
+        <% end %>
+      </h2>
+      <% if feature_enabled?(:new_plan_page) %>
```
Gnito: N.B. I'm not sure how these should work with "APP_CONFIG.external_plan_service_in_use" flag

mporkola: It shouldn't be displayed at all then, see https://github.com/sharetribe/sharetribe/pull/1481/files#diff-127a850b7d7b20173ad3fba88f8a25e3R779

app/assets/stylesheets/main.css.scss
```diff
@@ -649,133 +649,6 @@ a.lightbox-x {
...
-  margin-top: 10px;
-}
-.expiration-footer {
-  padding: 30px 10px;
-}
-.expiration-alternative-plans {
-  font-size: 14px;
-  line-height: 14px;
-  text-align: center;
-}
```
rap1ds: Very nice! :+1: All the code moved away from main.css is great :+1:

app/helpers/application_helper.rb
```diff
@@ -766,6 +766,25 @@ def payment_settings_url(gateway_type, person, url_params)
...
+  end
+
+  def display_expiration_notice?
+    if feature_enabled?(:new_plan_page)
+      display = APP_CONFIG.external_plan_service_in_use && 
+        @current_user && @current_user.has_admin_rights_in?(@current_community) &&
+        PlanUtils.expired?(@current_plan)
+      @external_plan_service_link = external_plan_service_login_url(@current_community.id) if display
+    else
+      @is_current_community_admin && PlanUtils.expired?(@current_plan)
```
rap1ds: Now that the `@is_current_community_admin` is not used in the application layout anymore (which is definitely good!) I'd suggest removing the whole `@is_current_community_admin` variable, since it's "almost global variable". Call `@current_user.has_admin_rights_in?(@current_community)` instead.

app/helpers/application_helper.rb
```diff
@@ -766,6 +766,25 @@ def payment_settings_url(gateway_type, person, url_params)
...
+    token = JWTUtils.encode(payload, secret)
+    URLUtils.append_query_param(external_plan_service_url, "token", token)
+  end
+
+  def display_expiration_notice?
+    if feature_enabled?(:new_plan_page)
+      display = APP_CONFIG.external_plan_service_in_use && 
+        @current_user && @current_user.has_admin_rights_in?(@current_community) &&
+        PlanUtils.expired?(@current_plan)
+      @external_plan_service_link = external_plan_service_login_url(@current_community.id) if display
```
rap1ds: This is a bit confusing. The name of the method suggests that the method returns boolean and does nothing else. However, here we set a new almost global variable.

What if you change the view code so, that you explicitely call the `external_plan_service_login_url` in the view? (See the comment for `application.haml`

app/views/layouts/application.haml
```diff
@@ -8,7 +8,7 @@
 
   = render :partial => 'layouts/global-header'
 
-  - if @is_current_community_admin && PlanUtils.expired?(@current_plan)
+  - if display_expiration_notice?
```
rap1ds: How about this:

```haml
- if display_expiration_notice?
  = render partial: "layouts/expiration_notice", external_plan_service_login_url: external_plan_service_login_url(@current_community.id)
```

app/helpers/application_helper.rb
```diff
@@ -766,6 +766,25 @@ def payment_settings_url(gateway_type, person, url_params)
...
+    secret = APP_CONFIG.external_plan_service_secret
+    external_plan_service_url = APP_CONFIG.external_plan_service_url + "login"
+    token = JWTUtils.encode(payload, secret)
+    URLUtils.append_query_param(external_plan_service_url, "token", token)
+  end
+
+  def display_expiration_notice?
+    if feature_enabled?(:new_plan_page)
+      display = APP_CONFIG.external_plan_service_in_use && 
+        @current_user && @current_user.has_admin_rights_in?(@current_community) &&
```
rap1ds: Instead of 

```
@current_user && @current_user.has_admin_rights_in?(@current_community)
```

You can also use `Maybe` here:

```
Maybe(@current_user).has_admin_rights_in?(@current_community).or_else(false)
```

Matter of taste.

app/helpers/application_helper.rb
```diff
@@ -766,6 +766,25 @@ def payment_settings_url(gateway_type, person, url_params)
...
+    payload = {user_id: marketplace_id}
+    secret = APP_CONFIG.external_plan_service_secret
+    external_plan_service_url = APP_CONFIG.external_plan_service_url + "login"
+    token = JWTUtils.encode(payload, secret)
+    URLUtils.append_query_param(external_plan_service_url, "token", token)
+  end
+
+  def display_expiration_notice?
+    if feature_enabled?(:new_plan_page)
+      display = APP_CONFIG.external_plan_service_in_use &&
```
rap1ds: In this branch https://github.com/sharetribe/sharetribe/pull/1503/ I made a change which returns always a valid plan if external plan service is not in use. So after that branch is merged, you could simplify this to:

```
Maybe(@current_user).has_admin_rights_in?(@current_community).or_else(false) && PlanUtils.expired?(@current_plan)
```

...which is basically the current else branch

app/helpers/application_helper.rb
mporkola: Good point, this is a good example of incrementally tweaking existing logic until it makes no sense. The evolution started from `fetch_chargebee_plan_data` in `ApplicationController`, which was all about setting globals.

app/helpers/application_helper.rb
```diff
@@ -766,6 +766,24 @@ def payment_settings_url(gateway_type, person, url_params)
...
+  def external_plan_service_login_url(marketplace_id)
+    payload = {user_id: marketplace_id}
+    secret = APP_CONFIG.external_plan_service_secret
+    external_plan_service_url = APP_CONFIG.external_plan_service_url + "login"
+    token = JWTUtils.encode(payload, secret)
+    URLUtils.append_query_param(external_plan_service_url, "token", token)
+  end
+
+  def display_expiration_notice?
+    if feature_enabled?(:new_plan_page)
```
rap1ds: I think we can now remove the feature_enabled, can we?

mporkola: Then the lightbox wouldn't be shown to marketplaces without the enabled feature flag, unless `APP_CONFIG.external_plan_service_in_use` is defined?

mporkola: Of course if we decide to turn `external_plan_service_in_use` on in prod, then the branches are equal

PR #1484, opened 2015-10-12
Simplify Google Analytics domain name setting, by using the whole host

gnomet:
Earlier we remove the subdomain part to pool subdomains, but that's
not actively needed now, and caused trouble as the script didn't
take acocunt naked domains

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -84,6 +84,8 @@ gem 'cache_digests'
 gem 'librato-rails'
 
 gem 'lograge'
+gem 'public_suffix' # Needed currently to set GA hostname right, probably not
+                    # needed anymore when GA script updated.
 
 group :staging, :production do
   gem 'newrelic_rpm', '~> 3.9.1.236'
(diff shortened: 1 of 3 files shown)


gnomet:
Can't deploy this fix, could cause issues on clients with separate landing pages (or other multi subdomain solutions)
Will rollback to earlier solution and do a quick fix for .com.br domains which is bit stupid, but will continue our client's analytics data flow for now, and the real fix is to update from this deprecated GA script to the newest version.

gnomet:
Noted that there's no good way to remove subdomain reliably without full list of public suffixes people use to register their domains, so used PublicSuffix gem to do the trick that maintains the list. (http://stackoverflow.com/a/8323839/150382)

Earlier solutions had trouble either with. co.uk com.br etc. TLDs or with naked domains. This should work with all. And when we update our GA script (should be done soon anyway) Google will take care of this part, so we can also remove PublicSuffix gem. :) (I put a comment about that to Gemfile)

PR #1485, opened 2015-10-12
Remove unused code from Person

rap1ds:


gnomet:
Looks good. Not 100% sure about remove_root_level_fields and chache related methods (I mean I'm not 100% sure of where they were used and if not used anymore, but the merge stuff can go. And I trust that you did some sort of global search to check that the rest was not used either, so looks like a very nice cleanup operation! :+1:

rap1ds:
Yep, did some global searching.

PR #1486, opened 2015-10-12
Implement a Logger (a Rails.logger wrapper)

rap1ds:
### Description

Implement a thin logger around the default `Rails.logger`. It takes care of JSON formatting and adds additional details, such as `community_id` and `user_id`

### The Problem

**JSON logging:** We should log everything in JSON format to allow Loggly to parse the log messages. However, we don't have tools that actually encourage JSON logging.

**Passing additional information:** We have a lot of valuable information that should be passed with each log message (community_id, user_id, request_id). Adding those pieces of information would make debugging a lot easier.

**Copy-paste code** PayPal Service and SES Service already implement a logger wrapper, and they are basically copy-paste code

**Testing logging:** In some places, logging is critical operation (receiving webhooks etc.) and thus it should be tested. Testing side-effects (STDOUT, in this case) is pretty damn difficult, so the wrapper allows us to inject the logging target, which in case of tests can be something that makes testing easier.

### The Solution

This pull request implements a class `SharetribeLogger`. The constructor takes `tag` and `logging_target` parameters.

```ruby
logger = SharetribeLogger.new(:paypal, Rails.logger)
```

**JSON logging:**

The logger implements all 4 logging level methods `debug`, `info`, `warn` and `error`. Each method takes two arguments: `msg` and `type`. From these information and additional information, such as `community_id`, the logger outputs the following JSON string:

```ruby
# in controller
logger.info("Free message, for convenience", :testing, {testing_logging: true})

INFO {"user_id":"bTarZQWnKr5iaakm_Phqdd","username":"rap1ds","community_id":36,"community_ident":"kuha","request_uuid":"2341824a3bc65bf730ccf19657a4728b","tag":null,"free":"Free message, for convenience","type":"testing","structured":{"testing_logging":true}}
```

**Passing additional information:** Controllers implement a method `logger`, which returns the default `Rails.logger`. This PR overrides that method in ApplicationController. The new `logger` method returns an instance of SharetribeLogger. This PR also adds code that sets the current `community_id` and `user_id` to the logger, so that those are printed when ever something is logged from the controller context.

**Copy-paste code:** The SharetribeLogger can be instantiated with a specific that, and that way it in practise replaces the PaypalService::Logger:

```ruby
logger_for_paypal = SharetribeLogger.new(:paypal, Rails.logger)
```

PaypalService::Logger contains also helpers such as `log_response`. That's something the is not in the SharetribeLogger, so options are either extract these methods to another helper/util class, or make a new Logger class that inherits everything from SharetribeLogger

**Testing logging:**

For testing purposes, we could inject a logger that has a logging target which is something else than Rails.logger. This is not yet implemented, so I don't know exact details, but it could be something like this:

```ruby
# some spec file
before(:each) do
  @test_logger = TestLogger.new
  logger = SharetribeLogger.new(:paypal, @test_logger)
  api = PayPalService::API::Payments.new(logger)
end

it "tests logging"
  expect(@test_logger.logged_errors.count).to eq(0)
  @api.do_some_error_operation(:error)
  expect(@test_logger.logged_errors.count).to eq(1) # logger.error has been called once
  expect(@test_logger.logged_errors.first[:free]).to eq("The operation failed")
end
```

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -166,6 +166,8 @@ class ApplicationController < ActionController::Base
   def fetch_logged_in_user
     if person_signed_in?
       @current_user = current_person
+      logger.user_id = @current_user.id
+      logger.username = @current_user.username
     end
   end
 
@@ -211,6 +213,9 @@ class ApplicationController < ActionController::Base
   # Before filter to get the current community
   def fetch_community
     @current_community = ApplicationController.find_community(community_identifiers)
+
+    logger.community_id = @current_community.id
+    logger.community_ident = @current_community.ident
 
     # Save :found or :not_found to community status
     # This is needed because we need to distinguish to cases
@@ -503,6 +508,10 @@ class ApplicationController < ActionController::Base
   end
 
   helper_method :fetch_feature_flags # Make this method available for FeatureFlagHelper
+
+  def logger
+    @logger ||= SharetribeLogger.new
+  end
 
   # Fetch temporary flags from params and session
   def self.fetch_temp_flags(is_admin, params, session)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -214,8 +214,9 @@ class ApplicationController < ActionController::Base
   def fetch_community
     @current_community = ApplicationController.find_community(community_identifiers)
 
-    logger.community_id = @current_community.id
-    logger.community_ident = @current_community.ident
+    m_community = Maybe(@current_community)
+    logger.community_id = m_community.id
+    logger.community_ident = m_community.ident
 
     # Save :found or :not_found to community status
     # This is needed because we need to distinguish to cases
@@ -510,7 +511,12 @@ class ApplicationController < ActionController::Base
   helper_method :fetch_feature_flags # Make this method available for FeatureFlagHelper
 
   def logger
-    @logger ||= SharetribeLogger.new
+    if @logger.nil?
+      @logger = SharetribeLogger.new
+      @logger.request_uuid = request.uuid
+    end
+
+    @logger
   end
 
   # Fetch temporary flags from params and session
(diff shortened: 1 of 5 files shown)


app/utils/sharetribe_logger.rb
```diff
@@ -0,0 +1,55 @@
...
+    }
+  end
+
+  def add_details(log_data)
+    {
+      user_id: @user_id,
+      username: @username,
+      community_id: @community_id,
+      community_ident: @community_ident,
+      request_uuid: @request_uuid
```
ovan: If this is nil should we leave this field out? Locally we won't have a uuid in request headers.

app/controllers/application_controller.rb
```diff
@@ -212,6 +214,10 @@ def save_current_host_with_port
   def fetch_community
     @current_community = ApplicationController.find_community(community_identifiers)
 
+    m_community = Maybe(@current_community)
+    logger.community_id = m_community.id
+    logger.community_ident = m_community.ident
```
ovan: This feels like random global scope mutating. How about a wrapper method to set these to logger instance so it would at least be clear who manages the logger instance:

```ruby
setup_logger(community_id: m_community.id, community_ident: community_ident: m_community.ident)
```

app/utils/sharetribe_logger.rb
```diff
@@ -0,0 +1,55 @@
+class SharetribeLogger
+  attr_writer(
+    :community_id,
+    :community_ident,
```
ovan: Should we call these marketplace_id and marketplace_ident as those are the new terms even though the data model under the hood still talks about communities?

ovan:
Good idea and solid design. When needing context specific logging functionality, like PayPal's log_response and log_request, the best option in my opinion is to wrap this instead of subclassing.

PR #1492, opened 2015-10-14
Instruction images could be way too wide for mobile layouts

Gnito:


mporkola:
saw demo, was convinced

PR #1494, opened 2015-10-14
Fix "How it works" page and remove some old code

gnomet:
Make "how it works" page act like other pages, and prevent a bug.

The bug case was that by saving empty how it works page and admin
could make it disappear without a possibility to bring it back

Also spotted some unused code (dashboard.haml) and deleted it


config/locales/en.yml
```diff
@@ -1208,6 +1207,7 @@ en:
     about:
       default_about_text_title: "What is Sharetribe?"
       default_about_text: "This marketplace is powered by Sharetribe platform. With Sharetribe you can easily create your own marketplace website. It's free and only takes a minute. %{click_here_link} to learn more!"
+      default_how_to_use_text: "<h1>How it works</h1><div>Here you can find information about how this site works.</div>"
```
rap1ds: How about this?

```
# en.yml
default_how_to_use_title: "How it works"
default_how_to_use_content: "Here you can find information about how this site works."
```

```
...
- else
  - content = "<h1>#{t('infos.about.default_how_to_use_title')}</h1><div>#{t('infos.about.default_how_to_use_content')}</div>"
...
```

It's not allowed to add any HTML/HAML markup to translation files. Ever. No exceptions. It has already caused us quite a lot of trouble (and will cause more when we have more languages and translators) and we don't want to repeat those mistakes 😉

app/views/infos/how_to_use.haml
```diff
@@ -7,7 +7,10 @@
 
 .left-navi-section.about-section
   = render :layout => "layouts/mercury_editable_page", :locals => { :content_type => "how_to_use_page_content" } do
-    - content = how_to_use_content.html_safe
+    - if @community_customization && ! @community_customization.how_to_use_page_content.nil?
```
rap1ds: Extra space

app/helpers/application_helper.rb
```diff
@@ -537,14 +537,12 @@ def about_links
         :name => "about"
       }
     ]
-    if @community_customization && !@community_customization.how_to_use_page_content.blank?
```
rap1ds: If I understood correctly, this will now change the behaviour so that marketplaces which do NOT have any how it works content (on purpose?) will now have the link there and it will take them to blank page. How should we handle those cases? I see two options:

1. Let's decide that there's no way to hide the How to use page. The page will be there, always. (Basically that's what this pull request does)
2. Let's add a new database column `how_to_use_page_in_use` and based on the value on that column we can decide whether or not show the link. This would also require some UI work (add a checkbox "Show How To Use page" or something)

I'd say number 1 is good, but then we need to handle somehow those marketplaces that actually on purpose want to hide the how to use page.

rap1ds: Oh, apparently this has been solved already on Flowdock so you can ignore this comment.

config/locales/en.yml
gnomet: Yep, good point!
I somehow didn't stop to think at this point at all, just copied the placeholder there from initial community_customizations row. Definitely good to now put html there!
Thanks for catching!

app/views/infos/how_to_use.haml
gnomet: do you mean there shouldn't be space after !
yea, it might be bit uncommon. I've maybe done that few times as I myself tend to miss the ! if it's connected to the word. But better to have uniform convention and me learning to look right than changing it because of that.. :)

app/helpers/application_helper.rb
gnomet: Yep, seemed to be so narrow use case that we can go with 1) which this PR does.

gnomet:
Now fixed the commented parts, by using existing method that I found.
I assume it's ok to call MarketplaceService::API::Marketplaces::Helper.how_to_use_page_default_content from the HAML file?

gnomet:
Still need to move method call (and whole if structure) from HAML to controller and pass as attribute the content.

gnomet:
Now fixed the commented part, probably no need for anymore review rounds, so will merge when tests are green..

PR #1496, opened 2015-10-14
Make Sphinx calls more resilient to errors

ovan:
* Change listingindex to expect search adapters to return Result types
* Adapt existing sphinx adapter to return Result types
* Add timeout to sphinx queries
* Handle sphinx errors, such as no connection to server, by wrapping to Result::Error
* Log errors before returning from service and show err msg in UI

### DIFF
--- a/app/services/listing_index_service/api/api.rb
+++ b/app/services/listing_index_service/api/api.rb
@@ -7,7 +7,7 @@ module ListingIndexService::API
     end
 
     def self.listings
-      self.listings_api ||= ListingIndexService::API::Listings.new
+      self.listings_api ||= ListingIndexService::API::Listings.new(Rails.logger)
     end
   end
 end
--- a/app/services/listing_index_service/search/sphinx_adapter.rb
+++ b/app/services/listing_index_service/search/sphinx_adapter.rb
@@ -114,7 +114,7 @@ module ListingIndexService::Search
           with: with,
           with_all: with_all,
           order: 'sort_date DESC',
-          max_query_time: 500 # Timeout and fail after 500ms
+          max_query_time: 1000 # Timeout and fail after 500ms
         )
 
         begin
--- a/app/services/listing_index_service/search/sphinx_adapter.rb
+++ b/app/services/listing_index_service/search/sphinx_adapter.rb
@@ -16,26 +16,34 @@ module ListingIndexService::Search
       included_models = includes.map { |m| INCLUDE_MAP[m] }
 
       if needs_db_query?(search) && needs_search?(search)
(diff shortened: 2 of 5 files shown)


app/services/listing_index_service/api/api.rb
```diff
@@ -7,7 +7,7 @@ class << self
     end
 
     def self.listings
-      self.listings_api ||= ListingIndexService::API::Listings.new
+      self.listings_api ||= ListingIndexService::API::Listings.new(Rails.logger)
```
rap1ds: You could (and should?) use the new and shiny SharetribeLogger.

```ruby
ListingIndexService::API::Listings.new(SharetribeLogger.new(:listing_index_service))
```

app/services/listing_index_service/search/sphinx_adapter.rb
```diff
@@ -105,10 +113,15 @@ def search_with_sphinx(community_id:, search:, included_models:)
...
-          order: 'sort_date DESC'
+          order: 'sort_date DESC',
+          max_query_time: 500 # Timeout and fail after 500ms
         )
 
-        {count: models.total_entries, listings: models}
+        begin
+          success_result(models.total_entries, models, includes)
+        rescue ThinkingSphinx::SphinxError => e
+          Result::Error.new(e)
```
rap1ds: This is very nice! 

I have seen two types of error thrown by Sphinx that we should catch:

1. In development, Sphinx is not running.
2. In production, Sphinx is reindexing.

Have you tested does this now catch them both? I.e. have you tested that the exception thrown from those two cases are `ThinkingSphinx::SphinxError` or subclasses of it?

rap1ds:
Maybe not part of this pull request, but the flash notification for Resut::Error is now "Something went wrong", in a red background.

If the error occured becase the developer didn't have Sphinx running, we could be more informative and say "Could not connect to the search engine", for example. If Sphinx is reindexing in the production, we could say "Search engine is currently unavailable. Please try again in a minute"

app/services/listing_index_service/api/api.rb
ovan: kewl, I was thinking about that but didn't realize it's already in master. Will definitely change.

app/services/listing_index_service/search/sphinx_adapter.rb
ovan: I haven't tested the indexing thing but this is what the Sphinx-documentation says, all errors raised are SphinxErrors. The case of sphinx not running in development is just a connection error, so it's indistinguishable from a network error in production or production sphinx being down.

PR #1497, opened 2015-10-15
Add a link to open the listing address in Google Maps

vsaarinen:
When viewing a list, it would be useful to be able to get directions to the location. Add a link to open the address in Google Maps, which provides directions. When opening on mobile, users can also open the Google Maps app.

screenshot 2015-10-15 12 24 26

### DIFF
--- a/app/assets/stylesheets/views/_listings.css.scss
+++ b/app/assets/stylesheets/views/_listings.css.scss
@@ -159,6 +159,12 @@ $listing-author-avatar-height: 108;
   margin: lines(0.5) 0; // Some top & bottom margin to give links some space
 }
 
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/views/_listings.css.scss
```diff
@@ -159,6 +159,12 @@ $listing-author-avatar-height: 108;
   margin: lines(0.5) 0; // Some top & bottom margin to give links some space
 }
 
+.map-link {
+  float: right;
+  font-size: 14px;
```
mporkola: we've generally used the em-size helper, in this case `em(14)`

PR #1501, opened 2015-10-22
Use bold font only for unread messages

rap1ds:
Small usability improvement: In inbox, use bold font only for unread messages. This makes is easier to differentiate unread and read messages.

Before:

screen shot 2015-10-22 at 08 53 39

After:

screen shot 2015-10-22 at 08 53 11


vsaarinen:
lgtm :shipit:

PR #1503, opened 2015-10-22
Add marketplace_plans and marketplace_trials tables

rap1ds:
Phase 1 (This PR):

- [X] Write all plans to `community_plans` table
- [X] Write trial plans to `marketplace_trials` table
- [X] Write non-trial plans to `marketplace_plans` table
- [ ] Deploy to production
- [ ] Setup extermal plan service to send plan provisioning events
- [ ] In ext. plan service, do the initial data import and send plan provisioning events
- [ ] Check that the data is correct

Phase 2 (Another PR)
- [ ] Read data only from `marketplace_trials` and `marketplace_plans`
- [ ] Check that everything works

Phase 3 (Another PR)
- [ ] Drop the `communities_plans` table

Other changes in this PR:
- [X] Marketplace API doesn't create plan anymore. Its job is to create marketplaces, not plans.
- [X] For OS installations (where environment variable `external_plan_service_in_use` is false) we don't create plans
- [X] For OS installations, we always return OS plan (level 99999)


app/controllers/int_api/marketplaces_controller.rb
```diff
@@ -22,7 +22,14 @@ def create
                    :marketplace_country,
                    :marketplace_language)
             .merge(payment_process: :preauthorize)
-      )
+    )
+
+    # Create initial trial plan
+    plan = {
+      plan_level: PlanUtils::FREE,
```
ovan: Not needed because we are already calling the `create_initial_trial`

app/services/plan_service/api/plans.rb
```diff
@@ -1,34 +1,99 @@
 module PlanService::API
   PlanStore = PlanService::Store::Plan
 
+  Plan = PlanService::DataTypes::Plan
+
   class Plans
 
+    def initialize(ext_plan_service)
+      @active = Maybe(ext_plan_service)[:active].map {
+        |v| !!v # to bool
```
ovan: already handled by datatype coercion

app/services/plan_service/api/plans.rb
```diff
@@ -1,34 +1,99 @@
...
+      }.or_else(false)
+    end
+
     def create(community_id:, plan:)
-      Result::Success.new(
-        with_expiration_status(
-          PlanStore.create(community_id: community_id, plan: plan)))
+      if !@active
```
ovan: We could get rid of all the active checks by having separate implementation for OS use case and external plan service use case. Injector would then choose which one to put in place.

ovan: Actually we can get rid of the whole injector and just do the iffing and building in api.rb directly.

app/utils/plan_utils.rb
```diff
@@ -9,10 +9,13 @@ module PlanUtils
   PRO = 2
   GROWTH = 3
   SCALE = 4
+  HOLD = 5
+  OS = 99999
```
ovan: These numbers don't make a whole lot sense because comparing in number order is now broken (hold > scale). We should drop the number semantics and store these as symbols instead.

db/migrate/20151022180225_create_marketplace_trials_table.rb
```diff
@@ -0,0 +1,12 @@
+class CreateMarketplaceTrialsTable < ActiveRecord::Migration
+
+  def change
+    create_table :marketplace_trials, :force => true do |t|
```
ovan: What does the :force => true do?

app/controllers/plans_controller.rb
```diff
@@ -125,16 +123,18 @@ def to_entity(hash)
...
       logger.error("Unauthorized", nil, token: params[:token])
       render json: {error: :unauthorized}, status: 401
     }
   end
 
   def ensure_external_plan_service_in_use!
-    unless external_plan_service[:active]
-      raise ActiveRecord::RecordNotFound
-    end
+    raise ActiveRecord::RecordNotFound unless plans_api.active?
```
ovan: Why ActiveRecord::RecordNotFound?

rap1ds: It shows 404

ovan: Ok. This doesn't feel like very obvious way to do that because it doesn't communicate the intent. Also, wouldn't this be something that can change?

app/services/plan_service/api/api.rb
```diff
@@ -1,7 +1,34 @@
 module PlanService::API
   class Api
+    Configuration = PlanService::DataTypes::Configuration
+
     def self.plans
-      @plans ||= PlanService::API::Plans.new
+      configuration = build_configuration()
```
ovan: Minor thing probably but now every time the plans is looked up the configuration gets rebuilt unnecessarily when the caching ```@plans ||= ...``` statement comes only after.

Issue #1504, opened 2015-10-23
gitignored files not removed


kimroen:
Some files listed in `.gitignore` are not actually removed from this repository.

The list includes a few things that we actually want, I believe they were added by mistake.

Things of note are:

- `Gemfile`
- `Gemfile.lock`

Run this to stage everything that would be deleted from git:

```
git rm -r --cached .
git add .
git status
```

Full list:

```
Gemfile
Gemfile.lock
app/assets/images/map_icons/._favor_offer.png
app/assets/images/map_icons/._favor_request.png
app/assets/images/map_icons/._group_all.png
app/assets/images/map_icons/._group_offer.png
app/assets/images/map_icons/._group_request.png
app/assets/images/map_icons/._housing_offer.png
app/assets/images/map_icons/._housing_request.png
app/assets/images/map_icons/._item_offer.png
app/assets/images/map_icons/._item_request.png
app/assets/images/map_icons/._rideshare_offer.png
app/assets/images/map_icons/._rideshare_request.png
```

kimroen:
Looks like I'm slightly wrong about the file list, or at least the cause. The map_icons are not gitignored, they are merely hidden files. They do seem to be redundant and not used though, so I think they should probably me removed. Introduced here, a long time ago: b4295076f3cb473496fe5f733fd7a2b19768b461

rap1ds:
Thank you so much for reporting and fixing the issue! I guess now all the files are removed and this issue can be closed?

kimroen:
Now they are indeed :)

PR #1505, opened 2015-10-23
Don't ignore the Gemfile

kimroen:
Partially addresses #1504 

This seems to have been introduced by accident. Please let me know if that is incorrect.


mporkola:
You're right, no reason to have them in `.gitignore`. Thanks!

PR #1506, opened 2015-10-23
Delete hidden, unused image files

kimroen:
Partially addresses #1505 

These files are not in use, and has not been touched for a few years.


rap1ds:
@kimroen Thanks! Sorry for some delay!

PR #1512, opened 2015-10-28
Charm jwt token payload

Gnito:


app/helpers/application_helper.rb
```diff
@@ -801,11 +801,15 @@ def payment_settings_url(gateway_type, person, url_params)
 
   def external_plan_service_login_url(marketplace_id)
     if APP_CONFIG.external_plan_service_url && APP_CONFIG.external_plan_service_secret
-      payload = {user_id: marketplace_id}
-      secret = APP_CONFIG.external_plan_service_secret
-      external_plan_service_url = APP_CONFIG.external_plan_service_url + "login"
-      token = JWTUtils.encode(payload, secret)
-      URLUtils.append_query_param(external_plan_service_url, "token", token)
+      default_locale = @current_community.locales[0]
```
rap1ds: Use `@current_community.default_locale`

app/services/plan_service/store/plan.rb
```diff
@@ -49,6 +49,15 @@ def get_current(community_id:)
     from_model(plan_model)
   end
 
+  def get_trial(community_id:)
```
rap1ds: Rebase and use `get_initial_trial`

app/helpers/application_helper.rb
```diff
@@ -801,11 +801,15 @@ def payment_settings_url(gateway_type, person, url_params)
 
   def external_plan_service_login_url(marketplace_id)
     if APP_CONFIG.external_plan_service_url && APP_CONFIG.external_plan_service_secret
```
rap1ds: You can remove this if-branch. Just ask the get_external_service_link from the plan service. The `NoPlans` should then return empty link (or Error result) if external plan service is not in use.

app/services/plan_service/api/plans.rb
```diff
@@ -73,6 +74,31 @@ def get_trials(after:)
...
+              trial_hash = CharmPlan.call(HashUtils.rename_keys({
+                id: :marketplace_plan_id,
+                community_id: :marketplace_id
+              }, trial_data))
+              payload = {
+                marketplace: marketplace_data,
+                initial_trial_data: trial_hash
+              }
+
+              secret = APP_CONFIG.external_plan_service_secret
```
rap1ds: You should not refer the `APP_CONFIG` directly here. Instead, you should add an `initialize` method here, which takes the configurations as a parameter. The api/api.rb will provide the configurations.

app/services/plan_service/api/plans.rb
```diff
@@ -73,6 +74,31 @@ def get_trials(after:)
...
+              url = external_plan_service_url + "/login"
+              token = JWTUtils.encode(payload, secret)
+              URLUtils.append_query_param(url, "token", token)
```
rap1ds: Since you're returning `Result` in the error branch, you should return also here a `Result`.

app/helpers/application_helper.rb
Gnito: true

app/helpers/application_helper.rb
Gnito: roger

app/services/plan_service/api/plans.rb
Gnito: roger
yeah, this was copy-paste without checking what has been changed regarding APP_CONFIG usage.

app/services/plan_service/api/plans.rb
Gnito: Good point. 
(Result seems to be a bit overkill for fetching link for (currently) internal use, but it's in line with API practices)

app/helpers/application_helper.rb
```diff
@@ -589,8 +589,8 @@ def admin_links_for(community)
         :topic => :general,
         :text => t("admin.left_hand_navigation.subscription"),
```
Gnito: Todo: need to be changed through wti

PR #1513, opened 2015-10-28
Add users' phone numbers and addresses to CSV user export

vsaarinen:
Has not been tested for performance, but shouldn't be dramatically slower than previous version.

### DIFF
--- a/app/controllers/admin/community_memberships_controller.rb
+++ b/app/controllers/admin/community_memberships_controller.rb
@@ -24,7 +24,7 @@ class Admin::CommunityMembershipsController < ApplicationController
         else
           @community.ident
         end
-        send_data generate_csv_for(all_memberships), filename: "#{marketplace_name}-users-#{Date.today}.csv"
+        send_data generate_csv_for(all_memberships, @community), filename: "#{marketplace_name}-users-#{Date.today}.csv"
       end
     end
   end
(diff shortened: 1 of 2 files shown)


gnomet:
Thanks for fast action taking here!
And delightfully small amount of code changes needed. :)
I noticed that there was a test added for phone number but not for location, although testing location might be more important as joined model. (although requires bit more model setup then too).. 

Travis is on red :/ otherwise good to go IMO

app/controllers/admin/community_memberships_controller.rb
```diff
@@ -91,6 +93,8 @@ def generate_csv_for(memberships)
             user.given_name,
```
sktoiva: On line 89, use 'find_each' instead of 'each'.

app/controllers/admin/community_memberships_controller.rb
```diff
@@ -17,7 +17,7 @@ def index
       format.csv do
         all_memberships = CommunityMembership.where(:community_id => @community.id)
                                               .where("status != 'deleted_user'")
-                                              .includes(:person => :emails)
```
sktoiva: Use `joins` instead of `includes`. As I understand it, it generates more queries, but allows more effective cache usage and in this case, more effective ActiveRecord instantiation. On large marketplaces lowers the query time from > 20 seconds to ~10 seconds.

app/controllers/admin/community_memberships_controller.rb
vsaarinen: [This page](http://apidock.com/rails/ActiveRecord/Batches/ClassMethods/find_each) says that find_each ignores order_by. Should it really be used?

vsaarinen: Since find_each sorts by id ASC, the end result is the same as with sorting by creation date. Went ahead with find_each.

PR #1514, opened 2015-10-30
Trial pagination

rap1ds:
- [X] Add pagination for trial synchronization
- [X] Do not write new plans (that are not coming from the external plan service) to `marketplace_plans` table
- [X] Trial synchronize should return plan ID from `marketplace_plans` table.

### DIFF
--- a/app/services/plan_service/api/plans.rb
+++ b/app/services/plan_service/api/plans.rb
@@ -81,15 +81,7 @@ module PlanService::API
       }
     end
 
-    # Returns paginated trials
-    #
-    # Result:
-    #
-    # ```
-    # { plans: [ ... ],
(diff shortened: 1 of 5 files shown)


app/services/plan_service/api/plans.rb
```diff
@@ -69,8 +81,30 @@ def expired?(community_id:)
...
 
-    def get_trials(after:)
-      Result::Success.new(PlanStore.get_trials(after: after))
+    # Returns paginated trials
+    #
+    # Result:
+    #
+    # ```
+    # { plans: [ ... ],
+    #   next_offset: 1234567890
```
rap1ds: next_after

PR #1516, opened 2015-10-31
Use new plan tables

rap1ds:
- Use new plan tables
- Do some renaming


app/services/plan_service/store/plan.rb
```diff
@@ -35,29 +33,31 @@ def create_trial(community_id:, plan:)
     from_trial_model(TrialModel.create!(plan_entity))
   end
 
-  def create_plan(community_id:, plan:)
-    plan_entity = NewPlan.call(plan.merge(community_id: community_id))
-    from_model(PlanModel.create!(plan_entity))
+  # TODO When we are ready to read from the new plans table, some renaming is needed
```
ovan: Renameing done

ovan:
Can you ensure that these new plan tables have relevant indexes for searches since these SQL calls run every request.

PR #1518, opened 2015-11-02
Tweak labels and add instruction text for price filter admin page

vsaarinen:
Users have been confused about what the price filter does. People have thought that the price limit affects the minimum and maximum prices for listings, not the filter. Rename the labels and add an instruction text to clarify this:

screenshot 2015-11-02 15 13 40

Mobile view:

screenshot 2015-11-02 15 13 59


rap1ds:
Very nice! Thanks @vsaarinen ! These kind of small improvements make the product so much better.

PR #1519, opened 2015-11-02
Use asset-url instead of image-path for logos

vsaarinen:
For marketplaces that hadn't set their own logo, we were using the static path for e.g. `/assets/logos/mobile/default-highres.png`. This is the wrong path since it doesn't contain the hash of the file in the filename.

Replace `url(image-path(...))` with `asset-url(...)` to fix this.

### DIFF
--- a/app/assets/stylesheets/main.css.scss
+++ b/app/assets/stylesheets/main.css.scss
@@ -2013,7 +2013,7 @@ figure.fluidratio {
   position: relative;
 
   figure.marketplace-cover {
-    background-image: url(image-path($cover-photo-url));
+    background-image: asset-url($cover-photo-url, image);
     height: em($cover-photo-mobile-height);
     @include media(tablet) {
       @include fluid-ratio(em(1920) em($cover-photo-height), em(300) em($cover-photo-height));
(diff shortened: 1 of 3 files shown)


Gnito:
This sounds strange - I guess, this is somehow related to Rails 3 vs 4?

On Charm we have header with background image on folder:
app/assets/images/bg.header.jpg

SCSS we have:
background: url(image-path("bg.header.jpg")) no-repeat center bottom;

which produces to CSS file:
background: url("/assets/bg.header-1ce7dce2b8072ed99a270a25fb6a3b202a40d53161da4ac348d7f1e1f34ff24a.jpg") no-repeat center bottom;

rap1ds:
@Gnito Good catch! Based on your comment, I couldn't resist myself from digging this a bit deeper. Sounded too weird.

Apparently, the version of Compass that we are using overrides the Rails Asset helpers, such as `image-path`.

Here's a PR that fixes the root cause by updating Compass to a version where this is already fixed: https://github.com/sharetribe/sharetribe/pull/1521

I think that this PR should be closed.

PR #1523, opened 2015-11-03
Fix handling of FB like button for Brazil

gnomet:
Add Brazil to the list of exceptions to fix FB like button locale.

### DIFF
--- a/app/views/layouts/_facebook_sdk.haml
+++ b/app/views/layouts/_facebook_sdk.haml
@@ -1,14 +1,3 @@
-- loc = case I18n.locale
-  - when :en then "en_US"
-  - when :ca then "ca_ES"
-  - when :"es-ES" then "es_ES"
-  - when :"zh" then "zh_CN"
-  - when :"ja" then "ja_JP"
-  - when :"sv" then "sv_SE"
-  - when :"da-DK" then "da_DK"
-  - when :"pt-BR" then "pt_BR"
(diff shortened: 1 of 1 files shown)


rap1ds:
@gnomet I feel that there's some room for improvement.

We have already solved this issue for Facebook. In the `marketplace_head.haml` we set the `og:locale` metatag, which contains the Facebook compatible locale code. You should have a look how it does that, because I think the Share button locale and the metatag og:locale should be consistent.

The `marketplace_head.haml` uses the `@facebook_locale_code` variable. However, **you should not use it here.** The `marketplace_head.haml` is part of the layout, and there's no easy way to pass local variables to the layout, and thus it's acceptable to use the global `@facebook_locale_code` variable there. However, in this case, you should pass the variable from controller as a local variable.

If you have questions, I'm happy to help.

gnomet:
I would feel very tempted to just merge this PR now to get the issue fixed for the client, because I don't know when I have the time to dive into this again. Although, I do appreciate the comment of how things could be done better.

rap1ds:
@gnomet I can help!

gnomet:
@rap1ds Thanks for offering help!
Now I reserved a bit of time to look at this and indeed the existing method does the same but better, so thanks for pointing it out! :)
Also, I noticed that also this file is part of layouts, so I assumed using @facebook_locale_code directly is the way to go as passing the locale from every controller wouldn't be practical.

PR #1524, opened 2015-11-04
Cleaner URL for plans

rap1ds:
- Add a new `Admin::PlansController`
- The URL is now `www.marketplace.com/admin/plan`
- Fix null pointer if ext plan service is not in use


sktoiva:
Do we have to take care of the old urls? Have we sent them in some emails?

rap1ds:
@sktoiva The old URLs are not yet public. They are behind a feature flag.

PR #1525, opened 2015-11-05
Don't throw on invalid date

sktoiva:
Ux is still bad, but this isn't encountered too often.

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -651,9 +651,9 @@ class ListingsController < ApplicationController
         answer
       when :date_field
         answer = DateFieldValue.new
-        answer.date_value = DateTime.new(answer_value["(1i)"].to_i,
-                                         answer_value["(2i)"].to_i,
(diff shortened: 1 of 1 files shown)


app/controllers/listings_controller.rb
```diff
@@ -651,9 +651,9 @@ def custom_field_value_factory(listing_id, custom_field_id, answer_value)
         answer
       when :date_field
         answer = DateFieldValue.new
-        answer.date_value = DateTime.new(answer_value["(1i)"].to_i,
-                                         answer_value["(2i)"].to_i,
-                                         answer_value["(3i)"].to_i)
+        answer.date_value = Time.utc(answer_value["(1i)"].to_i,
+                                     answer_value["(2i)"].to_i,
+                                     answer_value["(3i)"].to_i)
```
rap1ds: Did you test, what's the value now for the hour, minute, and second? Is it 0, 0, 0? I guess it should be 23, 59, 59.

PR #1526, opened 2015-11-05
JWT Changes

rap1ds:
Use `exp` and `sub` claims in JWT tokens.

### DIFF
--- a/app/controllers/plans_controller.rb
+++ b/app/controllers/plans_controller.rb
@@ -2,7 +2,6 @@ class PlansController < ApplicationController
   skip_before_filter :verify_authenticity_token, :fetch_logged_in_user, :fetch_community, :fetch_community_membership
   skip_filter :check_email_confirmation
   before_filter :ensure_external_plan_service_in_use!
-  before_filter :do_jwt_authentication!
 
   # Request data types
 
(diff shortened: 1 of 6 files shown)


app/controllers/plans_controller.rb
```diff
@@ -91,19 +100,38 @@ def create
...
   def get_trials
-    after = Maybe(params)[:after].to_i.map { |time_int| Time.at(time_int).utc }.or_else(nil)
-    limit = Maybe(params)[:limit].to_i.or_else(MAX_TRIALS_LIMIT)
+    parse_token(params).and_then {
+      plans_api.authorize_trial_sync(params[:token])
+    }.and_then {
+      after = Maybe(params)[:after].to_i.map { |time_int| Time.at(time_int).utc }.or_else(nil)
+      limit = Maybe(params)[:limit].to_i.or_else(MAX_TRIALS_LIMIT)
+
+      logger.info("Fetching plans that are created after #{after}", nil, {after: after})
```
Gnito: Why not log limit?

Gnito:
Looks good too. (Although, both of these are so long that the amount of code changes kind of blinds detailed review process.)

app/controllers/plans_controller.rb
rap1ds: For some reason I thought that it's not that important info, but I can add it.

PR #1527, opened 2015-11-05
Log fb_name_missing sign up data

Gnito:
Or should I just redirect people to sign_up_path ?


app/controllers/sessions_controller.rb
```diff
@@ -124,6 +124,10 @@ def facebook
         redirect_to sign_up_path and return
       end
 
+      if data.username.blank? && data.family_name.blank? && data.given_name.blank?
+        FacebookSignupFails.new(auth_data: data, community_id: @current_community.id)
```
Gnito: create...

PR #1528, opened 2015-11-05
Treat closed marketplace (hold plans and expired non-trial plans) as deleted

rap1ds:
- [X] Add `closed` status to plan entity
- [X] Redirect `closed` plans to not found
- [x] Make sure that the data is correct


app/services/plan_service/api/no_plans.rb
```diff
@@ -26,6 +26,8 @@ def get_current(community_id:)
         community_id: community_id,
         plan_level: PlanService::Levels::OS,
         expires_at: nil,
+        closed: false,
+        deleted: false,
```
rap1ds: Fix this

PR #1531, opened 2015-11-06
Add closed state to plan

rap1ds:
- [X] Reviewed
- [X] Tested on staging

### DIFF
--- a/app/services/plan_service/api/plans.rb
+++ b/app/services/plan_service/api/plans.rb
@@ -111,9 +111,16 @@ module PlanService::API
       )
     end
 
+    # Return true, if plan is closed, i.e.
+    # - Hold plan
+    # - Expired non-trial plan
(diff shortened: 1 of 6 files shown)


app/services/plan_service/api/plans.rb
```diff
@@ -53,10 +53,7 @@ def get_current(community_id:)
 
     def expired?(community_id:)
```
sktoiva: There isn't a need for publicly query `closed?` for a community?

sktoiva:
Added one comment, otherwise good to go.

app/services/plan_service/api/plans.rb
rap1ds: Good point. Yeah, apparently not, and apparently there's actually no need for `expired?` method.

PR #1540, opened 2015-11-10
Close any open tags in marketplace description when truncating

vsaarinen:
Fixes the bug where an HTML tag might not be closed when the description is truncated. This broke the front page layout. Uses the [truncate_html](https://github.com/hgmnz/truncate_html) gem that is already installed.


rap1ds:
:+1:

PR #1542, opened 2015-11-11
Paypal improvement fixes

vsaarinen:
1. Fix broken "Change PayPal account" for users on the second step.
2. Add padding above commission settings
3. Tweak copy for admins about having to connect their PayPal account twice.
4. Move PayPal account connected section to its own partial


rap1ds:
:+1:

PR #1543, opened 2015-11-11
Optimize categories sort

sktoiva:


app/controllers/admin/categories_controller.rb
```diff
@@ -57,18 +57,22 @@ def update
   end
 
   def order
-    sort_priorities = params[:order].each_with_index.map do |category_id, index|
+    sort_priorities = params[:order]
+                      .reject { |o| !o.match /[0-9]+/} #Guard against sql injection
```
mporkola: this probably isn't needed, since the ID:s are matched to actual category ID:s later on, and not entered into any SQL query before that

app/controllers/admin/categories_controller.rb
```diff
@@ -106,6 +110,29 @@ def destroy_and_move
...
+  #                            .
+  #                       END
+  #  WHERE id IN(1, 2, ...);
+  ##
+  def order_sql(sort_priorities)
+    base = "UPDATE categories
+              SET sort_priority = CASE id\n"
+
+    update_statements = sort_priorities.reduce(base) do |sql, (cat_id, priority)|
+      sql += "WHEN #{cat_id} THEN #{priority}\n"
```
rap1ds: + instead of +=

app/controllers/admin/categories_controller.rb
```diff
@@ -106,6 +110,29 @@ def destroy_and_move
...
+    end
+
+    update_statements += "END\n WHERE id IN (#{sort_priorities.keys.join(",")});"
```
rap1ds: Same here, to avoid mutating the value, use `+` instead of `+=`

app/controllers/admin/categories_controller.rb
rap1ds: What about `.map { |o| o.to_i }` ?

rap1ds: Or shorter `params[:order].map(&:to_i) ... `

rap1ds:
Would it make sense to move this code to `https://github.com/sharetribe/sharetribe/blob/master/app/services/listing_service/api/categories.rb` ?

sktoiva:
Maybe it would. I don't have time to it now.

app/controllers/admin/categories_controller.rb
sktoiva: No.

```ruby
"abcd".to_i 
#=> 0
```

PR #1545, opened 2015-11-12
Include translations

sktoiva:
Include translations to avoid multiple queries from views to db when there are
lots of categories and subcategories.


mporkola:
nice!

PR #1548, opened 2015-11-16
Clean up paypal admin path

rap1ds:
- [X] Old path: /admin/communities/<community_id>/paypal_preferences
- [X] New path: /admin/paypal_preferences
- [X] Redirect from old path to new path
- [X] Use the simpler and cleaner path definitions and ditch the restful route definition

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -115,6 +115,11 @@ Kassi::Application.routes.draw do
 
     namespace :admin do
 
+      get  "/paypal_preferences"                      => "paypal_preferences#index"
+      post "/paypal_preferences/preferences_update"   => "paypal_preferences#preferences_update"
+      get  "/paypal_preferences/account_create"       => "paypal_preferences#account_create"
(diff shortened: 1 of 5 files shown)


mporkola:
:clap:

config/routes.rb
```diff
@@ -115,6 +115,11 @@
 
     namespace :admin do
 
+      get  "/paypal_preferences"                      => "paypal_preferences#index"
+      post "/paypal_preferences/preferences_update"   => "paypal_preferences#preferences_update"
```
lautis: Rails protip: you can use `resource` instead of `resources` if the resource is a singleton. Using `resource :paypal_preferences, only: [:update, :show]` would create "resourceful" routes for both get and update without `:id` part. Some refactoring would be needed to use `PUT` instead of `POST` for the update action.

http://stackoverflow.com/questions/9194767/difference-between-resource-and-resources-methods

rap1ds: @lautis Thanks for the tip! :)

PR #1550, opened 2015-11-16
Mechanism to add country specific PayPal instruction texts and links

rap1ds:
- Clean up the old code regarding country specific links
- Add Australian instruction texts and links

### DIFF
--- a/app/view_utils/paypal_country_helper.rb
+++ b/app/view_utils/paypal_country_helper.rb
@@ -1,0 +1,74 @@
+
+
+# Specify all country specific PayPal instruction texts / link in this module
+#
+
+module PaypalCountryHelper
+
+  FEE_URL = {
+    # List all the contries that have the new fee page available
+    "us"      => "https://www.paypal.com/us/webapps/mpp/paypal-fees",
+    "de"      => "https://www.paypal.com/de/webapps/mpp/paypal-fees",
+    "br"      => "https://www.paypal.com/br/webapps/mpp/paypal-fees",
+    "fr"      => "https://www.paypal.com/fr/webapps/mpp/paypal-fees",
+    "au"      => "https://www.paypal.com/au/webapps/mpp/paypal-seller-fees",
+
+    "default" => "https://www.paypal.com/cgi-bin/marketingweb?cmd=_display-xborder-fees-outside",
+  }
+
+  POPUP_URL = {
+    # List all the contries that have the popup URL available
+    "us"      => "https://www.paypal.com/us/webapps/mpp/paypal-popup",
+    "de"      => "https://www.paypal.com/de/webapps/mpp/paypal-popup",
+    "fr"      => "https://www.paypal.com/fr/webapps/mpp/paypal-popup",
+    "au"      => "https://www.paypal.com/au/webapps/mpp/paypal-popup",
+
+    # List all the countries that should use the home URL, because popup is not available
+    # (and default English popup is not good)
+    "br"      => "https://www.paypal.com/br/webapps/mpp/home",
+
+    "default" => "https://www.paypal.com/webapps/mpp/paypal-popup",
+  }
+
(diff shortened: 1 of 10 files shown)


app/view_utils/paypal_country_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+    "au"      => "https://www.paypal.com/au/webapps/mpp/paypal-popup",
+
+    # List all the countries that should use the home URL, because popup is not available
+    # (and default English popup is not good)
+    "br"      => "https://www.paypal.com/br/webapps/mpp/home",
+  }
+
+  POPUP_URL.default => "https://www.paypal.com/webapps/mpp/paypal-popup",
+
+  CREATE_ACCOUNT_URL = {
```
vsaarinen: Missing `.default` for this.

app/view_utils/paypal_country_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+    "au"      => "https://www.paypal.com/au/webapps/mpp/account-selection",
+
+    "default" => "https://www.paypal.com/%{country_code}/webapps/mpp/home",
+  }
+
+  RECEIVE_FUNDS_INFO_LABEL_TR_KEY = {
```
vsaarinen: Missing `.default` for this.

app/view_utils/paypal_country_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+    find_or_default(CREATE_ACCOUNT_URL, country_code) % {country_code: country_code}
+  end
+
+  def receive_funds_info_label_tr_key(country_code)
+    find_or_default(RECEIVE_FUNDS_INFO_LABEL_TR_KEY, country_code)
+  end
+
+  # private
+
+  def find_or_default(hash, key)
```
vsaarinen: Once you set defaults for all hashes, couldn't this be simplified to:
```
hash[k.to_s.downcase]
```
? At which point having this separate function might not make sense anymore...

app/view_utils/paypal_country_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+    "de"      => "https://www.paypal.com/de/webapps/mpp/paypal-popup",
+    "fr"      => "https://www.paypal.com/fr/webapps/mpp/paypal-popup",
+    "au"      => "https://www.paypal.com/au/webapps/mpp/paypal-popup",
+
+    # List all the countries that should use the home URL, because popup is not available
+    # (and default English popup is not good)
+    "br"      => "https://www.paypal.com/br/webapps/mpp/home",
+  }
+
+  POPUP_URL.default => "https://www.paypal.com/webapps/mpp/paypal-popup",
```
vsaarinen: Should be an assignment, not `=>`.

vsaarinen: Also: unnecessary comma at the end of the line. Same thing for FEE_URL.

app/view_utils/paypal_country_helper.rb
rap1ds: Otherwise yes, but it throws a null pointer if `k` is null. The expected behaviour is to show the default link/instruction if `k` (i.e. country) is not specified.

rap1ds: Oh, well, actually it doesn't throw, since `nil.to_s` is an empty string. Anyway, I'd rather make it more explicit than that.

vsaarinen: I'm a bigger fan of this simpler style for the above functions:
```ruby
def receive_funds_info_label_tr_key(country_code)
  RECEIVE_FUNDS_INFO_LABEL_TR_KEY[country_code.to_s.downcase]
end
```
But that's, of course, a personal preference.

vsaarinen:
There's one comment, but you're free to ignore it and :shipit:

PR #1551, opened 2015-11-16
Remove Rack::Timeout

vsaarinen:
Rely on just Heroku timeouts instead of using Rack::Timeout.


ovan:
Looks clear! Easy to pull this back in if needed.

PR #1552, opened 2015-11-16
external plan service: feature flag for new trials

Gnito:


app/services/marketplace_service/api/marketplaces.rb
```diff
@@ -73,6 +73,9 @@ def create(params)
       processes = Helper.create_processes!(community.id, payment_process)
       shape = Helper.create_listing_shape!(community, p[:marketplace_type], payment_process)
 
+      # TODO: remove this after external_plan_page is available to everyone.
+      FeatureFlagService::API::Api.features.enable(community_id: community.id, features: [:new_plan_page])
```
rap1ds: This exactly right thing to do. However, we have usually done that on the controller level, since if "forbidden" for a service to call another services.

Here's an example: https://github.com/sharetribe/sharetribe/commit/6fd9adb778e313f5da1802311947eaddf5727c93

Gnito:
Travis doesn't want to play with us, but local rspec goes through nicely. I'll merge

PR #1553, opened 2015-11-17
Send domain only if it's in use

rap1ds:
Steps to reproduce:

Precondition: Community with `communities.domain = www.google.com` and `communities.use_domain = false`

1. Go to Admin
2. Click Your plan
3. Click Back to my marketplace

Expected: Goes back to marketplace

Actual: Goes to www.google.com


mporkola:
charm seems to be ok with this, looks good!

PR #1554, opened 2015-11-18
Fix typo

cqpx:
stringify_booking_data -> stringify_booking_date


rap1ds:
Nice catch! Thanks for the fix @cqpx !

PR #1556, opened 2015-11-19
Don't overwrite metadata passed into transition_to

valdis:
Don't overwrite metadata passed into transition_to


rap1ds:
Thanks for the fix!

PR #1572, opened 2015-11-30
Updage Fog gem - Fixes #1569

### LINKED ISSUE #1569: Update fog gem
> Digest::Digest is deprecated; use Digest

https://github.com/fog/fog/pull/2473

rap1ds:
Steps to reproduce: 

1. Run `rails console`

Expected: Console opens without warnings

Actual: `Digest::Digest is deprecated; use Digest`

Fix: run `bundle update fog`

Fixes #1569

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -20,6 +20,7 @@ GIT
 GEM
   remote: http://rubygems.org/
   specs:
+    CFPropertyList (2.3.2)
     action_mailer_cache_delivery (0.3.4)
       actionmailer (~> 3.0)
     actionmailer (3.2.21)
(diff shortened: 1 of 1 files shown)


mporkola:
Shouldn't the Gemfile also be changed?

rap1ds:
@mporkola Good point. That's not mandatory (since we don't have a direct dependency to `fog`, instead it's `asset_sync`'s dependency), but now when I looked at the Asset Sync [README](https://github.com/AssetSync/asset_sync#optimized-fog-loading), I noticed that `fog` loading can be optimized by adding it to Gemfile.

mporkola:
Ok, let's merge then!

PR #1573, opened 2015-11-30
Wrap lines also in browsers without hyphenation.

mporkola:
Only Firefox was able to wrap lines, since other browsers don't
seem to have hyphenation support. Not tested on IE, but according to MDN it should behave ok with this.

Before:
screenshot 2015-11-30 11 52 54

After:
screenshot 2015-11-30 11 53 06


rap1ds:
:+1:

PR #1574, opened 2015-11-30
Logging improvements for listing image uploader

rap1ds:
- [X] Don't use `.delay`, instead create a separate job `DownloadListingImageJob`. This allows us to set the correct priority.
- [X] Catch and log possible errors
- [X] Add logging here and there

### DIFF
--- a/app/controllers/listing_images_controller.rb
+++ b/app/controllers/listing_images_controller.rb
@@ -23,7 +23,7 @@ class ListingImagesController < ApplicationController
     url = escape_s3_url(params[:path], params[:filename])
 
     if !url.present?
-      logger.warn("No image URL provided", :no_image_url_provided, params)
+      logger.info("No image URL provided", :no_image_url_provided, params)
       render json: {:errors => "No image URL provided"}, status: 400, content_type: 'text/plain'
     end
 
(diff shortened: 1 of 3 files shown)


app/controllers/listing_images_controller.rb
```diff
@@ -23,6 +23,7 @@ def add_from_url
     url = escape_s3_url(params[:path], params[:filename])
 
     if !url.present?
+      logger.warn("No image URL provided", :no_image_url_provided, params)
```
ovan: I feel this should be either info level (Everything is ok, client messed up but we noticed and fixed it) or error (An error happened here but was caught and handled). Warn would then be something like "this field really should have a value but it doesn't, I'm gonna save it anyway but later on you will see broken images as result"). So Warn is a warning that something is not right and that can potentially lead to problems later.

app/controllers/listing_images_controller.rb
```diff
@@ -75,11 +72,16 @@ def new_image(params, url)
     listing_image.image_downloaded = if url.present? then false else true end
 
     if listing_image.save
-      unless listing_image.image_downloaded
-        listing_image.delay.download_from_url(url)
+      if !listing_image.image_downloaded
+        logger.info("Asynchronously downloading image", :start_async_image_download, listing_image_id: listing_image.id, url: url, params: params)
+        Delayed::Job.enqueue(DownloadListingImageJob.new(listing_image.id, url), priority: 1)
+      else
+        logger.info("Listing image downloaded already", :image_downloaded_already, listing_image_id: listing_image.id, params: params)
```
ovan: :image_downloaded_already => :image_already_downloaded

app/controllers/listing_images_controller.rb
```diff
@@ -75,11 +72,16 @@ def new_image(params, url)
...
       end
+
       render json: ListingImageJSAdapter.new(listing_image).to_json, status: 202, content_type: 'text/plain' # Browsers without XHR fileupload support do not support other dataTypes than text
     else
+      logger.error("Saving listing image failed", :saving_listing_image_failed, params: params)
```
ovan: Should this log row include something about the errors that prevented image being saved?

PR #1575, opened 2015-12-01
SphinxAdapter no more returns deleted listings without query

mporkola:
`SphinxAdapter` disregarded the `deleted` flag when doing a straight DB query to fetch listings, causing deleted (but not closed) listings to appear i.e. on homepages

### DIFF
--- a/app/models/listing.rb
+++ b/app/models/listing.rb
@@ -49,8 +49,8 @@
 #
 # Indexes
 #
-#  homepage_query                      (community_id,open,sort_date)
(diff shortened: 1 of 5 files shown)


rap1ds:
Make sure that the `delete` is indexed and keep an eye on frontpage performance after deploying.

mporkola:
Tested performance on frontpage listing query.

Avg. query times
before: 0.0022
after, with no index changes: 0.0030
after, with index changes: 0.0022

PR #1576, opened 2015-12-02
Fix empty title error

rap1ds:
If the listing title is empty, the homepage will break.

This PR fixes two errors:

* The backend has a validation that requires the title length to be 2-60. However, the front-end validation requires title to be 1-60 characters. This PR changes the front-end validation to 2-60.
* ListingIndexService allows empty titles. That will ensure that the homepage doesn't break.


mporkola:
Good catch, nice fix

PR #1585, opened 2015-12-04
Don't break the inbox if message content is empty

rap1ds:
In some rare edge cases, the message content may be null and it will break the inbox.

This PR fixes the bug by allowing null content. In the inbox, we show a message "Message content not available", if the content is empty.


sktoiva:
The tests are broken badly. The title isn't shown for some reason now?

PR #1588, opened 2015-12-07
[HIBERNATING] Cache homepage listings block

mporkola:
This PR adds caching to front page listings block. Previously, only the individual listing cards (red outline below) were cached, but now the complete blue block is cached.

screenshot 2015-12-07 14 00 49

The cache for the listings block (actually [cache key](https://signalvnoise.com/posts/3113-how-key-based-cache-expiration-works)) is expired when any of these happen:
 1. `community` is updated
 2. `listing` is updated
 3. `author` of a listing on frontpage is updated

In practise only `community.id` and `community.updated_at` are checked. Updating a listing touches `community.updated_at`, and updating a listing author touches `listing.updated_at` of all their listings, which causes also community to be touched. If author had a `belongs_to` relationship with community, this could be simplified.

Performance results on Testing were inconclusive, but DB queries were reduced. Let's see how this works in production, this PR can be reverted with no harm caused.

---
This change is [Reviewable](https://reviewable.io/reviews/sharetribe/sharetribe/1588)

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -54,7 +54,12 @@ class HomepageController < ApplicationController
         raise ArgumentError.new("Unknown view_type #{@view_type}")
       end
 
-    search_result = find_listings(params, per_page, compact_filter_params, includes.to_set)
+    search_param_digest = Digest::MD5.hexdigest([params, per_page, compact_filter_params, includes].join)
+    view_param_digest = Digest::MD5.hexdigest([params, per_page, compact_filter_params, includes, I18n.locale].join)
+
+    search_result = Rails.cache.fetch(["index_listings", @current_community.cache_key, search_param_digest]) do
+      find_listings(params, per_page, compact_filter_params, includes.to_set)
+    end
 
     shape_name_map = all_shapes.map { |s| [s[:id], s[:name]]}.to_h
 
@@ -74,6 +79,7 @@ class HomepageController < ApplicationController
       search_result.on_success { |listings|
         @listings = listings
         render locals: {
+                 param_digest: view_param_digest,
                  shapes: all_shapes,
                  show_price_filter: show_price_filter,
                  selected_shape: selected_shape,
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -113,11 +113,6 @@ class Person < ActiveRecord::Base
 
   has_and_belongs_to_many :followed_listings, :class_name => "Listing", :join_table => "listing_followers"
 
-  after_save do
-    # This is required to invalidate the fragment caches for listings on front page, since the author is shown there.
-    self.listings.all.map(&:touch)
-  end
-
   def to_param
     username
   end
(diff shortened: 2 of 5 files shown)


app/models/person.rb
```diff
@@ -113,6 +113,11 @@ class Person < ActiveRecord::Base
 
   has_and_belongs_to_many :followed_listings, :class_name => "Listing", :join_table => "listing_followers"
 
+  after_save do
+    # This is required to invalidate the fragment caches for listings on front page, since the author is shown there.
+    self.listings.all.map(&:touch)
+  end
+
```
mporkola: This is the problematic part, but shouldn't cause any issues unless people with a lot of listings update their profile a lot. With `belongs_to :community` this could be skipped, and the update chain would be `listing` -> `person` -> `community`.

app/views/homepage/index.haml
```diff
@@ -115,26 +115,27 @@
...
-              = render :partial => "list_item", :collection => @listings, :as => :listing, locals:{shape_name_map: shape_name_map, testimonials_in_use: testimonials_in_use}
-          .home-loading-more
-            = will_paginate(@listings)
-            - item_container = if @view_type.eql?("grid") then '.home-fluid-thumbnail-grid' else '.home-listings' end
-            = pageless(@listings.total_pages, item_container, request.fullpath, t('.loading_more_content'))
-      - else
-        .home-no-listings
-          - if params[:q] || params[:category] || params[:share_type] # Some filter in use
-            = t(".no_listings_with_your_search_criteria")
+    - cache(["index_haml", @current_community.cache_key, param_digest]) do
```
mporkola: only this line was added, see https://github.com/sharetribe/sharetribe/pull/1588/files?w=1

app/controllers/homepage_controller.rb
```diff
@@ -54,7 +54,12 @@ def index
         raise ArgumentError.new("Unknown view_type #{@view_type}")
       end
 
-    search_result = find_listings(params, per_page, compact_filter_params, includes.to_set)
+    search_param_digest = Digest::MD5.hexdigest([params, per_page, compact_filter_params, includes, request.xhr?].join)
+    view_param_digest = Digest::MD5.hexdigest([params, per_page, compact_filter_params, includes, request.xhr?, I18n.locale].join)
+
+    search_result = Rails.cache.fetch(["index_listings", @current_community.cache_key, search_param_digest]) do
```
rap1ds: I think this part could be done inside the SphinxAdapter (inside the ListingIndexService).

app/models/person.rb
rap1ds: This is indeed quite problematic. I'd suggest looking for these two options:

1. Can we do this in Controller layer? The model hooks are quite problematic, and some people even consider them as anti-pattern, so adding manual invalidation in the controller would solve that.
2. Listing.where(author_id: <author_id).update_all(updated_at: Time.now) can do the update in one query.

mporkola:
We're using `update_all` in quite many places to update listings and people, so `updated_at` column might lie to us. They need to be updated to manually update the timestamp, like this:
```ruby
#before
listings.where(:author_id => author.id).update_all(:open => false)

#after
listings.where(:author_id => author.id).update_all(:open => false, :updated_at => Time.now)
```
I'll keep the branch open, I believe it's a good starting point

mporkola:
We'll do this in the new React implementation instead.

PR #1589, opened 2015-12-08
Improved email notification content a bit.

thomasmalbaux:


sktoiva:
Why isn't this completely translated? Because it's sent only to admins?

thomasmalbaux:
Unfortunately no idea. @rap1ds suggested to change this file directly instead of moving to a translation approach as there might be underlying reasons for the current setup but we didn't investigate.
If you have some insights about this, happy to learn more!

rap1ds:
How to move this forward? @sktoiva was that only a question or also a suggestion to translate the message?

sktoiva:
It was only a question. If you have already done the pondering regarding the issue just move forward.

PR #1591, opened 2015-12-08
Listing image error handling

rap1ds:
### Why?

Every now and then we get support requests that complain that the **listing image processing has got stuck**. The reason is often that people are uploading broken image files or files that are now JPG, GIF or PNG, but they have just renamed the files and changed the extension. Solving these issues add support load.

### What?

Makes image processing more error resilient. Fixes two bug, here are steps to reproduce:

**Not showing error message**:

1. Go to new listings page
1. Upload a broken image file (e.g. take a .html file and rename it to .jpg)
1. Wait for awhile

Expected: You see an error saying that the image processing failed

Actual: Nothing. The "Processing, this takes only a second" text is there for ever.

**Listing images get stuck**:

1. Go to new listings page
1. Upload a broken image file (e.g. take a .html file and rename it to .jpg)
1. Wait until the image is uploaded and you see text "Processing, this takes only a second"
1. Save the listing

Expected: You arrive to listing page. You see a text that says that the image is being processed. Wait awhile and refresh and you see the listing image

Actual: Expected: You arrive to listing page. You see a text that says that the image is being processed. No matter how long you wait, you see the text saying that the listing image is being processed.

### How?

- [x] Raraise the exception in `on_error` handler in `DownloadListingImageJob`. This will make the Delayed Job to retry to job 3 times
- [x] Add a new boolean column `errored` to `listing_images`. If this column is true, it's a signal that the listing image processing failed and the listing image should be ignored.
- [x] Change the listing image status poller (during the image upload/process on new listing page) so that it shows error message if the processing fails.

### Screenshots

screen shot 2015-12-09 at 10 05 56

### DIFF
--- a/app/assets/javascripts/image_uploader.js
+++ b/app/assets/javascripts/image_uploader.js
@@ -288,10 +288,6 @@ window.ST.imageUploader = function(listings, opts) {
     return value.ready && !value.errored;
   });
 
-  var errorDuringUpload = imageUploaded.filter(function(value) {
-    return value.errored;
-  });
-
   var newPreviewRendered = successfullyUploaded.map(function(listing) {
     return renderPreview(listing);
   });
--- a/app/assets/javascripts/image_uploader.js
+++ b/app/assets/javascripts/image_uploader.js
@@ -284,7 +284,15 @@ window.ST.imageUploader = function(listings, opts) {
   var status = Bacon.combineTemplate(
     {loading: numberOfLoadingImages, processing: numberOfProcessingImages });
 
-  var newPreviewRendered = imageUploaded.map(function(listing) {
+  var successfullyUploaded = imageUploaded.filter(function(value) {
+    return value.ready && !value.errored;
+  });
+
+  var errorDuringUpload = imageUploaded.filter(function(value) {
+    return value.errored;
+  });
+
+  var newPreviewRendered = successfullyUploaded.map(function(listing) {
     return renderPreview(listing);
   });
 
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/image_uploader.js
```diff
@@ -284,7 +284,15 @@ window.ST.imageUploader = function(listings, opts) {
   var status = Bacon.combineTemplate(
     {loading: numberOfLoadingImages, processing: numberOfProcessingImages });
 
-  var newPreviewRendered = imageUploaded.map(function(listing) {
+  var successfullyUploaded = imageUploaded.filter(function(value) {
+    return value.ready && !value.errored;
+  });
+
+  var errorDuringUpload = imageUploaded.filter(function(value) {
```
rap1ds: Not in use

app/jobs/download_listing_image_job.rb
```diff
@@ -13,44 +13,60 @@ def perform
...
   end
 
   private
 
   def logger
-    @logger ||= SharetribeLogger.new(:download_listing_image_job)
+    if @logger
+      @logger
+    else
+      @logger = SharetribeLogger.new(:download_listing_image_job, logger_metadata.keys).tap { |logger|
```
rap1ds: Simplify with `||=`

app/jobs/download_listing_image_job.rb
```diff
@@ -13,44 +13,60 @@ def perform
...
-      logger.error(error_msg, :listing_image_download_failed, data)
+      logger.error(error_msg, :listing_image_download_failed_permanently, data)
     }
   end
 
   private
 
   def logger
-    @logger ||= SharetribeLogger.new(:download_listing_image_job)
+    if @logger
```
sktoiva: Isn't this equivalent of ```@logger ||= ...```

app/jobs/download_listing_image_job.rb
```diff
@@ -13,44 +13,56 @@ def perform
...
-      rescue StandardError => e
-        Result::Error.new(e)
-      end
+      raise ex # Reraise the exception to mark the delayed job failed
+    }
+  end
 
+  def failure
+    listing_image.on_success { |listing_image|
+      listing_image.update_column(:errored, true) # Have to use `update_column` here because `update_attribute` runs the before save hooks
```
ovan: I was wondering if instead of errored just being a boolean it could contain the error reason as symbol but I'm not sure if the exception data is available here anymore?

PR #1595, opened 2015-12-10
Reserve marketplace-academy ident

sktoiva:
This is used as a domain for Academy Wordpress installation running in
marketplace-academy.sharetribe.com


rap1ds:
:+1:

PR #1596, opened 2015-12-11
Change "en" time format from dd/mm/yyyy to mm/dd/yyyy

rap1ds:
The "en" locale is currently en-US, but the date format is dd/mm/yyyy and that's wrong. This PR changes the "en" date formats from dd/mm/yyyy to mm/dd/yyyy.

We have the en-GB which can be used for English with dd/mm/yyyy format.

Todo:

- [X] Change "en" timeformat
- [x] Go to WebTranslateIt and add the time formats for en-GB

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2120,10 +2120,9 @@ en:
     format: mm/dd/yyyy
   time:
     formats:
-      short: "%e %b %Y at %H:%M"
-      shorter: "%e %b at %H:%M"
-      short_date: "%e %b %Y"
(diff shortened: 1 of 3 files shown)


mporkola:
The breaking test seems to be actually relevant:

>Then I should see list of users with the following details:  # features/step_definitions/admin_manage_members_steps.rb:33
>      expected: "Mar 1, 2014"
>          got: "1 Mar 2014" (using ==) (RSpec::Expectations::ExpectationNotMetError)

PR #1600, opened 2015-12-14
Update gems

rap1ds:
I went through the list of Gems we are using with the tool http://www.ready4rails.net/

This PR upgrade some gems that need it.


ovan:
@rap1ds How did you arrive at this list of gems to update? What kind of testing have you done in addition to CI, so how confident are you that this won't break stuff?

ovan:
Sorry, should have read better the PR description so forget the first question.

PR #1602, opened 2015-12-15
Remove gems

rap1ds:
Remove unused gems that are not Rails 4 compatible

- [X] Remove postmark
- [X] Remove recaptcha
- [X] Remove validate_timeliness
- [x] Test signup
- [x] Test booking

Reviewing tip: Commit by commit


rap1ds:
D'oh, removed all the `recaptcha` related code, but didn't remove the Gem itself from the Gemfile. I'll do that.

app/utils/date_on_or_after_validator.rb
```diff
@@ -0,0 +1,19 @@
+class DateOnOrAfterValidator < ActiveModel::Validator
```
ovan: This guy feels like a bit too much hard coded assumptions to me. Could it e.g. be a more general validation that validates some given date field is on or after another given datefield? Then the record fields used would be parameterized instead of assuming start_on and end_on.

PR #1603, opened 2015-12-15
Update delayed_job gem

rap1ds:
Update delayed_job gem. Version 4.0.4 broke test, so that why I updated to 4.0.3

Testing: This is pretty well covered by the tests. No manual testing done.


sktoiva:
What about delayed_paperclip? Does it depend on this?

Otherwise :+1: . Did you check, does this fix the code loading issue to workers? As new versions of Delayed Job should take care of that.

rap1ds:
delayed_paperclip has quite loose dependency versioning, it depends on delayed_job version > 0. So this update is compatible with delayed_paperclip.

PR #1604, opened 2015-12-16
Update devise

rap1ds:
Update devise to version that is Rails 4 compatible.

Minor code change: `build_resource` takes the sign up parameters as arguments, which is very nice, less magic.

Testing: This is pretty well covered by the cucumber tests. In addition, I tested the signup process locally.


sktoiva:
Did you check Facebook connect? Because omniauth_facebook integrates somehow with devise.

rap1ds:
Tested locally, seems to work.

PR #1607, opened 2015-12-16
Prepare rails 4 cont

rap1ds:
Prepare for Rails 4 update. See commit messages for more detailed info.

Testing:

- [x] Test normal preauthorization flow
- [x] Test preauthorization with booking
- [x] Test free flow
- [x] Test signup/login
- [x] Test Mercury editor


ovan:
@rap1ds cucumbers cover poorly the transaction paths so I suggest you focus some manual efforts into paypal process, with and without shipping, etc.

ovan:
Olright, some comments and questions about match -> get conversions

PR #1609, opened 2015-12-17
Fix atom feed and homepage to use same array format for listing shapes

gnomet:
Earlier homepage used listing_shape_id as fixnum and atom feed could
with some filters use multiple listing_shape ids, now the parameters
is always calle listing_shape_id and is always an array.

This fixes the bug of offer/request filters not working with atom feed

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -134,7 +134,7 @@ class HomepageController < ApplicationController
     search = {
       # Add listing_id
       categories: filter_params[:categories],
-      listing_shape_id: Maybe(filter_params)[:listing_shape].or_else(nil),
+      listing_shape_id: filter_params[:listing_shape] ? [filter_params[:listing_shape]] : nil,
       price_cents: filter_params[:price_cents],
       keywords: filter_params[:search],
       fields: checkboxes.concat(dropdowns).concat(numbers),
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -134,7 +134,7 @@ class HomepageController < ApplicationController
     search = {
       # Add listing_id
       categories: filter_params[:categories],
-      listing_shape_id: filter_params[:listing_shape] ? [filter_params[:listing_shape]] : nil,
+      listing_shape_id: Array(filter_params[:listing_shape]),
       price_cents: filter_params[:price_cents],
       keywords: filter_params[:search],
       fields: checkboxes.concat(dropdowns).concat(numbers),
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -134,7 +134,7 @@ class HomepageController < ApplicationController
     search = {
       # Add listing_id
       categories: filter_params[:categories],
-      listing_shape_id: Array(filter_params[:listing_shape]),
+      listing_shape_ids: Array(filter_params[:listing_shape]),
       price_cents: filter_params[:price_cents],
       keywords: filter_params[:search],
       fields: checkboxes.concat(dropdowns).concat(numbers),
--- a/app/services/listing_index_service/data_types.rb
+++ b/app/services/listing_index_service/data_types.rb
@@ -17,7 +17,7 @@ module ListingIndexService::DataTypes
     [:per_page, :to_integer, :mandatory, gte: 1],
     [:keywords, :string, :optional],
     [:categories, :array, :optional],
-    [:listing_shape_id, :array, :optional],
+    [:listing_shape_ids, :array, :optional],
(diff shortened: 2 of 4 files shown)


app/services/listing_index_service/data_types.rb
```diff
@@ -17,7 +17,7 @@ module ListingIndexService::DataTypes
     [:per_page, :to_integer, :mandatory, gte: 1],
     [:keywords, :string, :optional],
     [:categories, :array, :optional],
-    [:listing_shape_id, :fixnum, :optional],
+    [:listing_shape_id, :array, :optional],
```
rap1ds: I guess this should be renamed to `listing_shape_ids`, since it's an array of multiple ids.

app/controllers/homepage_controller.rb
```diff
@@ -134,7 +134,7 @@ def find_listings(params, listings_per_page, filter_params, includes)
     search = {
       # Add listing_id
       categories: filter_params[:categories],
-      listing_shape_id: Maybe(filter_params)[:listing_shape].or_else(nil),
+      listing_shape_id: filter_params[:listing_shape] ? [filter_params[:listing_shape]] : nil,
```
rap1ds: I think that instead of `Array` or `nil`, the code would be simpler if there were Array of one and Array or none. That way all the code that uses this can rely on the fact that it's an array, and they don't need to build extra `nil` checks.

Btw, `Array` method may be useful for this. It returns empty array, if you pass it a `nil`:

```ruby
2.1.1 :012 > Array(nil)
 => []
2.1.1 :013 > Array(123)
 => [123]
2.1.1 :014 >
```

So the code would be simply:

```ruby
listing_shape_ids: Array(filter_params[:listing_shape]),
```

app/services/listing_index_service/data_types.rb
gnomet: @rap1ds  I thought of renaming to plural too, but with quick test passing that to search engine didn’t seem to work. I guess as sphinx index has listing_shape id it matches there. And changing it to plural wouldn’t make sense in the index as one listing only has one id. And if the param would be listing_shapre_id in the app it would need to be converted to listing_shape_id before passing to sphinx. 

So with my current understanding would be simpler to call it listing_shape_id all the time..

app/controllers/homepage_controller.rb
gnomet: Excellent suggestion! Thanks for tipping the Array()

app/services/listing_index_service/data_types.rb
ovan: What happens now, how does the conversion work at sphinx_adapter? This changes the data type from number to a list of numbers so someone downstream must convert it or else it wouldn't work with sphinx because the sphinx content hasn't changed? Does it work with more than one number in the array?

rap1ds: @ovan To my understanding sphinx can handle either a single number or an array of numbers.

@gnomet In that case you should rename it to `listing_shape_ids` in the ListingIndexService layer and then rename it back to `listing_shape_id` in SphinxAdapter. Remember that Sphinx is only an implementation detail that may be replaced with some other Adapter, e.g. ElasticSearchAdapter.

gnomet: @rap1ds yes, that's a good point so I'll do the renaming in sphinx spesific code.

app/services/listing_index_service/search/sphinx_adapter.rb
```diff
@@ -19,6 +19,12 @@ def search(community_id:, search:, includes:)
         Result::Error.new(ArgumentError.new("Both DB query and search engine would be needed to fulfill the search"))
       end
 
+      # rename listing_shape_ids to singular so that Sphinx understands it
+      if search[:listing_shape_ids] && search[:listing_shape_id].nil?
+        search[:listing_shape_id] = search[:listing_shape_ids]
+        search.delete(:listing_shape_ids)
+      end
+
```
rap1ds: More elegant way to rename keys:

```ruby
search = HashUtils.rename_keys({:listing_shape_ids => :listing_shape_id}, search)
```

(note about style: I usually prefer the new Hash syntax `{:listing_shape_ids => :listing_shape_id}`, put in some cases where the Hash describes a mapping **from** some value **to** another, then I prefer using the `=>` format because the arrow strongly points the direction of the mapping)

You should also move this before line 18 (`if ... needs_search?`). The `needs_search?` methods uses `:listing_shape_id`

gnomet: @rap1ds Thanks for good advice! Didn't know about us having the HashUtils.
And well spotted with needs_search?
I'll deploy to production when tests green

PR #1612, opened 2015-12-18
Prepare rails 4 cont 3

rap1ds:
* Rename partial with dash in the name
* Adds IDs to routes that worked in Rails 3 without the ID, but where ID is needed in Rails 4
* Adds Strong parameter validation for Braintree Account, Messages, Conversations and Comments

Testing:

- [x] PayPal transaction
- [x] PayPal transaction with booking
- [x] Contact
- [x] Free transaction
- [x] Profile messages
- [x] Comment


ovan:
ok

PR #1615, opened 2015-12-21
Remove Mocha tests from Travis suite

rap1ds:
Due to the latest changes to Sprockets in Rails 4, [it's impossible to compile
assets (like JavaScript files) without the digest in the file name](https://github.com/rails/sprockets-rails#changes-from-rails-3x). That
breaks the current setup, since for Travis tests we need to reference the
compiled file name outside of Rails application, and because there's the digest, it's hard to reference the file.

Mocha tests (JavaScript tests) are rarely used. The tests haven't probably
ever failed. I don't think these tests are super useful.

If needed, the Mocha tests can be still accessed if the Rails server is
running.

P.S. What do you think? Should I acctually remove all the tests and the Mocha framework? Do we need that, ever?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ env:
   - SUITE="rspec"
   - SUITE="rubocop"
   - SUITE="cucumber"
-  - SUITE="mocha"
   - SUITE="jshint"
   global:
     - BUNDLE_ARCHIVE="travis-bundle-cache"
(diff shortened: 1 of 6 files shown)


ovan:
Seems there's only a couple of tests so these are not important for the overall quality. Going forward I don't see us adding more to this JS codebase. Still, it feels a bit bad to decrease the test coverage. If travis is the problem I would maybe keep the tests still around so you can run them manually (how to do this, could you document it somewhere?)

rap1ds:
It's documented here: https://github.com/sharetribe/sharetribe/blob/master/docs/testing.md#run-mocha-tests-on-browser

At the bottom of the README file, there's a link to that doc, but it's not very visible. We could add the content of the Testing document to the README instead of linking the document...

PR #1616, opened 2015-12-22
Prepare Rails 4, cont. 5

rap1ds:


rap1ds:
The issue is in Conversation and Participants models. I quickly tried to fix the circular dependency, but I didn't know how... Seemed quite complicated. I guess I should have another look.

PR #1617, opened 2015-12-22
Fix circular dependency

rap1ds:
Destroying person will destroy participations which will destroy conversation

Destroying conversation will not destroy participation


ovan:
\o/  *-<:D

PR #1618, opened 2015-12-23
Remove asset sync

rap1ds:
This PR makes some code changes that allow us to deliver assets from CDN.

Changes:

- [X] Remove asset_sync
- [X] Add some names to blacklisted marketplace names
- [X] Add `asset_host` to ENV variables
- [X] Add `config/environments/common.rb`. This file contains all the configurations that are common to all environments (or can be controlled by environment variables). Ideally, all configurations should go here (i.e. all configurations should be in environment variables)


rap1ds:
Maybe I should write a short section to TODO how to configure CDN (or point links to good resources)

config/config.defaults.yml
```diff
@@ -37,6 +37,9 @@ default: &default_settings
   # Bucket for files users upload directly to S3. You can add an expiration policy to these files.
   s3_upload_bucket_name:
 
+  # Your CDN distribution host name
```
rap1ds: Example of %d

PR #1620, opened 2015-12-28
Add digest to missing profile image

rap1ds:
Currently, the picture for a person without an avatar is rendered without the digest. This will break in Rails 4, which will only deploy assets with digest.

- [X] Remove the default_url from the Person model
- [X] Where ever we use the image, check that it exists and use default image if that's not the case


Gnito:
Looks good. The only comment was to use new helper function even in .haml

PR #1621, opened 2015-12-29
Add comm pending ext IPN msg & xform pending reason

ovan:
When commission payment doesn't go through because the marketplace admin doesn't have balance in marketplace currency (a rare case) we get IPN message that informs commission payment is in pending ext state. This message was incorrectly recognized as payment pending ext message and later on the parsing failed because these messages have different fields. This led to commission pending ext IPN messages been dropped. This PR adds handling for the IPN message type. It also adds proper normalization for commission pending reason states before saving them to DB which was preciously missing.


app/services/paypal_service/store/paypal_payment.rb
```diff
@@ -161,15 +161,22 @@ def create_payment_update(update, current_state)
...
+    commission_new_status = transform_status(update[:commission_status]) if update[:commission_status]
+    commission_new_pending_reason = transform_pending_reason(update[:commission_pending_reason])
+
     if(new_state && valid_transition?(current_state, new_state))
       payment_update[:payment_status] = new_status
       payment_update[:pending_reason] = new_pending_reason
     end
 
-    payment_update[:commission_status] = transform_status(update[:commission_status]) if update[:commission_status]
+    if (commission_new_status)
```
rap1ds: Is it possible that the status remains the same (e.g. pending) but the pending reason changes? If that scenario is possible, then the new reason would be ignored.

ovan: No it wouldn't be ignored. It's only skipped if the incoming message doesn't contain any data about commission_status and commission_status is nil. This happens most of time because normal payment updates don't include commission payment information.

PR #1635, opened 2016-01-08
Fix shared account testimonials and list of followed people

gnomet:
Testimonials and followed people were shown even if they were not related to current community. This issue would not exists if user accounts would be separate, which is the next topic in the support/growth column of dev roadmap, so I assume it will be done pretty soon and then this PR will be nearly useless. :)

Anyway several clients complained about this and it indeed creates very broken feeling, so I did a fix that hides the wrong stuff. Based on the short life expectancy of this code I took two shortcuts to spend less time now:
 - the filtering is done in ruby, which can slow the profile page if there's lot of testimonials or followed people, but very few have loads of those, and profile page is not that super frequently viewed, so this hardly matters (before the big fix comes).
 - I did the temporary methods as Ruby model methods and did not create service classes for testimonials or followed people (as those helper methods could be removed after the bigger fix and service classes probably left empty.

I hope the use of these shortcuts is approved in this case. :)
As a bonus I cleaned up some code related to old testimonial model.. :)


ovan:
Had only quick look now. You don't need to build a service since we don't even have the structure for people service but instead of more stuff to 600+ lines person.rb I'd rather see these methods in controllers and then explicitly passed into views instead of views pulling stuff from models.

This makes it easier to follow the dataflow and also see why a certain method exists (it's needed by the controller that defines it) instead of going though 600+ line person.rb and trying to figure out what is dead code and what breaks when something is changed.

ovan:
Great, the dataflow is know much better and no views calling back to models!

There's one problem introduced know that you moved code to controllers. That is controllers callind other controllers. Namely testmonials_controller and followed_people_controller valling to PeopleController. This is a problem because the assumption is that if I delete all routes pointing to a controller it's always safe to delete the controller after that. There's two ways to deal with this:

1. Introduce a view util that hosts the shared finders
2. Just copy & paste the code

For PeopleController.followed_people_in_community I would just copy & paste. There's a case of too much DRYness and this is probably it. The logic is very clear even if it's not abstracted away behind a shared method. For received_testimonials, positive_testimonials and negative_testimonials I'd go with TestimonialsViewUtil and put them there.

gnomet:
@ovan thanks for comments. Although the refining rounds can at times feel frustrating, I also feel that I'm learning something while doing this code mode sessions. :)

I also felt that the controller didn't feel the best place for shared methods (and my old Rails logic suggested the model being the right place :) but the approach you suggested sounds to make sense so I'll follow that now.

ovan:
Looks good now! Good to go (couldn't check what the Travis failure was about because Travis doesn't load the output for me :/).

gnomet:
Thanks!
I tested the failed tests locally and passed, so should be safe to deploy. I'll do that tomorrow.

PR #1636, opened 2016-01-08
Handle UnknownFormat and RoutingError as 404 instead of 500

mporkola:
Users accessing weird paths can result to 500 errors in some cases. Semantically 404 would be more correct, and create less log clutter.

The `rescue_from` statements need to be in `ApplicationController` since other controllers inherit it, but our error handlers are in `ErrorsController`. I'm not sure what's the best way to solve this, right now I've duplicated the 404 handler to `ApplicationController`. Comments welcome!

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -44,6 +44,16 @@ class ApplicationController < ActionController::Base
 
   rescue_from RestClient::Unauthorized, :with => :session_unauthorized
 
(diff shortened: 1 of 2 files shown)


ovan:
I've no idea what's the Rails best practise for this but your solution looks ok to me. The risk with duplication would be that if we make changes to 404 page we forget to make the changes to 2 places. However, both render same view and the view file is the most likely spot for changes so I don't see a problem here.

Ok to merge.

PR #1638, opened 2016-01-13
Hide "contact support" message from email config page

mporkola:
If the marketplace doesn't use SES, they don't need to contact us for changing the sender email. In practice, this applies to open source installations

### DIFF
--- a/app/views/admin/communities/edit_welcome_email.haml
+++ b/app/views/admin/communities/edit_welcome_email.haml
@@ -34,7 +34,8 @@
   - content_for(:resend_link) do
     = link_to(t("admin.communities.outgoing_email.resend_link"), nil, class: "js-sender-address-resend-verification")
 
-  - if user_defined_address.nil?
+  - if user_defined_address.nil? || !ses_in_use
+    -# SES not in use, or the marketplace uses default sender address
     .row
       .col-12
         .sender-address-preview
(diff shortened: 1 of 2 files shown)


app/views/admin/communities/edit_welcome_email.haml
```diff
@@ -34,7 +34,7 @@
   - content_for(:resend_link) do
     = link_to(t("admin.communities.outgoing_email.resend_link"), nil, class: "js-sender-address-resend-verification")
 
-  - if user_defined_address.nil?
+  - if user_defined_address.nil? || !ses_in_use
```
mporkola: I'm not sure what the intent is in `user_defined_address.nil?`, so I was not sure if I should remove it or not.

ovan: I'm guessing that if user defined address is not set we show controls to set it. If it's set, then we don't show the controls because chancing address set one time is not supported by the ses integration at the moment. This needs manual intervention from us.

Should this be:
```ruby
if user_defined_address.nil? && ses_in_use
```

So if user hasn't defined an address and we have ses in use then show the controls? Or maybe I'm misinterpreting this but it's hard to tell what shown and hidden just from the haml below.

ovan:
I'm not sure I fully understood the intention here. Maybe if you demo this to me it's easier to get the idea?

PR #1642, opened 2016-01-14
add link to new listing page to empty search results text

gnomet:
Minor improvement based on client's idea.
At the same time added "ref" query params so we can track if anyone uses this kind of links actually

### DIFF
--- a/app/views/homepage/index.haml
+++ b/app/views/homepage/index.haml
@@ -135,6 +135,10 @@
       - else
         .home-no-listings
           - if params[:q] || params[:category] || params[:share_type] # Some filter in use
-            = t(".no_listings_with_your_search_criteria")
+            = t(".no_listings_with_your_search_criteria",
+                post_your_own_listing_link: link_to(t(".post_your_own_listing_link_text"),
+                new_listing_path(ref: "post_your_own_listing_link"))).html_safe
           - else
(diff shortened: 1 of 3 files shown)


app/views/homepage/index.haml
```diff
@@ -135,6 +135,10 @@
       - else
         .home-no-listings
           - if params[:q] || params[:category] || params[:share_type] # Some filter in use
-            = t(".no_listings_with_your_search_criteria")
+            = t(".no_listings_with_your_search_criteria",
```
rap1ds: As I mentioned already in Flowdock, I feel that "Post a new listing" is not the right call to action, if no listings is found. Maybe it should be something that points user to loosen the search criteria?

app/views/homepage/index.haml
```diff
@@ -137,4 +137,6 @@
           - if params[:q] || params[:category] || params[:share_type] # Some filter in use
             = t(".no_listings_with_your_search_criteria")
           - else
-            = t(".no_listings_notification", :add_listing_link => link_to(t(".add_listing_link_text"), new_listing_path)).html_safe
+            = t(".no_listings_notification",
+                  add_listing_link: link_to(t(".add_listing_link_text"),
+                  new_listing_path(ref: "add_listing_link"))).html_safe
```
rap1ds: What's the `ref` parameter? Are we actually using it to something?

rap1ds:
The code looks good :+1: Didn't test it. Added one question.

PR #1643, opened 2016-01-14
Handle UnknownFormat and RoutingError as 404 instead of 500

mporkola:
Users accessing weird paths can result in errors in some cases, which creates log clutter. In this PR I tell the reporting clients to ignore the most common errors resulting from accessing invalid paths, as they are not actual errors.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -42,6 +42,22 @@ Kassi::Application.configure do
       request_uuid: event.payload[:request_uuid] }
   }
 
+  # NB: ignores for other services are defined in airbrake.rb and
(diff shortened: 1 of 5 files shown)


config/environments/production.rb
```diff
@@ -42,6 +42,22 @@
       request_uuid: event.payload[:request_uuid] }
   }
 
+  # NB: ignores for other services are defined in airbrake.rb and
+  # newrelic.yml, consider keeping them in sync!
+  config.lograge.ignore_custom = lambda do |event|
```
rap1ds: I'd rather use the newer lambda syntax, `-> (event) { ... }`

PR #1645, opened 2016-01-15
Fix big cover image to be cover image

Gnito:
This adds the content of big cover image inside the cover image. So that the cover image will stretch with the title & slogan.

After working with Thomas we figured out that there ain't that many active marketplaces, which have made changes to the affected dom structure with custom scripts.

### DIFF
--- a/app/assets/stylesheets/main.css.scss
+++ b/app/assets/stylesheets/main.css.scss
@@ -1979,18 +1979,6 @@ figure.fluidratio {
   background-position: center center;
 }
 
-.lander-content {
-  position: absolute;
-  left: lines(0.5);
-  right: lines(0.5);
-  text-align: center;
-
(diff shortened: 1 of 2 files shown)


rap1ds:
The code looks good :+1: 

About making the cover photo to stretch also vertically, well, I have my doubts about that as we have been discussing, but if this is though with care and the end result of that though process is that this is the best way to do it, then I'm ok with it.

When deploying this change, the frontpage layout will break for every user who has custom styles, until the new stylesheet has been compiled (See the screenshots). Any ideas how to mitigate this? Are we able to implement the code so that both new and old stylesheet work (best option), or do we just add workers and try to make the affected time short as possible?

Before:

screen shot 2016-02-02 at 08 44 17

After:

screen shot 2016-02-02 at 08 42 26

PR #1650, opened 2016-01-18
Refactor custom fields

rap1ds:
This PR removes `attr_accessible` from CustomField and related models. `attr_accessible` needs to be removed because it's deprecated in Rails 4.

The parameter validation is handled by the EntityUtils in `custom_field_controller`.

Also, CustomField model is changed to use new Hash syntax (key: :value, instead of key => :value)

Todo: 

- [X] Use new style hash syntax
- [X] Get rid of attr_accessible

### DIFF
--- a/app/models/custom_fields/option_field.rb
+++ b/app/models/custom_fields/option_field.rb
@@ -45,11 +45,11 @@ class OptionField < CustomField
       }
     }
 
-    attributes_hash = attributes.map { |(option_id, opts)|
+    attributes_hash = attributes.map { |opts|
       {
-        id: Maybe(option_id).to_i.or_else(nil),
-        sort_priority: opts["sort_priority"].to_i,
-        title_attributes: opts["title_attributes"]
+        id: Maybe(opts[:id]).to_i.or_else(nil),
+        sort_priority: opts[:sort_priority].to_i,
+        title_attributes: opts[:title_attributes]
       }
     }
 
--- a/app/models/custom_fields/option_field.rb
+++ b/app/models/custom_fields/option_field.rb
@@ -47,7 +47,7 @@ class OptionField < CustomField
 
     attributes_hash = attributes.map { |opts|
       {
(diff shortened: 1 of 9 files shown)


app/views/admin/custom_fields/_new_option.haml
```diff
@@ -4,6 +4,7 @@
       - show_locales = available_locales.size > 1
       .col-8{:class => show_locales ? "" : "custom-field-column-without-bottom-margin"}
         = fields_for "custom_field[option_attributes][#{option_id}][]", option do |option_form|
+          = hidden_field_tag "custom_field[option_attributes][#{option_id}][id]", option_id
```
rap1ds: This is added to make the it possible to use Entities.

Previously the format of the options was following:

```
{
  "123" => { sort_priority: 1, ... other option values },
  "234" => { sort_priority: 2, ... other option values },
}
```

That format made it difficult to use the ID, because it's only in the hash key. However, changing the format to this:

```
{
  "123" => { id: "123" sort_priority: 1, ... other option values },
  "234" => { id: "234", sort_priority: 2, ... other option values },
}
```

...makes it easier to access to the ID and now it's possible to define an Entity with `id` and `sort_priority`.

app/models/custom_fields/option_field.rb
```diff
@@ -67,4 +67,15 @@ def option_attributes=(attributes)
       options.where(id: changed[:id]).first.update_attributes(changed)
     }
   end
+
+  def str_to_integer(s)
```
sktoiva: Why not:
```ruby
(i = s.to_i) >= 0 ? i : nil
```

rap1ds: It seems that `"string".to_i` returns `0`, and thus

```
s = "string"
(i = s.to_i) >= 0 ? i : nil
```

...returns 0 and not the desired `nil`.

PR #1653, opened 2016-01-19
Show and hide filters on homepage

rap1ds:
This PR adds a new feature which allows admins to define whether the custom field should be shown as filter on the homepage.

Filtering is only available but number, dropdown and checkbox fields. Date fields can not be filtered. Text fields are searched by the main text search.

Todo:

- [X] Add a new checkbox to the custom field admin page
- [X] Disable the checkbox for text and date fields
- [X] Show informative message for checkbox and text fields telling why the checkbox is disabled
- [X] Change the entity definition on the admin controller so that the `show_filter` value for text and date fields are always false
- [X] On homepage, use SQL query to define which filters to show
- [X] Add database migrations

Screenshots:

screen shot 2016-01-20 at 10 16 33
screen shot 2016-01-20 at 10 16 46
screen shot 2016-01-20 at 10 16 58


rap1ds:
Tests are failing...

db/migrate/20160119092239_add_show_filter_to_custom_fields.rb
```diff
@@ -0,0 +1,6 @@
+class AddShowFilterToCustomFields < ActiveRecord::Migration
+  def change
+    add_column :custom_fields, :show_filter, :boolean, after: :sort_priority, default: false, null: false
```
ovan: I have a feeling that show_filter as name doesn't communicate the idea clearly. How about acts_as_filter or search_filter ? My thinking is that show_filter is a view level idea but there's bigger idea behind which is that some categories act as search filter and others act as additional listing information only.

app/controllers/homepage_controller.rb
```diff
@@ -21,7 +21,9 @@ def index
     listing_shape_menu_enabled = all_shapes.size > 1
     @show_categories = @categories.size > 1
     show_price_filter = @current_community.show_price_filter && all_shapes.any? { |s| s[:price_enabled] }
-    @show_custom_fields = @current_community.custom_fields.any?(&:can_filter?) || show_price_filter
+
+    filters = @current_community.custom_fields.where(show_filter: true).sort
```
ovan: This needs at the minimum a DB index for the query that is generated. It would also feel like good idea to cache this to memcached and/or in-memory cache. There are loaded on every single pageload to homepage.

rap1ds: Damn! I always forgot the indecies!

db/migrate/20160119092239_add_show_filter_to_custom_fields.rb
rap1ds: Good point. I'll go with the `search_filter` option. In Ruby context `acts_as_x` might be confusing because there are a lot of popular gems with that name.

app/controllers/homepage_controller.rb
```diff
@@ -21,7 +21,9 @@ def index
     listing_shape_menu_enabled = all_shapes.size > 1
     @show_categories = @categories.size > 1
     show_price_filter = @current_community.show_price_filter && all_shapes.any? { |s| s[:price_enabled] }
-    @show_custom_fields = @current_community.custom_fields.any?(&:can_filter?) || show_price_filter
+
+    filters = @current_community.custom_fields.where(search_filter: true).sort
```
ovan: I'm wondering what kind of SQL query this generates with the sort. What's the sort based on when no field is specified? Is a single field index good enough to satisfy this (if it's e.g. filtering by search_filter but sorting by id)?

rap1ds: `sort` uses the default `<=>` method. So in this case the `sort` doesn't affect to the SQL query, instead it's just a normal sorting operation on the Ruby side.

The `CustomField` models includes `SortableByPriority` module, which implements the `<=>` method like this:

```ruby
  # Compare two object.
  #
  # Usage:
  # - Include this module to your ActiveRecord Model
  # - Add `sort_priority` to the model
  #
  # Returns 0, if objects are equal, otherwise by sort priority
  def <=> other
    (sort_priority || 0) <=> (other.sort_priority || 0)
  end
```

ovan: gotcha!

PR #1655, opened 2016-01-19
settings -> paypal: title was missing, subtitle h1 -> h2

Gnito:
Small fixes:
User's profile menu > settings > paypal
- no title on the top (should be "Settings")
- only one h1 on the page -> payout info title changed to h2


Gnito:
Antti said this is OK :)

PR #1656, opened 2016-01-19
remove version control from selenium-webdriver

Gnito:
Suggestion: Since Firefox is updated behind the scenes all the time, should we remove version lock from selenium-webdriver?

I guess, this gemfile is more often broken with version lock than without it.


Gemfile
```diff
@@ -149,7 +149,7 @@ group :test do
   gem 'capybara', "~> 2.2.1"
   gem 'cucumber-rails', '~> 1.4.0', :require => false
   gem 'cucumber'
-  gem 'selenium-webdriver', '~> 2.45.0'
```
rap1ds: What about `~> 2.45`? That would allow updating to 2.46, 2.47, 2.48 etc, but it would not allow to update to 3.0.

Gnito: Sounds OK to me. 
(Current version is 2.49 i.e. we have not updated this very often.)

PR #1658, opened 2016-01-20
Versioning, Change log and release instructions

rap1ds:
This PR defines a set of practises for keeping versioning/changelog/upgrade notes up-to-date:

Todo

- [X] Add CHANGELOG.md. Follow the best-practices from [keepchangelog.com](http://keepchangelog.com)
- [X] Add UPGRADE.md. This file contains upgrade notes
- [X] Add RELEASE.md. This file contains steps how to make a new release
- [x] Modify CONTRIBUTING.md: Add a new bullet about updating CHANGELOG
- [X] Upgrade README: Add links to the new files. Say explicitely that Sharetribe is using Semantic Versioning.

Review tip: [Click here](https://github.com/sharetribe/sharetribe/tree/69a3300922278565dd4641fba9518799063c5976) to see how these changes look in the Github UI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to this project will be documented in this file.
 
-This project adheres to [Semantic Versioning](http://semver.org/).
+This project adheres to [Semantic Versioning](http://semver.org/) where possible.
 
 This file follows the best practices from [keepchangelog.com](http://keepachangelog.com/).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The Change Log starts from the release 5.0.0. For older release, see [RELEASE_NO
 
 ### Added
 
-- CHANGELOG, UPGRADE and RELEASE files (#1234)
+- CHANGELOG, UPGRADE and RELEASE files (#1658)
 
 ## [5.0.0] - 2015-12-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The Change Log starts from the release 5.0.0. For older release, see [RELEASE_NO
 
 ### Added
 
-- CHANGELOG, UPGRADE and RELEASE files (#1658)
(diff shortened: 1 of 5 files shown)


rap1ds:
I'm still wondering one thing: I think we don't need to do a new release each time we merge a Pull Request. We can use common sense to gather a nice group of additions/fixes and then make a new release. Which may lead to a problem: If I want to make a new release now to get my PATCH level change out, how do I know if there are other unreleased MAJOR or MINOR level changes? Should we explicitly add that information to the CHANGELOG so that by reading the CHANGELOG you know what part of the version number to increase?

Here's an example:

```
CHANGELOG.md

## [Unreleased]

### Added
- MINOR Added support for payment gateway xyz [#99999](https://github.com/sharetribe/sharetribe/pull/99999)

### Fixes
- PATCH Fixes typos in README.md
- PATCH Fixed broken link in About page
```

By looking at this CHANGELOG it's obvious that the next version number needs to increase the MINOR version.

bladealslayer:
I think it would be nice if every PR comes with a corresponding entry in CHANGELOG.

CHANGELOG.md
```diff
@@ -0,0 +1,24 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+
+This project adheres to [Semantic Versioning](http://semver.org/).
```
kpuputti: I know this isn't legally binding text, but should this be stated with a warning? Something like "This project adheres to Semantic Versioning where possible." etc. Otherwise we might have people being surprised when a minor version change breaks existing functionality. We cannot really account for all cases like a library with a public API.

rap1ds:
@bladealslayer I totally agree. I added a note about that in the CONTRIBUTING.md https://github.com/sharetribe/sharetribe/pull/1658/files#diff-6a3371457528722a734f3c51d9238c13R15

CHANGELOG.md
rap1ds: That's very true. I'll change this as you suggested.

PR #1659, opened 2016-01-20
Replace "home page" by "homepage" in locale file

thomasmalbaux:
In texts, we mostly use "homepage" and there were a few places left where "home page" was used instead.
This updates texts to use "homepage" everywhere to be consistent.


rap1ds:
Looks good!

PR #1663, opened 2016-01-21
Add pessimistic version number to all gems in Gemfile

rap1ds:
This PR adds pessimistic version number (e.g. `~> 1.2.3`) to all gems in Gemfile. This means that we can finally run `bundle update` to update all gems where there are new PATCH level updates.

Merge after #1625


Gemfile
```diff
@@ -133,38 +134,35 @@ group :development, :staging do
...
-  gem 'cucumber'
+  gem 'cucumber-rails', '~> 1.4.0', require: false
+  gem 'cucumber', '~> 1.3.14'
   gem 'selenium-webdriver', '~> 2.45.0'
-  gem 'launchy'
-  gem 'ruby-prof'
-  gem "pickle"
+
+  # Launchy is needed by Capybara, e.g. save_and_open command needs Launchy to open a browser
+  gem 'launchy', '~> 2.4', '>= 2.4.3'
```
sktoiva: Why on others do we limit the patch level only and here the minor version?

Gemfile
```diff
@@ -54,21 +57,20 @@ gem 'possibly', '~> 0.2.0'
...
-gem 'json', "~>1.8.0"
-gem 'multi_json', "~>1.7.3" # 1.8.0 caused "invalid byte sequence in UTF-8" at heroku
-gem 'web_translate_it'
-gem 'rails-i18n'
+gem 'delayed_job_active_record', "4.0.3"
+
+gem 'json', "~>1.8.3"
+gem 'multi_json', "~>1.7.9" # 1.8.0 caused "invalid byte sequence in UTF-8" at heroku
+gem 'web_translate_it', "~>2.1.8"
+gem 'rails-i18n', '~> 4.0', '>= 4.0.8'
```
sktoiva: Why on others do we limit the patch level only and here the minor version?

sktoiva:
Tests fail.

Gemfile
rap1ds: That's a good point. I'll change that.

In general, if all gems would follow strictly Semantic Versioning, we should be some to limit the updates to MINOR versions, since only MAJOR versions add breaking changes. However, according to my tests, that's not the case. Also MINOR updates will break. So better to be on the safe side and be a bit more pessimistic.

rap1ds:
@sktoiva 's review comment was fixed and tests are now working. I'll take the liberty to add the label review done.

PR #1666, opened 2016-01-22
Search settings table

Gnito:
~~This is an old branch, which I haven't pushed to the production due to changes in plans last fall.~~

Things to consider: beside deciding main search (and maybe secondary search later) is there other search settings that we could add to this table? (Current plan is to add distance unit there too - km & miles)


mporkola:
I think we shouldn't guess the settings we need in the future, let's add them when needed. I'm pretty sure we should have index for `community_id` field since we'll be searching by that a lot. Which brings to my mind, should these settings maybe be stored in the same place as other community related settings? Just a thought, haven't really participated in planning this initiative

rap1ds:
I didn't review the code but just wanted to remind that please add a CHANGELOG entry before merging! 😉

app/models/search_settings.rb
```diff
@@ -0,0 +1,21 @@
...
+#
+#  id           :integer          not null, primary key
+#  community_id :integer          not null
+#  main_search  :string(255)      default("KEYWORD"), not null
+#  created_at   :datetime         not null
+#  updated_at   :datetime         not null
+#
+
+class SearchSettings < ActiveRecord::Base
+  attr_accessible(
```
sktoiva: With the new RoR is attr_accessible anymore needed? These should be validated in controller level.

rap1ds: @sktoiva Good catch, that's exactly right. We shouldn't use `attr_accessible` anymore. I'd also consider about the presence validators. Is it a Model level thing or should that be also done in Controller/Service API ?

Gnito:
@mporkola Pretty much all the other (this kind of) community related settings are on Communities table. (Community_customizations contain language related stuff.) Our current policy is not to add anything there but rather remove stuff to other tables. 

@rap1ds, thanks for the reminder!

app/models/search_settings.rb
Gnito: @sktoiva @rap1ds thanks!
I can remove those. In the next step / branch I'm planning to add service API access point to MarketplaceService::API (unless you have better suggestions). 

Currently, I feel that a separate Search Settings Service for one flag is too much (and the actual new "search service" is called Listing Index Service, which seems to be something else than this flag for search bar type).

app/services/marketplace_service/store/search_settings.rb
```diff
@@ -0,0 +1,49 @@
+module MarketplaceService::Store::SearchSettings
+
+  SearchSettingsModel = ::SearchSettings
+
+  SearchSettings = EntityUtils.define_builder(
+    [:community_id, :mandatory, :fixnum],
+    [:main_search,:to_symbol, :mandatory, one_of: [:keyword, :location]]
```
rap1ds: `:mandatory` is not mandatory (pun intended). `one_of` guarantees that the value is not `nil`

app/services/marketplace_service/store/search_settings.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def from_model(model)
+    Maybe(model)
+      .map { |m| EntityUtils.model_to_hash(m) }
+      .map { |hash| SearchSettings.call(hash) }
+      .or_else(nil)
+  end
+
+  def find(community_id)
+    SearchSettingsModel.where(community_id: community_id).first
```
rap1ds: New Rails 4 thing! You can also use `SearchSettingsModel.find_by(community_id: community_id)`

rap1ds: Actually, `find_by` may throw an error if the record is not found.

app/controllers/homepage_controller.rb
```diff
@@ -91,6 +91,8 @@ def index
                  listing_shape_menu_enabled: listing_shape_menu_enabled }
       }
     end
+
+    @main_search = MarketplaceService::API::Marketplaces.search_settings(community_id: @current_community.id).data[:main_search]
```
rap1ds: I guess you could use locals instead of instance variable.

app/controllers/admin/communities_controller.rb
```diff
@@ -180,12 +180,29 @@ def test_welcome_email
...
+    # When feature flag is removed, make this pretty
+    if feature_enabled?(:location_search)
+      search_settings = MarketplaceService::API::Marketplaces.search_settings(community_id: @current_community.id).data
+
+      main_search_select_options = [:keyword, :location]
+        .map { |type|
+          [SettingsViewUtils.search_type_translation(type), type]
+        }
+
+      locals.merge!({
```
rap1ds: Using bang-method like `merge!` is kinda bad, because you are mutating the value. Instead of mutating the `locals`, you could do something like this:

```ruby
locals = { ... }

search_settings = 
  if feature_enabled?(:location_search)
    {
      main_search: ...
      ...
    }
  else
    {}
  end

render :settings, locals: locals.merge(search_settings)

File: app/views/homepage/_location_bar.haml [hunk 6]
```diff
@@ -8,12 +8,17 @@
...
+      autocomplete.setTypes(['geocode']);
 
       google.maps.event.addListener(autocomplete, 'place_changed', function(event){
         var place = autocomplete.getPlace();
-        if(place != null && place.name != "") {
-          console.log('name:', place.name, '\naddress: ', place.formatted_address, '\ncoordinates: ', place.geometry.location.toUrlValue());
-          document.getElementById('lc').value = place.geometry.location.toUrlValue();
+        if(place != null) {
+          if( place.geometry != null ) {
+            console.log('name:', place.name, '\naddress: ', place.formatted_address, '\ncoordinates: ', place.geometry.location.toUrlValue());
```
rap1ds: remove

app/views/homepage/_location_bar.haml
```diff
@@ -8,12 +8,17 @@
...
       var autocomplete = new google.maps.places.Autocomplete(input);
+      autocomplete.setTypes(['geocode']);
 
       google.maps.event.addListener(autocomplete, 'place_changed', function(event){
         var place = autocomplete.getPlace();
-        if(place != null && place.name != "") {
-          console.log('name:', place.name, '\naddress: ', place.formatted_address, '\ncoordinates: ', place.geometry.location.toUrlValue());
-          document.getElementById('lc').value = place.geometry.location.toUrlValue();
+        if(place != null) {
+          if( place.geometry != null ) {
```
rap1ds: Somewhat inconsistent coding style here. Is it `if(place...)` or `if( place... )`?

app/views/homepage/_location_bar.haml
```diff
@@ -8,12 +8,17 @@
...
+            console.log('name:', place.name, '\naddress: ', place.formatted_address, '\ncoordinates: ', place.geometry.location.toUrlValue());
+            document.getElementById('lc').value = place.geometry.location.toUrlValue();
+          } else {
+            console.log("Please select some place from the autocomplete list.")
+          }
         }
         document.getElementById('homepage-filters').submit();
```
rap1ds: This is already a good amount of JavaScript. Should it be in it's own .js file instead of inlining?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The Change Log starts from the release 5.0.0. For older releases, see [RELEASE_N
 
 - Added pessimistic version number for all the gems in Gemfile. Now we can safely run `bundle update` to update gems with patch level updates. [#1663](https://github.com/sharetribe/sharetribe/pull/1663)
 - Added a new environment variable `delayed_job_max_run_time` which controls the maximum time for a single Delayed Job job. [#1668](https://github.com/sharetribe/sharetribe/pull/1668)
+- Added search_settings table and API access point to MarketplaceService for storing community related search settings (at this point: search type of the main search) [#1666](https://github.com/sharetribe/sharetribe/pull/1666)
```
rap1ds: Now that I think it again, should we remove this entry? I mean, this PR doesn't add anything new for a open source user. It's not end-to-end ready yet. So someone reading the changelog may be confused.

Sorry about this, I know that I was the one who originally pointed out that you should add the CHANGELOG. 😞

app/controllers/homepage_controller.rb
Gnito: Yeah, I was kind of thinking between these options. 
I thought, that since homepage-index is a huge function and new additions have not take that step there might be a reason / plan to make changes on one huge swipe when the new browse view is coming.

app/controllers/admin/communities_controller.rb
Gnito: The whole merge should be removed when we remove the feature flag. Therefore, I thought that this is more clear of what is the unnecessary part of code - maybe I missed that goal.

app/services/marketplace_service/store/search_settings.rb
Gnito: roger

app/views/homepage/_location_bar.haml
Gnito: true!

app/views/homepage/_location_bar.haml
Gnito: These have been here for testing purposes - to check what kind of data we can get and pass to Rails. 
Both console.logs should be removed, but I think the correct place is to remove those when we decide what data we want to pass to the controller. (Currently this passes just location LatLng, and we want to pass at least bounding box too (NorthEast & SouthWest corners)

app/views/homepage/_location_bar.haml
Gnito: Yes! I should have left some comment that there are two things to do:
1) Add some error handling for situations where user has not selected any of the suggestions
2) Decide what data should be passed to Homepage controller

I was planning to do the first step at least before raising my hands off this part of the search. (i.e. new branch & more .js code there.)

CHANGELOG.md
Gnito: That is true. I'll remove this.

app/controllers/homepage_controller.rb
Gnito: So, essentially I took the easy way out :/

rap1ds: Don't take the easy way 😉

app/views/homepage/_location_bar.haml
rap1ds: Ok, I understand. But do we really want to have them also in production? To me it gives a bit unprofessional feeling if I see a site that is recording test data to the console on production.

Gnito: These are behind feature flag, but OK - I'll take them out.

app/services/marketplace_service/store/marketplace_configurations.rb
```diff
@@ -0,0 +1,49 @@
+module MarketplaceService::Store::MarketplaceConfigurations
+
+  MarketplaceConfigurationsModel = ::MarketplaceConfigurations
+
+  MarketplaceConfigurations = EntityUtils.define_builder(
+    [:community_id, :mandatory, :fixnum],
+    [:main_search,:to_symbol, one_of: [:keyword, :location]]
```
rap1ds: Space between  `:main_search,:to_symbol`

app/controllers/admin/communities_controller.rb
```diff
@@ -266,6 +285,8 @@ def update_settings
 
     maybe_update_payment_settings(@current_community.id, params[:community][:automatic_confirmation_after_days])
 
+    MarketplaceService::API::Marketplaces.update_marketplace_configurations(community_id: @current_community.id, main_search: params[:main_search]) if feature_enabled?(:location_search)
```
rap1ds: The namespace (Marketplace) is repeated two times here. Could there be a simple and shorter name with less repetition, e.g. `MarketplaceService::API::Configuration.update( ... )`?

Gnito: So, you would like to create another API file, which would be called from the current one (MarketplaceService::API::Marketplaces) when marketplace is created?

I feel that we are going beyond what we planned with this micro service approach, if we create own APIs for every db tables (and we already have a store layer inside services, which makes this several-tables-behind-unified-api approach possible). 

How about reducing repetition just a bit:
MarketplaceService::API::Marketplaces.update_marketplace_configurations
=> MarketplaceService::API::Marketplaces.update_configurations

Unless, someone thinks this configuration is going to bloat that existing API? Current future plan is to add km/miles selection there too at some point, but that should not add any additional functions to API.

PR #1667, opened 2016-01-22
Update more gems

rap1ds:
- [x] Update `uglifier` gem
- [x] Update `rest-client` gem
- [x] Update `jquery-rails` gem to 2.3.0
- [x] Update `jquery-rails` gem to 3.1.3
- [x] Instead of removed `jQuery.live` use `jQuery.on`


app/assets/javascripts/kassi.js
```diff
@@ -670,17 +670,20 @@ function initialize_update_account_info_form(locale, change_text, cancel_text, e
...
-    },
-    function() {
+    } else {
       $('#account_password_form').hide();
       $(this).text(change_text);
     }
-  );
+
+    changeLinkVisible = !changeLinkVisible;
+  });
```
ovan: There's no toggle in jQuery anymore?

app/assets/javascripts/kassi.js
```diff
@@ -1065,7 +1068,7 @@ function set_textarea_maxlength() {
 
       }
     })
-    .live('keydown', function(event) {
+    .on('keydown', function(event) {
```
ovan: Wasn't this live stuff something about event handlers working even when elements inserted dynamically (ajax) later? Is this an issue here?

app/assets/javascripts/kassi.js
rap1ds: There is, but it does totally different thing.

The old `toggle( fn, fn2, fn3, ... )` run the `fn` on the first click, `fn2` on the second, `fn3, on third and so on. The new `toggle` is only visual show/hide thing.

app/assets/javascripts/kassi.js
rap1ds: Yep, `live` was exactly for that purpose. In this case the event handler is attached directly to the textarea element and thus there will be no dynamically added children there, so it's safe to use `on`.

PR #1669, opened 2016-01-26
Remove initializer that allowed downgrading to Rails 3

rap1ds:
After this, you may not be able to downgrade to Rails 3 without some errors.


mporkola:
looks clean :+1:

PR #1670, opened 2016-01-26
Transaction conversation status with new translation key

thomasmalbaux:
Earlier try to change the sentence and add a variable (%sum) didn't work.
Trying again by creating a new translation key (admin.conversations.message.received_payment) instead of using the current one (admin.conversations.message.accepted_request).

Also reverted the text for old translation key (admin.conversations.message.accepted_request) back to the original value.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -638,7 +638,8 @@ en:
       free_message: "Free message"
       message_content_not_available: "Message content not available"
     message:
-      accepted_request: "accepted the request, received payment for %{sum}"
+      accepted_request: "accepted the request"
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -638,7 +638,8 @@ en:
       free_message: "Free message"
       message_content_not_available: "Message content not available"
     message:
-      accepted_request: "accepted the request, received payment for %{sum}"
+      accepted_request: "accepted the request"
```
rap1ds: This key can be now removed if this is the only code path that uses this key.

PR #1671, opened 2016-01-26
Add secret_key_base

rap1ds:
Rails 4 added secret_key_base. See more information:
http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#action-pack


sktoiva:
Tests fail, and it seems that they are not the usual culprits. Otherwise ok.

rap1ds:
Good point. This needs to be rebased. That should fix the tests.

PR #1672, opened 2016-01-27
Update devise gem to 3.1

rap1ds:
Merge after #1671 

- [x] Update to 3.1
- [x] Fix password reset
- [x] Remove deprecated `token_authenticable`
- [x] Add note about `allow_insecure_token_lookup`
- [x] Test in staging


Gnito:
Had to say that quite much of this stuff would need a bit more understanding than what I have about how the gem works or has been used in our application.

PR #1673, opened 2016-01-27
Fix a custom field order bug

rap1ds:
Steps to reproduce:

1. Create a new custom field "1"
2. Create another custom field "2"
3. Create third custom field "3"
4. Change the order so that you move "3" one step up. The order is now 1, 3, 2
5. Edit the custom field 2. Change its name to "2a"

Expected result:

The order is 1, 3, 2a

Actual:

The order is 1, 2a, 3

### DIFF
--- a/app/controllers/admin/custom_fields_controller.rb
+++ b/app/controllers/admin/custom_fields_controller.rb
@@ -38,6 +38,7 @@ class Admin::CustomFieldsController < ApplicationController
   CUSTOM_FIELD_SPEC = [
     [:name_attributes, :hash, :mandatory],
     [:category_attributes, collection: CategoryAttributeSpec],
+    [:sort_priority, :fixnum, :optional],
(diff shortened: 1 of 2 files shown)


app/controllers/admin/custom_fields_controller.rb
```diff
@@ -160,7 +159,9 @@ def update
     params[:custom_field][:min] = ParamsService.parse_float(params[:custom_field][:min]) if params[:custom_field][:min].present?
     params[:custom_field][:max] = ParamsService.parse_float(params[:custom_field][:max]) if params[:custom_field][:max].present?
 
-    custom_field_entity = build_custom_field_entity(@custom_field.type, params[:custom_field])
+    custom_field_entity = build_custom_field_entity(@custom_field.type, params[:custom_field]).merge(
+      sort_priority: @custom_field.sort_priority
+    )
```
ovan: It feels weird that sort_priority is merged in here when all other fields are set inside constructor build_custom_field_entity. Makes it harder to follow the code / debug.

PR #1675, opened 2016-01-28
Update devise 3.5

rap1ds:
Update Devise gem to 3.5

TODO

- [x] Merge #1672 first
- [ ] Add Changelog


config/routes.rb
```diff
@@ -65,6 +65,9 @@
 
   resources :communities, only: [:new, :create]
 
+
+  devise_for :people, only: :omniauth_callbacks, controllers: { omniauth_callbacks: "sessions"}
```
ovan: Can you quickly explain the effects and idea of pulling omniauth_callbacks here?

rap1ds: I followed the instructions Devise gave me:

```
/Users/mikko/.rvm/gems/ruby-2.1.2/gems/devise-3.5.4/lib/devise/rails/routes.rb:418:in `devise_omniauth_callback': 
Devise does not support scoping OmniAuth callbacks under a dynamic segment
and you have set "/(/:locale)/people". You can work around by passing
`skip: :omniauth_callbacks` to the `devise_for` call and extract omniauth
options to another `devise_for` call outside the scope. Here is an example:

    devise_for :users, only: :omniauth_callbacks, controllers: {omniauth_callbacks: 'users/omniauth_callbacks'}

    scope '/(:locale)', locale: /ru|en/ do
      devise_for :users, skip: :omniauth_callbacks
    end
 (RuntimeError)
```

rap1ds: To my understanding this doesn't have any major effect. The path for the omniauth callback is fetched from `person_omniauth_authorize_path` method, so I guess from now on that method returns a path without any locale in it?

Anyway, I tested this and FB auth seems to be working normally.

PR #1679, opened 2016-01-28
Rename AR relation from `transaction` to `tx`

rap1ds:
`transaction` gives an error in Rails 4.1, because it clashes with
ActiveRecord's `transaction` method


mporkola:
Tried to `git grep transaction` to see if you've missed any, but it's apparently the most common word in our codebase so I'll just trust you. Otherwise, looks good

PR #1686, opened 2016-02-01
Fix checking search args

kpuputti:
The if expression checks if a search needs DB as well as the Sphinx
search engine, and seemingly attempted to make an early return in that
case. However, the execution is continued and a search is always made to
the search engine and the `Result:Error` object is just ignored.


rap1ds:
:+1:

PR #1687, opened 2016-02-01
Improve delayed job tests

rap1ds:
- [x] Make a new hash for options that are passed to DelayedJob
- [x] Update DelayedJob gem
- [x] Test helper without Observer
- [x] Remove Observer gem


spec/support/sync_delayed_job_observer.rb
```diff
@@ -0,0 +1,41 @@
+require 'delayed_job'
+
+class SyncDelayedJobObserver < Delayed::Plugin
```
rap1ds: Remove the `Delayed::Plugin`. This just a PORO, not a plugin.

PR #1688, opened 2016-02-01
Remove ForbiddenAttributesProtection

rap1ds:
- This is included by default in Rails 4


sktoiva:
:+1: Should we do something to that failing test? It doesn't serve as is? Remove?

rap1ds:
Yep, it should be fixed 😕

PR #1689, opened 2016-02-01
Improve transaction conversations sentences

thomasmalbaux:
**Fullfil vs Deliver**
In transaction conversations, the "deliver" status has been really confusing to users. Lots of people understand "deliver" as "ship". This status is the one after the payment happened (so after the seller has accepted the transaction) and before the review process.

The earlier label was `Waiting for %{listing_author_name} to deliver %{listing_title}` and used a moving van icon.

With this change, text is now `Waiting for %{listing_author_name} to fulfill the order for %{listing_title}` with a clockwise icon.

**Payment canceled when request is rejected**
If a request is rejected by the seller, there is not mention that the payment didn't happen. Some users reported this and asked what happened to the payment.

The sentence has been updated to mention that the payment was canceled when the request was rejected.

### DIFF
--- a/app/helpers/transaction_helper.rb
+++ b/app/helpers/transaction_helper.rb
@@ -348,7 +348,7 @@ module TransactionHelper
         t("conversations.status.waiting_for_current_user_to_deliver_listing",
           :listing_title => link_to(conversation.listing.title, conversation.listing)
         ).html_safe,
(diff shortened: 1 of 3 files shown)


rap1ds:
Looks good :+1:

PR #1696, opened 2016-02-02
Location search works without selecting from autocomplete list

Gnito:
- Added possibility to just go with first suggestion (instead of explicitly choosing from Google's Autocomplete list)
- Relocated javascript to a separate file
- Google maps library can't be added twice (i.e. view=map)
- Change search icon back (globallocation wasn't good enough)
- Missing locales

Current parameter naming system: 
 lc = "location coordinates", ls = "location service status", 
[and maybe ne = northeast corner of bounding box, sw = SouthWest corner]
We need to discuss this through when this is integrated with actual search. 
(Status could be logged, to keep track if there's something strange going on - like quota exceeded.)


app/assets/javascripts/location_search.js
```diff
@@ -0,0 +1,73 @@
+window.ST = window.ST || {};
+
+(function(module) {
+
+  module.initializeLocationSearch = function() {
+    var input = document.getElementById('q');
```
rap1ds: Is there a reason why you are not using jQuery?

app/assets/javascripts/location_search.js
```diff
@@ -0,0 +1,73 @@
...
+        }, function(place, status) {
+
+          if(status === serviceStatus.OK) {
+            document.getElementById('lc').value = place.geometry.location.toUrlValue();
+            document.getElementById('ls').value = serviceStatus.OK;
+            document.getElementById('homepage-filters').submit();
+          } else {
+            // Save received service status for logging
+            document.getElementById('ls').value = status;
+            document.getElementById('homepage-filters').submit();
```
rap1ds: The `else` branch is actually the same as the last two lines in `if` branch, so you could remove the `else` branch and take the two lines out of the `if`-block to reduce repetition.

app/assets/javascripts/location_search.js
```diff
@@ -0,0 +1,73 @@
...
+
+      if(status === serviceStatus.OK) {
+        var map = new window.google.maps.Map(document.createElement('div'));
+        var placeService = new window.google.maps.places.PlacesService(map);
+
+        placeService.getDetails({
+          placeId: predictions[0].place_id // first prediction is default
+        }, function(place, status) {
+
+          if(status === serviceStatus.OK) {
```
rap1ds: This is a bit confusing. When I was reading this I was first like "why is this check repeated?" (see line 44), but then I realized it's not the same `status`. Maybe you could rename the later `status` to something else and that way avoid shadowing the first `status`?

app/assets/javascripts/location_search.js
```diff
@@ -0,0 +1,73 @@
...
+            // Save received service status for logging
+            document.getElementById('ls').value = status;
+            document.getElementById('homepage-filters').submit();
+          }
+
+        });
+      } else {
+        // Save received service status for logging
+        document.getElementById('ls').value = status;
+        document.getElementById('homepage-filters').submit();
```
rap1ds: All these code branches seem to end up submitting the form, so maybe you could take the `submit()` away from all the `if-else` branches and just do it at the end of `handlePredictions` function and that way reduce repetition?

app/assets/javascripts/location_search.js
Gnito: With these old sites not a huge one. Vanilla JS is faster, and the code is easier to be moved somewhere without jQuery support (and my secret plan is to get rid of jQuery in the future since it's way too bloated library).

app/assets/javascripts/location_search.js
```diff
@@ -0,0 +1,73 @@
...
+        var map = new window.google.maps.Map(document.createElement('div'));
+        var placeService = new window.google.maps.places.PlacesService(map);
+
+        placeService.getDetails({
+          placeId: predictions[0].place_id // first prediction is default
+        }, function(place, status) {
+
+          if(status === serviceStatus.OK) {
+            document.getElementById('lc').value = place.geometry.location.toUrlValue();
+            document.getElementById('ls').value = serviceStatus.OK;
```
rap1ds: `lc`, `ls` and `homepage-filters` are accessed quite many times. Maybe you could search them once from DOM and cache them to local variable?

app/assets/javascripts/location_search.js
Gnito: True.

app/assets/javascripts/location_search.js
Gnito: Good point

rap1ds:
Couple comments in otherwise looks good :+1:

rap1ds:
Good :+1:

PR #1697, opened 2016-02-03
Fix CSS compilation job, improve metrics

rap1ds:
- [X] Duplicate delayed job parameters before passing them
- [X] Change metrics service code so that only minimal part of the code is environment specific.

### DIFF
--- a/app/services/monitoring_service/monitoring.rb
+++ b/app/services/monitoring_service/monitoring.rb
@@ -1,26 +1,64 @@
 module MonitoringService::Monitoring
+  module LibratoReporter
+    module_function
 
-  SAMPLING_PERIOD_IN_SECONDS = ENV['MONITORING_SAMPLING_PERIOD'] || 15
+    def report(group, measurements)
+      Librato.group group do |g|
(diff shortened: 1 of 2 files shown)


app/services/monitoring_service/monitoring.rb
```diff
@@ -1,26 +1,64 @@
...
   module_function
 
   def report_queue_size
-    if Librato.tracker.should_start? && should_report?
+
+    if should_report?
+      @delayed_job_last_reported = Time.now
+
       # Priorities are from 0..10, where 0..5 are high priority and 6..10 low
       priority_counts = Delayed::Job.where('attempts < ? AND run_at < ?', 3, Time.now).group(:priority).count
```
sktoiva: I guess that originally this was inside the `should_start?` and `should_report?` checks so that this query wouldn't be run if no need for it. 

For debugging, it's beneficial to have it run every time and the no-op reporter in place, for OS developers that's not the case. But maybe here the cost isn't that bad, and the benefits of having the code run in staging and in local env outweigh this.

PR #1700, opened 2016-02-04
Update to RSpec 3

rap1ds:
- [X] Update RSpec
- [X] Specific which error is expected to be raised
- [X] Make delayed job workers silent
- [X] Use `bundle exec rspec` instead of `bundle exec rake` in Travis
- [X] Fix Selenium Webdriver errors


bladealslayer:
Looking good!

PR #1701, opened 2016-02-04
Convert specs to newer rspec syntax

rap1ds:
Used handy transpec tool

TODO

- [x] Convert spec/ syntax
- [x] Add spec types


Gnito:
Looks good and generated

PR #1709, opened 2016-02-05
Update statesman

rap1ds:
The current version uses serialize_attributes, which is deprecated in Rails 4.2


mporkola:
The change from 0.5.0 to 1.3.1 seemed big, so had a look at the changelog. At least 1.0.0 contains some breaking changes, and 1.2.0 seems to have some update steps?

Our tests seem to pass though, which is already suspicious on it's own

rap1ds:
@mporkola Very good point. I was a bit sloppy and just upgraded it and ran the tests (✅) so I didn't look at the changelog at all.

* 1.0.0 breaking change: Doesn't have any effect on us
* 1.2.0 upgrade path: The tests are passing because _"All queries are backwards-compatible, so everything still works without the new column)"_. However, I think I should do the steps in the upgrade path because there seems to be some performance improvements.

rap1ds:
Ok, now I reread the 1.2.0 upgrade path and I changed my mind... maybe we should not run the upgrade steps:

* The upgrade steps assume `ParentModel`, e.g. `Transaction` model should be tightly coupled with the statemachine. Now we don't do that and instead we are using the state machine like this:

  ```
  StateMachine.transition_to(transaction.id, "paid")
  ```

  instead of

  ```
  Transaction.find(123).transition_to("paid")
  ```

  I think that the less coupled is the way to go.

* I hope we can get rid of the Statesman gem soon
* Everything works ok without the upgrade steps (however, we can not upgrade to 2.0.0, but I think that's not so big issue if the aim is to gem rid of the gem.

PR #1710, opened 2016-02-05
Pass person_id explicitely, not person model

rap1ds:
Previously, it has been possible to pass Person model, and Rails magically picks the ID. However, this isn't working anymore in Rails 4.2


mporkola:
makes sense

PR #1712, opened 2016-02-08
old bug: bubble-price is not visible.

Gnito:


rap1ds:
Can you add a screenshot to the pull request?

Gnito:
screen shot 2016-02-08 at 17 03 19

Gnito:
I also checked if this bubble price contain new unit types - they didn't. So, code is now updated with new commit. 
(N.B. if unit price doesn't fit to its space, it will have ellipsis.)

screen shot 2016-02-08 at 18 13 21

rap1ds:
Looks good to me :+1:

PR #1714, opened 2016-02-09
Cache only attributes, serialize/deserialize the result

rap1ds:
`TranslationCache` saves ActiveRecord models to the cache. Saving complex data (e.g. instances of classes) to the cache is a bad idea. This causes problems in Rails 4.2

**Fix:** Save only model attributes (Hash) to the cache and instantiate it manually by hand.

(Btw. `TranslationCache` will be hopefully removed soon, since we have a lot better and more performant solution built, the `TranslationService`)

### DIFF
--- a/app/models/translation_cache.rb
+++ b/app/models/translation_cache.rb
@@ -1,3 +1,5 @@
+# DEPRECATED: Use TranslationService instead
+#
 # Simplifies access to translations that are stored in database. Implements also caching.
 #
 # Usage:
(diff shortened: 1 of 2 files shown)


rap1ds:
Funny thing, this is the exactly same piece of code and logic that caused the cache issue when upgrading to Rails 4.0

app/models/translation_cache.rb
```diff
@@ -36,12 +38,45 @@ def fetch_cache(cache_key, &block)
 
   def translations
     fetch_cache(cache_key) do
-      @model.send(@translation_attr_sym).to_a
-    end
+      @model.send(@translation_attr_sym).to_a.map { |model|
+        serialize(model)
+      }
```
ovan: I cannot say I would fully understand this part of the code. Not a blocker for review.

PR #1716, opened 2016-02-09
Rename serve_static_files and add log level

rap1ds:
- [X] Rename `serve_static_assets` to `serve_static_files`. `serve_static_assets` is deprecated in Rails 4.2
- [X] Add `log_level` configuration.


mporkola:
fine!

PR #1717, opened 2016-02-10
Add a MailCarrier

rap1ds:
`MailCarrier` is a utility layer that is a single point where all mails get delivered. It currently doesn't do much (or anything) but the fact that all mail deliveries are going through the same code path allows us to build features, such as:

- Better logging
- Resend logic (either manually on clients request or automatically if sending fails)

In addition, `Mail::Message#deliver` method is deprecated in Rails 4.2, so the `MailCarrier` layer uses `#deliver_now` instead of `#deliver`

### DIFF
--- a/app/utils/mail_carrier.rb
+++ b/app/utils/mail_carrier.rb
@@ -1,0 +1,15 @@
+module MailCarrier
+
+  module_function
+
+  # This is just a placeholder. Delivering later
+  # hasn't been implemented you.
+  def deliver_later(message)
+    deliver_now(message)
+  end
+
+  def deliver_now(message)
+    message.deliver_now
+  end
+
+end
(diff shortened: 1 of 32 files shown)


app/utils/mail_carrier.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  # This is just a placeholder. Delivering later
+  # hasn't been implemented you.
+  def deliver_later(message)
+    deliver_now(message)
+  end
+
+  def deliver_now(message)
+    message.deliver_now
+  end
```
ovan: Does this endpoint need to make a separation between deliver_now and deliver_later? What if we just have deliver and later we make it so that the only option is to deliver_later, because that's really what we should always do (deliver from worker).

rap1ds: Having only one method `deliver` was indeed my original idea too. However, there are cases where we probably want to send the email right away, e.g. from a worker job.

ovan: Ok, makes sense.

PR #1718, opened 2016-02-10
Update gems

rap1ds:
Run `bundle update` to update gems


ovan:
Not much to say about this except that the Cucumber random failures really suck. For a PR like this it would feel so much better to see a green light there too.

PR #1720, opened 2016-02-10
Add support for new search params for the new Search API

kpuputti:
Include new search params:

 - price_min
 - price_max
 - locale
 - include_closed
 - listing_shape_ids

Also fix search engine selection outside homepage controller.

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -141,10 +141,13 @@ class HomepageController < ApplicationController
       keywords: filter_params[:search],
       fields: checkboxes.concat(dropdowns).concat(numbers),
       per_page: listings_per_page,
-      page: params[:page].to_i > 0 ? params[:page].to_i : 1
+      page: Maybe(params)[:page].to_i.map { |n| n > 0 ? n : 1 }.or_else(1),
+      price_min: params[:price_min],
+      price_max: params[:price_max],
+      locale: I18n.locale,
+      include_closed: false
     }
 
-    search_engine = feature_enabled?(:new_search) || APP_CONFIG.external_search_in_use ? :zappy : :sphinx;
     raise_errors = Rails.env.development?
 
     ListingIndexService::API::Api.listings.search(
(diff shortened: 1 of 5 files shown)


app/controllers/homepage_controller.rb
```diff
@@ -141,10 +141,13 @@ def find_listings(params, listings_per_page, filter_params, includes)
       keywords: filter_params[:search],
       fields: checkboxes.concat(dropdowns).concat(numbers),
       per_page: listings_per_page,
-      page: params[:page].to_i > 0 ? params[:page].to_i : 1
```
rap1ds: I didn't test this, but it looks like there's a possiblity for NullPointer if `params[:page]` is `nil`

kpuputti: Seems so, I didn't even think about that line since it was already there before.

app/helpers/feature_flag_helper.rb
```diff
@@ -12,4 +12,8 @@ def feature_flags
     @feature_flags ||= fetch_feature_flags # fetch_feature_flags is defined in ApplicationController
   end
 
+  def search_engine
+    feature_enabled?(:new_search) || APP_CONFIG.external_search_in_use ? :zappy : :sphinx;
```
rap1ds: Remove ;

app/controllers/listings_controller.rb
```diff
@@ -49,7 +49,18 @@ def index
             per_page: per_page
           }
 
-          listings = ListingIndexService::API::Api.listings.search(community_id: @current_community.id, search: search, includes: includes).and_then { |res|
+          raise_errors = Rails.env.development?
```
rap1ds: Could or should this be environment variable? Maybe? Maybe not? I'm not sure...

rap1ds:
Added some comments. Check the possible `NullPointer`, but other than that no need to fix if you think they are good the way they are now.

app/controllers/homepage_controller.rb
kpuputti: Fixed.

app/controllers/listings_controller.rb
kpuputti: Let's leave it as-is for now.

kpuputti:
Thanks!

rap1ds:
LGTM :+1:

PR #1721, opened 2016-02-10
Update table infos (probably reason is gem/rails updates)

Gnito:
Table info in models has changed
Schema has changed accordingly too.


rap1ds:
:+1:

PR #1722, opened 2016-02-10
Distance units to admin ui

Gnito:


Gnito:
screen shot 2016-02-10 at 15 19 42

config/locales/en.yml
```diff
@@ -224,6 +224,9 @@ en:
         default_search_type: "Search type: %{select_search_type}"
         keyword_search: "Keyword search"
         location_search: "Location search"
+        select_distance_unit: "Show distance from search location in %{distance_units_selector}"
```
vsaarinen: How about `Show location search distances in` or even just `Show distances in`?

Gnito: Roger. I'll go with the shorter version.

app/controllers/admin/communities_controller.rb
```diff
@@ -189,13 +189,20 @@ def settings
...
       render :settings, locals: {
         supports_escrow: escrow_payments?(@current_community),
         delete_redirect_url: delete_redirect_url(APP_CONFIG),
         delete_confirmation: @current_community.ident,
         can_delete_marketplace: can_delete_marketplace?(@current_community.id),
         main_search: marketplace_configurations[:main_search],
-        main_search_select_options: main_search_select_options
+        main_search_select_options: main_search_select_options,
+        distance_unit: marketplace_configurations[:distance_unit],
+        distance_unit_select_options: distance_unit_select_options
```
mporkola: Is this the usual way to give dropdown options? To me it looks like it could be done in HAML with regular translations

mporkola:
Should the labels in general be "metric" and "imperial"? "Km" and "miles" seem to be used in many places here, I think in this case the more general term would be better

app/controllers/admin/communities_controller.rb
```diff
@@ -285,7 +292,13 @@ def update_settings
 
     maybe_update_payment_settings(@current_community.id, params[:community][:automatic_confirmation_after_days])
 
-    MarketplaceService::API::Api.configurations.update(community_id: @current_community.id, main_search: params[:main_search]) if feature_enabled?(:location_search)
+    if(feature_enabled?(:location_search))
```
mporkola: There's a helper method related to this: https://github.com/sharetribe/sharetribe/blob/master/app/helpers/feature_flag_helper.rb#L15

mporkola:
otherwise :+1:

app/controllers/admin/communities_controller.rb
Gnito: I think our general approach has been to remove variables (like these options) away from haml files and try to keep those files as clean 'templates' as possible. In this case the select_tag function takes in data about keys and values of options - and I prefer that data to be processed in controller instead of haml views.

app/models/marketplace_configurations.rb
```diff
@@ -2,11 +2,12 @@
...
 #
-#  id           :integer          not null, primary key
-#  community_id :integer          not null
-#  main_search  :string(255)      default("keyword"), not null
-#  created_at   :datetime         not null
-#  updated_at   :datetime         not null
+#  id            :integer          not null, primary key
+#  community_id  :integer          not null
+#  main_search   :string(255)      default("keyword"), not null
+#  distance_unit :string(255)      default("miles"), not null
```
mporkola: not miles, metric!

app/controllers/admin/communities_controller.rb
mporkola: :+1:

app/controllers/admin/communities_controller.rb
Gnito: Good point. There is now couple of flags related to this:
: location_search & :new_search (& zappy in use)

PR #1732, opened 2016-02-12
Fix bug in Zappy search SQL query generation

kpuputti:
In the Search API response, the total number is separate from the count
of listings in the data. For example, if the user requests &page=2 from
a search that has less listings that fit on one page, the total count is
the number of listings in the first page (i.e. total count), but the
data array is empty, making the SQL query syntax invalid with an empty
array of ids.

The fix is to check the length of the data array to check if we need to
query the DB.

### DIFF
--- a/app/services/listing_index_service/search/zappy_adapter.rb
+++ b/app/services/listing_index_service/search/zappy_adapter.rb
@@ -73,16 +73,19 @@ module ListingIndexService::Search
 
       ids = id_obs.map { |r| r['id'] }
 
-      Listing
-        .where(id: ids)
-        .order("field(listings.id, #{ids.join ','})")
-        .map {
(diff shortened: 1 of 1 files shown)


app/services/listing_index_service/search/zappy_adapter.rb
```diff
@@ -73,16 +73,21 @@ def listings_from_ids(id_obs, includes)
...
+      if ids.empty?
+        []
+      else
+        Listing
+          .where(id: ids)
+          .order("field(listings.id, #{ids.join ','})")
+          .map {
           |l| ListingIndexService::Search::Converters.listing_hash(l, includes)
         }
+      end
```
ovan:
```ruby
Maybe(ids).map { Listings.where(id: ids).order...
```

PR #1737, opened 2016-02-16
Google maps script fix

Gnito:
- Handle feature_enabled?(:location_search) checks a little bit better on index.haml
- Add scripts only on page level (google.maps should not be handled on partials)

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -58,7 +58,6 @@ class HomepageController < ApplicationController
 
     shape_name_map = all_shapes.map { |s| [s[:id], s[:name]]}.to_h
 
-
     if request.xhr? # checks if AJAX request
       search_result.on_success { |listings|
         @listings = listings # TODO Remove
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -51,8 +51,8 @@ class HomepageController < ApplicationController
         raise ArgumentError.new("Unknown view_type #{@view_type}")
       end
 
-    main_search = (feature_enabled?(:location_search) && search_engine == :zappy) ? MarketplaceService::API::Api.configurations.get(community_id: @current_community.id).data[:main_search] : :keyword
-    location_search_in_use = (feature_enabled?(:location_search) && search_engine == :zappy && main_search == :location)
+    main_search = location_search_available ? MarketplaceService::API::Api.configurations.get(community_id: @current_community.id).data[:main_search] : :keyword
+    location_search_in_use = location_search_available && main_search == :location
 
     search_result = find_listings(params, per_page, compact_filter_params, includes.to_set)
 
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -52,7 +52,7 @@ class HomepageController < ApplicationController
       end
 
     main_search = location_search_available ? MarketplaceService::API::Api.configurations.get(community_id: @current_community.id).data[:main_search] : :keyword
-    location_search_in_use = location_search_available && main_search == :location
+    location_search_in_use = main_search == :location
 
     search_result = find_listings(params, per_page, compact_filter_params, includes.to_set)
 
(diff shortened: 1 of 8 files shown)


app/views/homepage/index.haml
```diff
@@ -124,7 +128,7 @@
           - if @view_type.eql?("grid")
             %div{class: @category_menu_enabled ? "home-fluid-thumbnail-grid-narrow" : "home-fluid-thumbnail-grid-wide"}
               .home-fluid-thumbnail-grid
-                = render :partial => "grid_item", :collection => @listings, :as => :listing, locals:{show_distance: show_distance}
+                = render :partial => "grid_item", :collection => @listings, :as => :listing, locals:{show_distance: location_search_in_use}
```
Gnito: "locals:{" is changed to "locals: {" in another branch which probably goes to master first (so, some conflicts ahead)

app/views/homepage/_location_bar.haml
```diff
@@ -1,7 +1,3 @@
--content_for(:location_search) do
-  - if(feature_enabled?(:location_search) && search_engine == :zappy)
-    = javascript_include_tag "https://maps.google.com/maps/api/js?sensor=true&libraries=places"
```
rap1ds: This is good :+1: Definitely we shouldn't require the map component ad hoc in partials :+1:

app/controllers/homepage_controller.rb
```diff
@@ -51,21 +51,23 @@ def index
         raise ArgumentError.new("Unknown view_type #{@view_type}")
       end
 
-    search_result = find_listings(params, per_page, compact_filter_params, includes.to_set)
+    main_search = (feature_enabled?(:location_search) && search_engine == :zappy) ? MarketplaceService::API::Api.configurations.get(community_id: @current_community.id).data[:main_search] : :keyword
+    location_search_in_use = (feature_enabled?(:location_search) && search_engine == :zappy && main_search == :location)
```
rap1ds: These two lines seem to have a lot of repetition and the logic feels a bit too complicated. Can we figure out a way to make it simpler?

app/controllers/homepage_controller.rb
```diff
@@ -243,8 +247,8 @@ def shapes
     ListingService::API::Api.shapes
   end
 
-  def location_search_params(latlng, search_engine_in_use)
-      if(feature_enabled?(:location_search) && latlng.present? && search_engine_in_use == :zappy)
+  def location_search_params(latlng, search_engine_in_use, main_search)
+      if(feature_enabled?(:location_search) && latlng.present? && search_engine_in_use == :zappy && main_search == :location)
         coordinates = latlng.split(',')
```
rap1ds: Ruby trick:

```
lat, lng = latlng.split(',')
```

app/controllers/homepage_controller.rb
```diff
@@ -243,8 +247,8 @@ def shapes
     ListingService::API::Api.shapes
   end
 
-  def location_search_params(latlng, search_engine_in_use)
-      if(feature_enabled?(:location_search) && latlng.present? && search_engine_in_use == :zappy)
+  def location_search_params(latlng, search_engine_in_use, main_search)
+      if(feature_enabled?(:location_search) && latlng.present? && search_engine_in_use == :zappy && main_search == :location)
```
rap1ds: Maybe we can pass all search parameters to the adapter and let the search engine to decide which once to use?

app/controllers/homepage_controller.rb
```diff
@@ -51,21 +51,21 @@ def index
         raise ArgumentError.new("Unknown view_type #{@view_type}")
       end
 
+    main_search = location_search_available ? MarketplaceService::API::Api.configurations.get(community_id: @current_community.id).data[:main_search] : :keyword
+    location_search_in_use = location_search_available && main_search == :location
```
rap1ds: The `location_search_available` is redundant in this line. The previous line takes that already into account.

rap1ds:
Looks good! Only on minor comment

app/controllers/homepage_controller.rb
Gnito: true :)

PR #1741, opened 2016-02-17
Fix Fastclick to work with autocomplete

Gnito:
Mobile phone tapping on Google Places Autocomplete doesn't work with Fastclick.
There is old PR that would fix this issue, but Fastclick hasn't merged it yet.
https://github.com/ftlabs/fastclick/pull/347


rap1ds:
Had a quick look, didn't test, but to code and comments look good :+1:

PR #1743, opened 2016-02-18
Adjust results map viewport according to search input

mporkola:


app/controllers/homepage_controller.rb
```diff
@@ -58,6 +58,16 @@ def index
 
     shape_name_map = all_shapes.map { |s| [s[:id], s[:name]]}.to_h
 
+    if @view_type == 'map'
+      viewport = if params[:boundingbox]
+        p = params[:boundingbox].split(',')
+        { sw: [p[0], p[1]],
+          ne: [p[2], p[3]] }
```
Gnito: Something like this might help?
```
sw_lat, sw_lng, ne_lat, ne_lng = params[:boundingbox].split(",")
{ sw: [sw_lat, sw_lng], ne: [ne_lat, ne_lng]}
```
or shorter
```
coords = p.each_slice(2).map{ |x,y| [x,y] }
{ sw: coords[0], ne: coords[1] }
```

PR #1747, opened 2016-02-18
Fix viewport placement in map view

mporkola:
If Google doesn't return a bounding box for a place, use center attribute to place viewport.

### DIFF
--- a/app/assets/javascripts/location_search.js
+++ b/app/assets/javascripts/location_search.js
@@ -11,13 +11,17 @@ window.ST = window.ST || {};
     var autocomplete = new window.google.maps.places.Autocomplete(searchInput);
     autocomplete.setTypes(['geocode']);
 
+    boundingboxInput.value = null;
+
     window.google.maps.event.addListener(autocomplete, 'place_changed', function(){
       var place = autocomplete.getPlace();
       if(place != null) {
         if(place.geometry != null) {
           coordinateInput.value = place.geometry.location.toUrlValue();
           statusInput.value = window.google.maps.places.PlacesServiceStatus.OK;
-          boundingboxInput.value = place.geometry.viewport.toUrlValue();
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/location_search.js
```diff
@@ -11,13 +11,17 @@ window.ST = window.ST || {};
...
+
     window.google.maps.event.addListener(autocomplete, 'place_changed', function(){
       var place = autocomplete.getPlace();
       if(place != null) {
         if(place.geometry != null) {
           coordinateInput.value = place.geometry.location.toUrlValue();
           statusInput.value = window.google.maps.places.PlacesServiceStatus.OK;
-          boundingboxInput.value = place.geometry.viewport.toUrlValue();
+          if (place.geometry.viewport) {
+            boundingboxInput.value = place.geometry.viewport.toUrlValue();
```
Gnito: I guess, we need this if statement to the other "place.geometry.viewport.toUrlValue();" too

app/controllers/homepage_controller.rb
```diff
@@ -59,11 +59,14 @@ def index
     shape_name_map = all_shapes.map { |s| [s[:id], s[:name]]}.to_h
 
     if @view_type == 'map'
-      viewport = if params[:boundingbox]
-        sw_lat, sw_lng, ne_lat, ne_lng = params[:boundingbox].split(",")
-        { sw: [sw_lat, sw_lng], ne: [ne_lat, ne_lng]}
+      coords = Maybe(params[:boundingbox]).split(',').or_else
```
Gnito: What does .or_else do? (might be more readable with .or_else(nil)

mporkola: Exactly that, returns `nil`. I agree that this sounds a bit ominous, I'll have a look around what's the convention mostly used in our code.

PR #1753, opened 2016-02-22
Community specific signup

sktoiva:
The purpose of this PR is to limit users to a single community. A feature flag is set for testing purposes to disable the new community checks when signing in and therefore allow rollback if required


app/controllers/int_api/marketplaces_controller.rb
```diff
@@ -64,7 +64,7 @@ def check_email_availability
     ProspectEmail.create(:email => params[:email])
 
     response.status = 200
-    render :json => {:email => email, :available => (Email.email_available?(email))} and return
```
rap1ds: instead of `Email.email_available?(email)` return always `true`

lyyder: Eventually yes when we remove the :new_login feature flag.

app/forms/form.rb
```diff
@@ -18,7 +18,7 @@ class EmailAvailableValidator < ActiveModel::Validator
     def validate(form)
       options[:fields].each do |f|
         email_address = form.send(f)
-        form.errors.add(f, "Email address #{email_address} is not available.") unless Email.email_available?(email_address)
+        form.errors.add(f, "Email address #{email_address} is not available.") unless Email.email_available?(email_address, nil)
```
lyyder: Email availability check to be removed when :new_login feature flag is removed.

lyyder:
The purpose of this PR is to limit users to a single community. A feature flag is set for testing purposes to disable the new community checks when signing in and therefore allow rollback if required.

app/controllers/application_controller.rb
```diff
@@ -333,19 +337,12 @@ def fetch_community_membership
   end
 
   # Before filter to direct a logged-in non-member to join tribe form
```
rap1ds: Comment needs refreshing. We don't have anymore logged-in non-members.

app/controllers/sessions_controller.rb
```diff
@@ -67,21 +64,15 @@ def create
...
-      if session[:return_to]
-        redirect_to session[:return_to]
-        session[:return_to] = nil
-      elsif session[:return_to_content]
-        redirect_to session[:return_to_content]
-        session[:return_to_content] = nil
-      else
-        redirect_to root_path
-      end
+    flash[:notice] = t("layouts.notifications.login_successful", :person_name => view_context.link_to(@current_user.given_name_or_username, person_path(@current_user))).html_safe
```
rap1ds: Use new hash syntax

app/models/person.rb
```diff
@@ -62,14 +62,15 @@ class Person < ActiveRecord::Base
 
   include ErrorsHelper
   include ApplicationHelper
+  include Devise::Models::DatabaseAuthenticatable
```
rap1ds: Add a comment explaining why this is included.

app/models/person.rb
```diff
@@ -232,8 +214,16 @@ def self.username_blacklist
     USERNAME_BLACKLIST
   end
 
-  def self.username_available?(username)
-     !Person.find_by_username(username).present? && !username.in?(USERNAME_BLACKLIST)
+  def self.username_available?(username, community_id)
+    if FeatureFlag.where(community_id: community_id, feature: :new_login, enabled: true).present?
```
rap1ds: This is accessing the model directly. What if you add this to the Service layer? Something like this:

```ruby
    def enabled?(community_id:, feature:)
      Result::Success.new(@feature_flag_store.get(community_id).include?(feature))
    end
```

PR #1755, opened 2016-02-22
Improve installation instructions

rap1ds:
- [X] Remove default username/password from the database.example.yml. The current once are just guesses which are probably never right.
- [X] Use `rake` to create databases
- [X] Move optional Mailcatcher instructions under the critical instructions
- [X] Use `rake secret` to generate secret key
- [X] Add `config.example.yml`

*Review tip:* use [?w=0](https://github.com/sharetribe/sharetribe/pull/1755/files?w=0)

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,15 +188,22 @@ To automatically run unit tests when code is changed, start [Guard](https://gith
 
 Before starting these steps, perform [steps 1-5 from above](#setting-up-the-development-environment).
 
-1. Set environment variables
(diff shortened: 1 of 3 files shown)


gnomet:
nice changes @rap1ds !
Didn't know  the rake command to set up databases, but definitely a better way. Also good idea to add example file for config, makes it probably much more clear what to do to override settings.
Goog to go! :+1:

PR #1759, opened 2016-02-22
Fix infinity scroll when there are lots of custom filters

Gnito:
(patching jquery.pageless.js for our needs)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -389,7 +389,7 @@ module ApplicationHelper
       :totalPages => total_pages,
       :url        => url,
       :loaderMsg  => loader_message,
-      :div1       => target_id
+      :targetDiv  => target_id # extra parameter for jquery.pageless.js patch
     }
 
     content_for :extra_javascript do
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -389,7 +389,7 @@ def pageless(total_pages, target_id, url=nil, loader_message='Loading more resul
       :totalPages => total_pages,
       :url        => url,
       :loaderMsg  => loader_message,
-      :div1       => target_id
```
Gnito: We already had some random extra parameter passed to the jquery.pageless.js

vendor/assets/javascripts/jquery.pageless.js
```diff
@@ -145,13 +145,24 @@
 
   // distance to end of the container
   var distanceToBottom = function () {
-    return (container === window)
-         ? $(document).height()
-         - $container.scrollTop()
-         - $container.height()
-         : $container[0].scrollHeight
-         - $container.scrollTop()
-         - $container.height();
```
Gnito: This has assumed that the _result container_ is at the end of the page. (i.e. it doesn't take into account _result container's_ position or height)

PR #1760, opened 2016-02-23
Add locale to error message

vtorves:
I found and change these files.
hugs


rap1ds:
@vtorves Thanks for the fixes!

Can you remove the Portuguese translations from the Pull Request? All translations (except English) are managed with WebTranslateIt. Here's more info: https://github.com/sharetribe/sharetribe#translation

PR #1761, opened 2016-02-23
Save listing_id to option selections

ovan:
To make indexing option selections easier add the id of the related listing directly to custom_field_option_selections table. This PR:

* Adds a DB column to store the listing_id in custom_field_option_selectsions
* Modifies listing controller so that listing_id is always saved when listing is edited. This is the only code path that creates option selections so we always have it in place from now on.

Another PR will follow that migrates the listing_id into previously created custom_field_option_selections rows after this has been successfully deployed to production.


bladealslayer:
Looks good!

PR #1762, opened 2016-02-23
Populate listing id to previously created option selections

ovan:


bladealslayer:
Looks good. I'd put a short description of why this change is made in the 0b64de7 commit message.

PR #1764, opened 2016-02-23
Add locale to error message

vtorves:
thanks @rap1ds for your code review and attention a my PR.
i fixed according your comented #1760.


rap1ds:
Thanks for the fix @vtorves !

PR #1766, opened 2016-02-24
Implement optional HTTP basic authentication

bladealslayer:
Allows site administrator to enable HTTP basic authentication for all incoming requests. Can be useful for test installations that are otherwise publicly accessible.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,7 +16,8 @@ class ApplicationController < ActionController::Base
   protect_from_forgery
   layout 'application'
 
-  before_filter :check_auth_token,
+  before_filter :check_http_auth,
+    :check_auth_token,
     :fetch_community,
     :fetch_community_plan_expiration_status,
     :perform_redirect,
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -470,6 +471,15 @@ def fetch_translations
     WebTranslateIt.fetch_translations
   end
 
+  def check_http_auth
+    return true unless APP_CONFIG.use_http_auth
```
rap1ds: If the `use_http_auth` is set using environment variable, the value is string.

Because only `nil` and `false` and falsey values, string such as `"false"` is considered truthy.

So before using the `APP_CONFIG` value, you should convert it to boolean, if it's a string.

bladealslayer: @rap1ds Good point. Interestingly, this works at the moment. I wonder if conversion to boolean happens somewhere in the middle by the configurator.

ovan: Nice, missed this.

PR #1767, opened 2016-02-24
Fix bug: Author can not give feedback if starter skipped it

rap1ds:
Steps to reproduce:

1. Buyer: Start a new transaction
2. Seller: Accept it
3: Buyer: Click mark as completed
4. Buyer: Check the "Give feedback" radio button and move on
5. Buyer: Do not write the feedback, instead leave the page
6. Seller: Go and skip feedback
7. Buyer: Try to give feedback

Expected: Buyer can give feedback

Actual: Error is shown, feedback given already


sktoiva:
Damn. Of course like this. :+1:

PR #1768, opened 2016-02-25
Add utm_ parameters to redirects to marketing site

ovan:
When we redirect to an external configured community_not_found site we add utm_ parameters to make the incoming traffic identifiable at receiving end.

When we redirect because the marketplace has been closed:
* utm_source=:attempted address host:
* utm_medium=redirect
* utm_campaign=qc-auto-redirect

When we redirect because the given host doesn't match any marketplace:
* utm_source=:attempted address host:
* utm_medium=redirect
* utm_campaign=na-auto-redirect


rap1ds:
:+1:

PR #1769, opened 2016-02-25
Allow charging commission if commission status is errored

rap1ds:


ovan:
I don't see any problem with this. Right now we don't expose any mechanism to trigger a retry for commissions so this only applies to manual attempts via console. Still, I think the logic is sound and would work in general case. Go ahead.

PR #1770, opened 2016-02-25
Fix turkish font rendering

rap1ds:
This PR fixes an issue with font-rendering in Turkish.

Before:

screen shot 2016-02-25 at 13 58 03
screen shot 2016-02-25 at 13 57 57

After:

screen shot 2016-02-25 at 13 59 58
screen shot 2016-02-25 at 13 59 54

Background: latin-ext subset was missing from the font we load from Google. This issue only occured with Safari, Firefox etc. Google Chrome was able to render characters even before this change, because it support unicode-range and can smartly use the correct font subset.

This will increase the font-size for Safari, Firefox e.g. The current font-size is ~14 kb and this will increase it to ~21 kb per font-weight (we're currently including 400, 600, 700 font weights)

### DIFF
--- a/app/services/paypal_service/api/lookup.rb
+++ b/app/services/paypal_service/api/lookup.rb
@@ -89,7 +89,7 @@ module PaypalService::API
         return log_and_return(Result::Error.new("Payment is not in :completed state. State was: #{payment[:payment_status]}."))
       end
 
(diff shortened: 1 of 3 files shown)


rap1ds:
If you see "Fix syntax error" commit in this branch, it doesn't belong here. It's in master branch, but Github for some reason shows it here also.

PR #1771, opened 2016-02-25
Encode custom fields to search-api search requests

ovan:
Encode custom field filter parameters and include them in search requests going to search-api. The encoding uses a "minilanguage" to include the custom field id, type and values like this:

Numeric ranges:
```
custom[13]=double:10.2:10.5
````

Option selections (or & and correspond to dropdowns and checkboxes):

```
custom[14]=opt:or:1,3,12
custom[15]=opt:and:60,61,62
```


mporkola:
good! let's test still :+1:

PR #1776, opened 2016-02-25
Touch-action or Fastclick. Remove placeholder plugin

Gnito:
Cleaning some old codes.
- If touch-action is supported => No need for Fastclick
- Placeholder attribute is widely supported nowadays. Do we need extra js?
(i.e. removed https://github.com/mathiasbynens/jquery-placeholder )


Gnito:
- We have quite few input fields that actually use placeholder texts. Most of them are in admin panel.
- I think there's no need to initialize FastClick library if there are touch-action put in place (so, I put it in place).
- I also thought of making FastClick as lazy loaded script, but I'm not sure if it's worth the trouble at this point.

Gnito:
IE9 support for placeholder attributes is removed by removing this polyfill library. But then again, we are not supporting IE9 anymore.

app/views/layouts/application.haml
```diff
@@ -67,8 +67,11 @@
 
   :javascript
     $(function() {
-      FastClick.attach(document.body, { excludeNode: '^pac-'} );
-      $('input, textarea').placeholder();
+      if ('touchAction' in document.body.style) {
+        document.body.style.touchAction = 'manipulation';
+      } else {
+        window.FastClick.attach(document.body, { excludeNode: '^pac-'} );
```
mporkola: out of curiosity, what's `excludeNode: '^pac-'`? I see it's not new, but anyways.

Gnito: FastClick & Google Autocomplete don't want to be in the same page. We had to patch FastClick while making Location search.

PR #1778, opened 2016-02-26
Distance precision changes

Gnito:
This should follow the rules:
if distance is less than 100 meters: "< 0.1 km"
if distance is less than 1 km: "0.6 km" (one significant number)
else we use just 2 significant numbers: 2 543 km => "2 500 km" and 9.63 km => "9.6 km"

### DIFF
--- a/app/views/homepage/_grid_item.haml
+++ b/app/views/homepage/_grid_item.haml
@@ -19,7 +19,8 @@
           .home-fluid-thumbnail-grid-details
             = link_to(person_path(id: listing.author.username), class: "home-fluid-thumbnail-grid-details-author-name", title: name) do
               = name
-            - dist = number_with_delimiter(distance.round(2), locale: locale)
+            - precision = (distance < 1) ? 1 : 2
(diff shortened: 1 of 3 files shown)


app/views/homepage/_grid_item.haml
```diff
@@ -19,7 +19,8 @@
           .home-fluid-thumbnail-grid-details
             = link_to(person_path(id: listing.author.username), class: "home-fluid-thumbnail-grid-details-author-name", title: name) do
               = name
-            - dist = number_with_delimiter(distance.round(2), locale: locale)
+            - dist_w_decimal_precision = (distance < 1) ? number_with_precision(distance, precision: 1, significant: true, locale: locale) : number_with_precision(distance, precision: 2, significant: true, locale: locale)
```
Gnito: I'll change this:
precision = (distance < 1) ? 1 : 2
dist = (distance < 0.1) ? "< 0.1" : number_with_precision(distance, precision: precision, significant: true, locale: locale)

mporkola:
Could the logic be extracted to a helper function? No duplication, and it might be easier to read with the different structure it allows.

PR #1781, opened 2016-02-29
Add handlers for migrating old passwords into new

sktoiva:
Moves old passwords to legacy_encrypted_password field.
Adds valid_password? functionality to people, that will authenticate
against legacy_encrypted_password field if present, hash the entered
password again when valid and store it into encrypted_password and
delete the old legacy password.
Removes legacy asi_encryptor.


bladealslayer:
Looking good!

rap1ds:
Remember to add CHANGELOG entry

app/models/person.rb
```diff
@@ -632,4 +627,32 @@ def followed_people_by_id
   def self.members_of(community)
     joins(:communities).where("communities.id" => community.id)
   end
+
+  def valid_password?(password)
+    if self.legacy_encrypted_password.present?
+      if digest(password, self.password_salt).casecmp(self.legacy_encrypted_password)
+        self.password = password
+        self.legacy_encrypted_password = nil
```
rap1ds: Maybe set salt to NULL?

PR #1785, opened 2016-03-01
Emphasize steps to reproduce

rap1ds:


mporkola:
Good change, communicates the focus better

PR #1789, opened 2016-03-01
Community specific user data migrations

lyyder:
Data migrations related to making people community specific.


db/migrate/20160301141911_duplicate_peoply_by_community_membership.rb
```diff
@@ -0,0 +1,113 @@
...
+        facebook_id,
+        authentication_token,
+        community_updates_last_sent_at,
+        min_days_between_community_updates,
+        is_organization,
+        organization_name,
+        deleted,
+        cloned_from)
+        (SELECT
+          #{quote(UUID.timestamp_create.to_s22)},
```
rap1ds: Hmm... this is now timestamp based UUID. If the migrations runs quick enough, then this may potentially return two identical UUIDs?

db/migrate/20160301141911_duplicate_peoply_by_community_membership.rb
```diff
@@ -0,0 +1,113 @@
+class DuplicatePeoplyByCommunityMembership < ActiveRecord::Migration
```
sktoiva: Class name `Peoply` -> `People`

db/migrate/20160301141911_duplicate_peoply_by_community_membership.rb
sktoiva: The resolution is on nanosecond level so in practice there shouldn't be overlaps.

db/migrate/20160301141911_duplicate_people_by_community_membership.rb
```diff
@@ -0,0 +1,113 @@
...
+        facebook_id,
+        authentication_token,
+        community_updates_last_sent_at,
+        min_days_between_community_updates,
+        is_organization,
+        organization_name,
+        deleted,
+        cloned_from)
+        (SELECT
+          #{quote(UUID.timestamp_create.to_s22)},
```
rap1ds: We should rebase and take #1825 and use the `SecureRandom`

db/migrate/20160301140542_add_cloned_from_and_community_id_to_people.rb
```diff
@@ -0,0 +1,13 @@
+class AddClonedFromAndCommunityIdToPeople < ActiveRecord::Migration
+  def up
+    add_column :people, :cloned_from, :string, { limit: 22 }
+    add_index :people, :cloned_from
+    add_column :people, :community_id, :integer
```
rap1ds: I would add this right after `id`

`add_column :people, :community_id, :integer, after: :id`

db/migrate/20160301141911_duplicate_people_by_community_membership.rb
```diff
@@ -0,0 +1,113 @@
...
+               LEFT OUTER JOIN community_memberships AS cm
+               ON p.id = cm.person_id
+               GROUP BY p.id
+               HAVING count(cm.person_id) > 1
+             ")
+    people.each do |person_id|
+      memberships = select_all("
+        SELECT * FROM community_memberships WHERE person_id = #{quote(person_id)}
+      ").to_ary
+      unless memberships.empty?
```
rap1ds: There's a change to optimize and simplify this:

If you change

```
SELECT p.id
FROM people AS p
LEFT OUTER JOIN community_memberships AS cm
ON p.id = cm.person_id
GROUP BY p.id
HAVING count(cm.person_id) > 1
```

to

```
SELECT p.id, count(cm.person_id)
FROM people AS p
LEFT OUTER JOIN community_memberships AS cm
ON p.id = cm.person_id
GROUP BY p.id
HAVING count(cm.person_id) > 1
```

...then we already know how many new people we need to insert and can skip the `SELECT * FROM community_memberships WHERE person_id = #{quote(person_id)}` query.

db/migrate/20160301140542_add_cloned_from_and_community_id_to_people.rb
rap1ds: In addition, this shouldn't allow NULL values, so we can add `null: false`

db/migrate/20160301140542_add_cloned_from_and_community_id_to_people.rb
```diff
@@ -0,0 +1,13 @@
+class AddClonedFromAndCommunityIdToPeople < ActiveRecord::Migration
+  def up
+    add_column :people, :cloned_from, :string, { limit: 22 }
```
rap1ds: This shouldn't allow NULL values, so we can add null: false

db/migrate/20160301140542_add_cloned_from_and_community_id_to_people.rb
lyyder: Currently our migrations run like 
duplicate people -> updated community_memberships table -> insert community_id value to people table
so we need to allow NULL values, or merge those migrations into one

db/migrate/20160301141911_duplicate_people_by_community_membership.rb
lyyder: Nice!

db/migrate/20160308143022_update_feedback_author_id_based_on_cloned_from.rb
```diff
@@ -0,0 +1,25 @@
...
+  def up
+    ActiveRecord::Base.transaction do
+      cloned_people = select_all("
+         SELECT id, cloned_from, community_id FROM people WHERE cloned_from IS NOT NULL
+      ").to_ary
+      cloned_people.each { |p|
+        execute("
+          UPDATE feedbacks SET author_id = #{quote(p['id'])}
+          WHERE community_id = #{quote(p['community_id'])} AND author_id = #{quote(p['cloned_from'])}
+        ")
```
rap1ds: This is now looping through all cloned users. I'm wondering, could we implement this with only one SQL query?

```sql
UPDATE feedbacks
LEFT JOIN people ON people.cloned_from = feedbacks.author_id AS cloned_people
SET feedbacks.author_id = cloned_people.id
WHERE feedbacks.author_id = cloned_people.cloned_from AND feedbacks.community_id = cloned_people.community_id
```

PR #1790, opened 2016-03-01
Rubocop fixes

sktoiva:


rap1ds:
First two commits look good. Nice to get rid of the hacky `become_true` thing :+1: 

I added one comment to the last commit. Could you check all the places where you use `Maybe` and see if `first_or_create` would do the trick instead?

PR #1793, opened 2016-03-02
Redirect to HTTPS before requesting HTTP basic authentication

bladealslayer:
This is possibly an additional redirect that only happens if basic auth is enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ This file follows the best practices from [keepachangelog.com](http://keepachang
 
 ## [Unreleased]
 
+### Security
+
+- Redirect to HTTPS (if configured) before requesting HTTP basic authentication: [#1793](https://github.com/sharetribe/sharetribe/pull/1793)
+
 ## [5.5.0] - 2016-03-02
 
 ### Changed
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -471,10 +471,21 @@ class ApplicationController < ActionController::Base
     WebTranslateIt.fetch_translations
   end
 
+  def redirect_https?
+    always_use_ssl = APP_CONFIG.always_use_ssl.to_s.downcase == "true"
+    !request.ssl? && always_use_ssl
+  end
+
+  def redirect_https!
+    redirect_to protocol: "https://", status: :moved_permanently
+  end
+
   def check_http_auth
     return true unless APP_CONFIG.use_http_auth.to_s.downcase == 'true'
     if authenticate_with_http_basic { |u, p| u == APP_CONFIG.http_auth_username && p == APP_CONFIG.http_auth_password }
       true
+    elsif redirect_https?
+      redirect_https!
     else
       request_http_basic_authentication
     end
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ This file follows the best practices from [keepachangelog.com](http://keepachang
 
 ## [Unreleased]
 
+- Redirect to HTTPS (if configured) before requesting HTTP basic authentication: [#1793](https://github.com/sharetribe/sharetribe/pull/1793)
```
rap1ds: Could you add the `### Security` header here?

app/controllers/application_controller.rb
```diff
@@ -471,11 +471,23 @@ def fetch_translations
     WebTranslateIt.fetch_translations
   end
 
+  def redirect_https?
+    always_use_ssl = APP_CONFIG.always_use_ssl.to_s.downcase == "true"
+    return !request.ssl? && always_use_ssl
```
rap1ds: You can remove the `return` here. Last expression in the method is always returned so the Ruby way is not to add `return` in this case.

app/controllers/application_controller.rb
```diff
@@ -471,11 +471,23 @@ def fetch_translations
...
+  end
+
   def check_http_auth
     return true unless APP_CONFIG.use_http_auth.to_s.downcase == 'true'
     if authenticate_with_http_basic { |u, p| u == APP_CONFIG.http_auth_username && p == APP_CONFIG.http_auth_password }
       true
     else
+      if redirect_https?
+        return redirect_https
+      end
```
rap1ds: Might be only matter of taste, but I would remove the `return` in front of the `redirect_https` and put the `request_http_basic_authentication` in an `else` branch. That would make the code a bit more "balanced", imo.

app/controllers/application_controller.rb
```diff
@@ -471,11 +471,23 @@ def fetch_translations
     WebTranslateIt.fetch_translations
   end
 
+  def redirect_https?
+    always_use_ssl = APP_CONFIG.always_use_ssl.to_s.downcase == "true"
+    return !request.ssl? && always_use_ssl
+  end
+
+  def redirect_https
+    redirect_to protocol: "https://", status: :moved_permanently if redirect_https?
```
rap1ds: `if redirect_https?` is redundant here. This has been already checked in the `check_http_auth` method. So it can be removed.

Again, might be a matter of taste, but I'd maybe add an exclamation mark to the method name (`redirect_https!`) to emphasize that the redirect is _really_ happening.

spec/requests/http_basic_auth_spec.rb
```diff
@@ -7,6 +7,16 @@
     APP_CONFIG.http_auth_password = "secret"
   end
 
+  before(:each) do
+    @fqdn = "market.custom.org"
```
rap1ds: That's quite cryptic variable name... What does it mean? Could it be changed to something more clear?

rap1ds:
:+1:

PR #1795, opened 2016-03-03
HOTFIX google analytics script was missing ga.js

Gnito:
I quickly made this change so that production deploy from master doesn't break Google analytics scripts.
However, I will not push this to production tonight since this was quickly reviewed by olli and to me it might be better that someone tomorrow morning checks this with fresh eyes. (There's probably some ways to improve this.)

### DIFF
--- a/app/views/layouts/_google_analytics_script.haml
+++ b/app/views/layouts/_google_analytics_script.haml
@@ -51,13 +51,14 @@
       ['b._trackPageLoadTime']
       );
 
-      (function() {
-      var ga = document.createElement('script'); ga.type = 'text/javascript'; ga.async = true;
-      ga.src = ('https:' == document.location.protocol ? 'https://ssl' : 'http://www') + '.google-analytics.com/ga.js';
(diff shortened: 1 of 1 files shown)


app/views/layouts/_google_analytics_script.haml
```diff
@@ -51,14 +51,15 @@
       ['b._trackPageLoadTime']
       );
 
-      (function() {
-      var ga = document.createElement('script'); ga.type = 'text/javascript'; ga.async = true;
-      ga.src = ('https:' == document.location.protocol ? 'https://ssl' : 'http://www') + '.google-analytics.com/ga.js';
-      var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(ga, s);
-      })();
-
```
Gnito: If @current_community.google_analytics_key is nil, there's no analytics script added => we'll miss all the events.
(There's 2 ga tracking code one is for us and one is community specific -> both cases need to have this script put in place.)

PR #1802, opened 2016-03-08
Refactor and fix radiobuttons

rap1ds:
screen shot 2016-03-08 at 16 30 24

The radio buttons had some issues: A customer reported that if he tries to accept a transaction and clicks the buttons back and forth, the wrong action is executed when he sends the form.

Here are steps to reproduce:

1. create a transaction and pay to send request to seller
1. log in as seller
1. open inbox
1. open transaction conversation thread by clicking message "Payment authorized..." then (from customer) after clicking on "Accept request" and "Send reply“ the Order Details window opens. In this window "Accept the request" is automatically chosen but when you switch to „Not this time“ without accepting but immediately switch back to „Accept the request" and then push the „Accept“ button to confirm it, the request gets rejected, but it should get confirmed.

In addition to that, the code needed refactoring. It had several short comings:

1. Relied on global DOM class names, that might be different per each view
2. Wrong abstractions (the icon to use in testimonials view was defined in the Testimonials Model)
3. Over using DRY: ["confirm", "cancel"].each { ... complicated code with a lot of string concatenations } instead of just simply duplicating.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ This file follows the best practices from [keepachangelog.com](http://keepachang
 
 - Update Ruby to 2.2.4 [#1774](https://github.com/sharetribe/sharetribe/pull/1774)
 
+### Fixed
+
+- Wrong action was executed when radio buttons were clicked back and forth [#1802](https://github.com/sharetribe/sharetribe/pull/1802)
+
 ### Security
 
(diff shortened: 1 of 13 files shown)


ovan:
I didn't spot anything to comment. But I don't feel I was the best person to review this so perhaps you should get another look from e.g. @Gnito. The description makes a ton of sense. :+1:

Gnito:
Looks good. 
Although, I didn't test this and there are quite much stuff changed => It's hard to keep in mind the whole picture and different consequences.

PR #1804, opened 2016-03-08
Send GA event: which main search results are fetched.

Gnito:
Report search usage if specific search results are fetched (keyword vs location)

### DIFF
--- a/app/views/homepage/_search_bar.haml
+++ b/app/views/homepage/_search_bar.haml
@@ -1,3 +1,10 @@
+- # report search usage: keyword search results fetched
+- content_for :extra_javascript do
+  :javascript
+    if(document.getElementById('q').value !== "" && document.getElementById('lc') == null) {
+      window.report_analytics_event('user', "main search", "keyword search");
+    }
+
 .search-form-input.search-form-text-field
   = text_field_tag "q", params[:q], :placeholder => (@community_customization && @community_customization.search_placeholder) || t("homepage.index.what_do_you_need")
 .search-form-input
--- a/app/assets/javascripts/location_search.js
+++ b/app/assets/javascripts/location_search.js
@@ -13,6 +13,11 @@ window.ST = window.ST || {};
 
     boundingboxInput.value = null;
 
+    // report search usage: location search results fetched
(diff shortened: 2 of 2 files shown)


app/views/homepage/_search_bar.haml
```diff
@@ -1,3 +1,10 @@
+- # report search usage: keyword search results fetched
+- content_for :extra_javascript do
+  :javascript
+    if(document.getElementById('q').value !== "" && document.getElementById('lc') == null) {
+      window.report_analytics_event('user', "main search", "keyword search");
+    }
+
```
rap1ds: So we have two code branches: A) Using keyword search and B) Using location search.

To me it feels weird that these two logically connected code branches are far away from each other.

Would it make sense to add an `else` branch here, which would send the event if location search is in use?

```js
if ( keyword search into use ) {
  send event about keyword search
} else {
  send event about location search
}

Comment (Gnito) [inline app/views/homepage/_search_bar.haml]: They are used independently and feature flags doesn't reflect that (:location_search, :new_search & with keyword search it doesn't matter if we use search-api or sphinx). The idea was to just collect info to GA about how much we get search queries in different marketplaces. First to figure out location search usage, but later (when both options are visible) we could maybe use that info to differentiate marketplaces.

We could move this to index.haml to cover both cases, but I felt that it's better to reference input-elements inside the partials where they are defined.

**Comment by Gnito** [comment]
@mporkola I'd like you to check this too, since you have been thinking analytics lately.

File: app/views/homepage/_search_bar.haml (hunk 1 above)
Comment (rap1ds) [inline app/views/homepage/_search_bar.haml]: You are already referencing the input element here, aren't you?

```
if( ... && document.getElementById('lc') == null) { ... }
```

**Comment by mporkola** [comment]
I think the event data sent could be made more useful. I'm not sure I agree that "user" is the correct category, for example. I'd suggest something like
```
category: search
action: location search/keyword search
label: <the thing entered in search bar>
```

Also, if you can be bothered to add optional `value` parameter to the `report_analytics_event` function we could use that to send page number to help judge our results relevance by seeing how far the users need to go to find what they're searching, but that's "nice to have".

PR #1805, opened 2016-03-08
Add possible youtube embeds at the end of the listing's description

Gnito:
Embeds at the end of description, the second column

I chose this one to be initial suggestion over embed-videos-inside-description-texts option because:
- It doesn't break the form of description, written and designed by user
- It doesn't loose the last part of the description - push it below the fold (embeds take quite much space)
- It is more future proof: we can get rid of these embeds without breaking users descriptions (sellers have not written descriptions with a mindset that there are going to be embeds inside the text part)
- It looks better than embedding lots of Youtube videos into the beginning of the description area.

embeds_first_last_inside

### DIFF
--- a/app/views/listings/show.haml
+++ b/app/views/listings/show.haml
@@ -71,6 +71,12 @@
         - if @listing.description && !@listing.description.blank?
           - text_with_line_breaks do
             = @listing.description
+          - if(feature_enabled?(:youtube_embeds))
+            - ids = youtube_video_ids(@listing.description)
+            - if(ids.present?)
+              - ids.each do |youtube_id|
+                .listing-description-youtube-wrapper
+                  %iframe.listing-description-youtube-iframe{ title: t("listings.show.youtube_video_player"), width: 640, height:390, src: "//www.youtube.com/embed/#{ youtube_id }", frameborder: 0, allowfullscreen: true}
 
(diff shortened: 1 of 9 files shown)


Gnito:
From Travis:
app/helpers/listings_helper.rb:144:15: C: Use %r around regular expression.
    pattern = /^.*(?:(?:youtu\.be\/|youtu.*v\/|youtu.*embed\/)|youtu.*(?:\?v=|\&v=))([^#\&\?]*).*/
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Do we want to follow this %r rule?
It seems that r forces us to use totally different syntax which means that we can't use nice online tools like https://regexper.com to visualize the regular expression match flow.

app/views/listings/show.haml
```diff
@@ -71,6 +71,12 @@
         - if @listing.description && !@listing.description.blank?
           - text_with_line_breaks do
             = @listing.description
+          - if(feature_enabled?(:youtube_embeds))
+            - ids = youtube_video_ids(@listing.description)
```
sktoiva: Send id's from controller as locals.

PR #1810, opened 2016-03-10
Use Google tag manager instead of Analytics.js for our tracking

mporkola:
We rethought our analytics upgrade a bit, and instead of moving from legacy `ga.js` to shiny `analytics.js`, we'll move to Google Tag Manager instead. It's a tool that the Analytics Guy can use to manage analytics things from an admin panel instead of bothering a developer, to e.g. add a new tracker or define some analytics events.

The diff of `_google_analytics_script.haml` here is quite messy, since I went back and forward by first moving to `analytics.js` and then away from it. Since `analytics.js` was behind feature flag and never actually used, diffing this commit to `afbf0ca` gives you a better view. e.g. like this:

> git diff -b --minimal afbf0ca -- app/views/layouts/_google_analytics_script.haml

[Result here](https://gist.github.com/mporkola/4046bc5cd4169ea10fbc)

### DIFF
--- a/app/views/layouts/_google_analytics_script.haml
+++ b/app/views/layouts/_google_analytics_script.haml
@@ -1,17 +1,15 @@
 - if (APP_CONFIG.use_google_analytics)
   <script type="text/javascript">
+  - if APP_CONFIG.use_google_tag_manager.to_s == "true"
+    <!-- Google Tag Manager -->
+    (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+    '//www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+    })(window,document,'script','dataLayer','#{APP_CONFIG.google_tag_manager_key}');
+    <!-- End Google Tag Manager -->
 
-  - if feature_enabled?(:universal_analytics)
-    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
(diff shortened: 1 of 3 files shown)


app/views/layouts/_google_analytics_script.haml
```diff
@@ -1,17 +1,17 @@
 - if (APP_CONFIG.use_google_analytics)
   <script type="text/javascript">
+  - if APP_CONFIG.use_google_tag_manager
```
rap1ds: Environment variables are always strings, so string "false" will be a truthy value. So one thing you can do is to always convert to string and compare to "true"

For example: APP_CONFIG.use_google_tag_manager.to_s == "true"

app/views/layouts/_google_analytics_script.haml
```diff
@@ -61,18 +61,23 @@
...
 
   function report_analytics_event(category, action, opt_label) {
-  - if feature_enabled?(:universal_analytics)
-    :plain
-      if (typeof ga === 'function'){
-        ga('send', 'event', category, action, opt_label)
-      }
-  - else
+  :plain
+    if (typeof dataLayer != 'undefined') {
```
rap1ds: You could use strict comparison here `!==`

app/views/layouts/_google_analytics_script.haml
```diff
@@ -1,17 +1,17 @@
...
-    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-    })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
+    <!-- Google Tag Manager -->
+    (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+    '//www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+    })(window,document,'script','dataLayer',gtm_key);
```
rap1ds: I guess we don't need to `gtm_key` variable elsewhere, so you could also replace the variable with `'#{APP_CONFIG.google_tag_manager_key}'`. The `gtm_key` seems to be a global variable attached straight to `window` so removing the variable would pollute the global namespace a bit less.

PR #1811, opened 2016-03-10
Try to fix the failing test

rap1ds:
This Pull Request (hopefully!) fixes the randomly failing Cucumber test.

**Background:** When editing the listing, the form is rendered by the server, but it's contains a `hidden` class. When the JavaScript is execute, the script will remove the `hidden` class and show the form to the user.

The Cucumber tests need to wait until the form is loaded and visible, because Cucumber (to my understanding) can't interact with elements that are visible.

The waiting was implemented by adding an additional step `I should see "Listing Title*` before interacting with the form. However, as opposed to what I remembered, Cucumber CAN find elements that are not visible. Due to this, the guard assertion that was waiting the form to be visible didn't work and the test execution continued even though the form wasn't visible.

**Solution:** Implement new step that specifically says to Cucumber to only find the element when it's visible.

**How did I make sure that this really fixes the issue:**

I checked out the `master` branch and ran test file `features/listings/user_edits_his_own_listing.feature ; pz cucumber features/listings/user_edits_his_own_listing.feature` three times. Then I checked out the `fix-broken-test` branch and rerun it three times. Results:

* `master`, 15 specs in total (3 x 5): 2 specs failed
* `fix-broken-test` 15 specs in total (3 x 5): 0 failed

### DIFF
--- a/features/listings/user_edits_his_own_listing.feature
+++ b/features/listings/user_edits_his_own_listing.feature
@@ -18,7 +18,7 @@ Feature: User edits his own listing
     And I log in as "kassi_testperson1"
     When I follow "Hammer"
     And I follow "Edit listing"
-    Then I should see "Listing title*"
+    Then I should see the listing form
     When the "listing_title" field should contain "Hammer"
     And the "description" field should contain "test"
     And I fill in "listing_title" with "Sledgehammer"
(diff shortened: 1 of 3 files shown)


rap1ds:
Apparently my latest push didn't trigger a Travis build, so here's Travis build for the previous one: https://travis-ci.org/sharetribe/sharetribe/builds/116115390

The latest push contained white space fixes, and I squashed that to the "Print PhantomJS version" commit

kpuputti:
Great work!

PR #1813, opened 2016-03-10
Add a small time cap to add some priorities to the CSS compilation

rap1ds:
**Problem:** For some reason that is still unknown to me, Delayed Job doesn't pick the jobs in the ID order if the `priority` and `run_at` values are the same. It picks the Jobs randomly. However, we do want to do some prioritization here so that the marketplaces with most users will be compiled first. 

The weird thing is that I can not reproduce this locally... I'd like to really understand what is going on here.

**Solution:** Add a small time cap to the `run_at` values. The first 100 marketplaces will have time cap of 0 secs, 100-200 1 sec, 200-300 2 secs etc.

### DIFF
--- a/app/services/community_stylesheet_compiler.rb
+++ b/app/services/community_stylesheet_compiler.rb
@@ -16,13 +16,21 @@ module CommunityStylesheetCompiler
   class << self
 
     def compile_all(delayed_opts={})
-      prepare_compile_all do |community|
-        Delayed::Job.enqueue(CompileCustomStylesheetJob.new(community.id), delayed_opts.dup)
+      communities_to_recompile.each_with_index do |community, index|
+        # Do prioritization here
+        #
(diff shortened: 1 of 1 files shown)


ovan:
The reason is that when ordering by priority and run_at you tell DB how you want to order things. DB will follow that but then for rows that have the same values for these two fields the ordering is an implementation detail. This varies probably by exact DB version, amount of data, etc. The bottom line is it's not stable and cannot be relied on.

This solution, while arguably pragmatic, uses run_at for something that it's not intended for. Can't we do this instead with priorities? Like divide marketplaces to 3 groups and set priorities 8, 9 and 10?

rap1ds:
You're absolutely right. The implementation is not very good. I'll fix.

rap1ds:
Closing this. See #1814 and #1815

PR #1815, opened 2016-03-11
Add new job queue "css_compile"

rap1ds:
This Pull Request adds a new job queue "css_compile". That queue is used for long running CSS compilations during the deploy.

Separating CSS compilations to own queue has some benefits:

* The CSS compilations are now completely isolated. They don't interfere the "default" job worker any way.
* CSS compilation workers can be scaled separately
* We can use different priorities for CSS queue

TODO:

- [x] Changelog and upgrade
- [X] Migration that adds "default" queue name for all existing jobs
- [X] Add the "css_compile" queue
- [X] Changes to Procfile

Merge this PR only after #1814

### DIFF
--- a/lib/tasks/sharetribe.rake
+++ b/lib/tasks/sharetribe.rake
@@ -10,8 +10,7 @@ namespace :sharetribe do
   desc "Generates customized CSS stylesheets in the background"
   task :generate_customization_stylesheets => :environment do
     # If preboot in use, give 2 minutes time to load new code
-    delayed_opts = {priority: 10, :run_at => 2.minutes.from_now }
-    CommunityStylesheetCompiler.compile_all(delayed_opts)
+    CommunityStylesheetCompiler.compile_all(run_at: 2.minutes.from_now)
   end
 
   desc "Generates customized CSS stylesheets immediately"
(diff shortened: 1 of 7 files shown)


lib/tasks/sharetribe.rake
```diff
@@ -10,8 +10,7 @@ namespace :sharetribe do
   desc "Generates customized CSS stylesheets in the background"
   task :generate_customization_stylesheets => :environment do
     # If preboot in use, give 2 minutes time to load new code
-    delayed_opts = {priority: 10, :run_at => 2.minutes.from_now }
-    CommunityStylesheetCompiler.compile_all(delayed_opts)
+    CommunityStylesheetCompiler.compile_all(run_at: 2.minutes.from_now)
```
ovan: Not that this is changed here but I'm curious about why do we have the 2 minute wait before we start processing the queue here?

rap1ds: Preboot. The 2 minute delayed is there in order to do our best to sync the compilation with the code changes. We don't want to start compiling new stylesheet if it doesn't work with the old code (which will be still running couple of minutes)

PR #1819, opened 2016-03-14
Add zh-TW as a available unsupported locale

Gnito:


rap1ds:
Should it be supported, if it's 96,66% translated?

PR #1825, opened 2016-03-16
Use SecureRandom for person ids

sktoiva:
Removes np_guid script/plugin and replaces it with
SecureRandom.urlsafe_base64.

The np_guid plugins random creation is broken and the timestamp based
creation and conversion to 22 character strings has a high probability
of generating case insesitively colliding strings. By using SecureRandom
we essentially remove an old library and start using the default Ruby
provided tools.


bladealslayer:
Looking good!

PR #1831, opened 2016-03-18
Google tag manager identification & email confirmation

mporkola:
Send person email to Google Tag Manager for identification, and send event to analytics when admin confirms email

### DIFF
--- a/app/views/analytics/_customer_analytics.haml
+++ b/app/views/analytics/_customer_analytics.haml
@@ -1,0 +1,38 @@
+- if feature_enabled?(:customer_universal_analytics)
+  :plain
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+    })(window,document,'script','//www.google-analytics.com/analytics.js','ga_customer');
+
+    var community_ga_key = "#{@current_community.google_analytics_key}"
+    var community_cookie_domain = "#{PublicSuffix.parse(request.host).domain}"
+    ga_customer('create', community_ga_key, 'auto', {'legacyCookieDomain': community_cookie_domain,   'allowLinker': true});
+    ga_customer('send', 'pageview');
+    var secondary_analytics_in_use = true;
+
+    var customer_report_event = function(category, action, opt_label) {
+      if (typeof ga_customer === 'function'){
+        ga_customer('send', 'event', category, action, opt_label);
+      }
+    };
+- else
+  :plain
+    var _gaq = _gaq || [];
+    _gaq.push(['b._setAccount', '#{@current_community.google_analytics_key}']);
+    _gaq.push(['b._setDomainName', '.#{PublicSuffix.parse(request.host).domain}']);
+    _gaq.push(['b._addIgnoredOrganic', '#{Maybe(@current_community.name(I18n.locale)).gsub("'","").or_else("")}']);
+    _gaq.push(['b._addIgnoredOrganic', '#{@current_community.domain || @current_community.ident}']);
+    _gaq.push(
+      ['b._setAllowLinker', true],
+      ['b._trackPageview'],
+      ['b._trackPageLoadTime']
+    );
+    var secondary_analytics_in_use = true;
+
+    var customer_report_event = function(category, action, opt_label) {
+      var params_array = [category, action, opt_label]
+      if (typeof _gaq != 'undefined') {
+        _gaq.push(['b._trackEvent'].concat(params_array));
+      }
+    };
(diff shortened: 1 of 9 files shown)


app/views/layouts/_google_analytics_script.haml
```diff
@@ -1,6 +1,14 @@
 - if (APP_CONFIG.use_google_analytics)
   <script type="text/javascript">
   - if APP_CONFIG.use_google_tag_manager.to_s == "true"
+    var dataLayer = dataLayer || [];
+    (function(dl){
+    = "var person_emails = [#{person_emails(@current_user).map { |e| "'"+e+"'" }.join(',')}];"
```
mporkola: This is horrible, how could I do it otherwise? I want an array of strings, e.g. `["mikko.porkola@sharetribe.com"]`. With `person_emails(@current_user).to_json` I get
`[&quot;mikko.porkola@sharetribe.com&quot;]` which isnt' exactly doesn't work.

Gnito: You can maybe use .html_safe for the string?

app/views/layouts/_google_analytics_script.haml
```diff
@@ -1,6 +1,14 @@
...
+    if (person_emails) {
```
Gnito:
```
> [] ? true : false
> true
```

Gnito:
_google_analytics_script.haml: To me this file is very confusing. Could it be somehow separated structurally to more readable blocks (based on what analytics are added) or add partials for those different analytic lines.
(It might be easier to understand what happens if this file would only contain 'if' statements choosing correct partials to add.)

app/views/layouts/_google_analytics_script.haml
```diff
@@ -1,6 +1,14 @@
...
+    dl.push({"event": "identify", "id": person_emails});
```
Gnito: Indentation

mporkola: HAML, can't do

Gnito: I thought that it would be possible with those ":plain" sections.

mporkola:
Yep, it's difficult to read. I tried splitting earlier, but that was more complicated than I thought is reasonable, because it's mostly about feature flags. I'll have another look.

app/views/analytics/_customer_analytics.haml
```diff
@@ -0,0 +1,38 @@
...
+    ga_customer('send', 'pageview');
+    var secondary_analytics_in_use = true;
+
+    var customer_report_event = function(category, action, opt_label) {
+      if (typeof ga_customer === 'function'){
+        ga_customer('send', 'event', category, action, opt_label);
+      }
+    }
+- else
+  :javascript
```
Gnito: I'd suggest that this would also be splitted to 2 partials: _customer_legacy_analytics & _customer_universal_analytics
It would make it more clear what to remove when legacy is no longer needed.

app/views/layouts/_google_analytics_script.haml
```diff
@@ -75,19 +33,9 @@
...
-  - else
-    :plain
-      var params_array = [category, action, opt_label]
-      if (typeof _gaq != 'undefined') {
-        if (typeof secondary_analytics_in_use != 'undefined' && secondary_analytics_in_use) {
-          _gaq.push(['b._trackEvent'].concat(params_array));
-        }
-      }
+  :plain
+    if (typeof customer_report_event === 'function') {
```
Gnito: If you are refactoring this more, you could add some comments in these if-elses. So that it would be more clear which analytics contribute to these and removing legacy + features would be easier.

app/views/layouts/_google_analytics_script.haml
```diff
@@ -1,73 +1,31 @@
...
-    ga.src = ('https:' == document.location.protocol ? 'https://ssl' : 'http://www') + '.google-analytics.com/ga.js';
-    var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(ga, s);
-    })();
-
+    :javascript
+      (function() {
+      var ga = document.createElement('script'); ga.type = 'text/javascript'; ga.async = true;
+      ga.src = ('https:' == document.location.protocol ? 'https://ssl' : 'http://www') + '.google-analytics.com/ga.js';
+      var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(ga, s);
+      })();
```
Gnito: I guess, this could be added already in legacy_google_analytics? 
(Isn't this something that needs to be there when we or marketplace admins use legacy analytics.)

app/views/layouts/_google_analytics_script.haml
```diff
@@ -1,73 +1,31 @@
 - if (APP_CONFIG.use_google_analytics)
-  <script type="text/javascript">
```
Gnito: Probably not a biggie, but removing this and using :javascript (instead of :plain) creates several script tags to the page <head>. While debugging this might look a bit weird.
":javascript Surrounds the filtered text with <script>[...]"

PR #1835, opened 2016-03-21
Maintenance mode warning

rap1ds:
This PR adds ability to show a maintenance mode warning. Here's a screenshot:

screen shot 2016-03-21 at 14 35 45

There's a new environment variable `next_maintenance_at` which can be set to the time when the next maintenance happens. I guess environment variable is a good place for this information, since maintenance is very environment specific thing (e.g. some servers might go maintenance while others keep running)

There's a new `initializer` which reads the environment variable and parses it during the app initialization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ This file follows the best practices from [keepachangelog.com](http://keepachang
 
 ### Added
 
-- Add new job queue (css_compile) for css compilations [#1815](https://github.com/sharetribe/sharetribe/pull/1815)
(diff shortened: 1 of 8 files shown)


thomasmalbaux:
Looks great from a visual point of view!

Should we use red though? I remember some discussion about red being only for errors (although it makes sense to drive attention there a lot).

I think the text could be a bit less technical and general. Maybe something like "The website will be offline in x minutes for maintenance" or "We won't be available because of a maintenance in x minutes" or something like that?
"System" sounds quite technical ;-)

lyyder:
Code looks good.

PR #1837, opened 2016-03-22
Remove :new_search and :location_search feat flags

kpuputti:
This enables the new search as well as the location search for all
marketplaces if the config option external_search_in_use is set to true.


rap1ds:
LGTM :+1:

PR #1839, opened 2016-03-22
Cleaner admin/settings path

rap1ds:
Use `/admin/settings` path instead of `admin/communities/1234/settings`

- [X] Add the new paths, using the non-resourceful routes (they are easier to read and understand)
- [X] Add redirection from old URL to new URL
- [X] Change the path name from `settings_admin_community_path` to `admin_settings_path`. Also, generating the path doesn't require `@current_community` parameter anymore, because the community ID is not part of the URL.


mporkola:
The code looks fine, so in case you tested that the routing actually works it's ok for me _b

PR #1845, opened 2016-03-24
Copy profile pictures

rap1ds:
Copy profile images from existing user to the new cloned user.

The code is pretty short and simple:

```
cloned_user.image = cloned_from_user.image
cloned_user.save!
```

... but under the hood Paperclip gem is doing the hard work and copying the image in S3


db/migrate/20160324065708_copy_profile_pictures_for_cloned_users.rb
```diff
@@ -0,0 +1,18 @@
...
+    Person.where("cloned_from IS NOT NULL").pluck(:id).each { |person_id|
+      Delayed::Job.enqueue(CopyProfilePictureJob.new(person_id), priority: 10)
+    }
+  end
+
+  def down
+    # Don't delete the cloned images.
+    # The problem is that we can't know whether the current image is a cloned image
+    # or actually an image that the user uploaded there after the up migration
+    # was run
```
lyyder: The same problem applies to other Hydra migrations too. So the rollback could be implemented on a condition that it can only be run when the community specific signup feature flags are in place.

But as the possible redundant images are not a critical problem and the scheduled nature of this migration makes the rollback more complicated, maybe just for now update the comment on why the ``down`` operation is left out.

Issue #1824, opened 2016-03-15
fontawesome-webfont missing after upgrade


ptzool:
##### Sharetribe Version:

bbf5acaa05c29fa2dbce92c929da97e73a6be85b
##### Environment:

Own server with S3 assets
##### Server mode:

Production
##### Steps To Reproduce:

I upgraded sharetribe to https://github.com/sharetribe/sharetribe/commit/bbf5acaa05c29fa2dbce92c929da97e73a6be85b and after it the system doesn't find fontawesome-webfont on S3. 
![sharetribe](https://cloud.githubusercontent.com/assets/350709/13791029/7fcf3ffe-eaeb-11e5-8946-bd3e0281ecba.png)

I ran the following commands:

```
bundle install
RAILS_ENV=production rake db:migrate
RAILS_ENV=production rake sharetribe:generate_customization_stylesheets_immediately
rake assets:precompile
```

pcm211:
copy your local assets to s3 bucket... partial bucket working is not allowed....in code

in the assets folder (S3) bucket ... add the font-awesome folder from the local server you are working on ....

Its bit of a hack ....but wrks fine

rap1ds:
@ptzool Thanks for reporting.

What was the version from which you updated?

ptzool:
From 5.5.0

rap1ds:
Sometime ago we made a change and got rid of the `asset-sync` gem which was responsible for syncing the static assets to S3. So nowadays the recommended way is to use CDN to serve assets: 

README:

> It is not recommended to serve static assets from a Rails server in production. Instead, you should use a CDN (Content Delivery Network) service, such as Amazon CloudFront. To serve the assets from the CDN service, you need to change the `asset_host` configuration in the the config/config.yml file to point your CDN distribution.

There was also a bug that fixed for 5.5.0 https://github.com/sharetribe/sharetribe/pull/1756

pcm211:
![image](https://cloud.githubusercontent.com/assets/13811638/13978784/bd1cf022-f0f9-11e5-8ea5-894b4a820d0d.png)


the problem is supposedly that its got this wierd versioning thing in the end which is not available even in the local assets .... or some script check parameter which it needs to execute .. preloading .. and fails finally to load ... most of the icons ..... (*.ttf?v=3.0.1 & *.woff?v=3.0.1)

[...]

PR #1850, opened 2016-03-29
Fix broken fonts

### LINKED ISSUE #1849: After customizing cover images on s3, assets, particularly fonts, are not served from s3
##### Sharetribe Version:

07825deab4d58b438641ad5f1a3de1b478babaeb

##### Environment:

Heroku

##### Server mode:

Production

##### Description:

When I customize the cover images using the admin screen, and have my assets served from s3, the `public/assets` directory is not pushed to s3, and in particular the fonts are not.  

To be honest, this may have happened when I set up s3 initially and I may have missed it, but am not sure about that.  (In other words, updating the cover images may not have caused this.)

##### Steps To Reproduce:

1. Install sharetribe on heroku with an s3_bucket_name and s3_upload_bucket_name in the production config.
1. Visit site, everything should be fine.
1. Install a new custom cover image.
1. Visit site, you should see issues like this, where the search, grid, map and list icons are all replaced with squares.

sharetribe-icon-issues

If you look in the devtools, you can see that a number of assets are not available:
   * /assets/modernizr.min-799bf3857c6ec71e5491c8f6e9faa935.js
   * /assets/application-741029e8014a31147bdf6160d7fbc0ee.js
   * /assets/fontawesome-webfont-00022e35cd0adcdcaf29550c7d8b4460.woff?v=3.0.1

There are more

I also tried setting the asset host to the s3 bucket url: https://xxx.s3.amazonaws.com  

Expected result: Assets are pushed to s3

Actual result: Assets are not pushed to s3

##### Additional information

May be related to https://github.com/sharetribe/sharetribe/issues/1824  After reading through that thread, I realize I can manually push the assets to s3, or use a CDN, but am looking for guidance on how to accomplish that without the `asset_sync` gem.

rap1ds:
Problem: The font-awesome fonts are currently broken after customizing the marketplace appearance. The CSS compilation creates a relative path that expects that the font is found in S3 (where the compiled CSS is saved). However, we shouldn't rely that the fonts are there.

Solution: There's no reason why the fonts should be included in the main stylesheet. We can include the font stylesheet in separate link-tag.

Fixes #1849 #1824

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,6 +1,3 @@
-// The fonts file has to be the first one, since there are @imports in the fonts file
-//= require ./fonts
-
 // Vendor
 //= require jquery.nouislider
 //= require jquery.fileupload
--- a/app/assets/stylesheets/font-awesome.min.css.erb
+++ b/app/assets/stylesheets/font-awesome.min.css.erb
@@ -23,10 +23,10 @@
 
 @font-face{
   font-family:'FontAwesome';
-  src:asset-url('fontawesome-webfont.eot?v=3.0.1');
-  src:asset-url('fontawesome-webfont.eot?#iefix&v=3.0.1') format('embedded-opentype'),
-  asset-url('fontawesome-webfont.woff?v=3.0.1') format('woff'),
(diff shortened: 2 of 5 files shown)


rap1ds:
Not happening right now. I'm closing the PR but keeping the branch.

PR #1855, opened 2016-03-29
Add rudimental A/B test solution for onboarding process redesigns

Gnito:
Using feature flags as alternatives and sending that data to GTM
TODO:
- [x] Is this approach enough as initial solution? 
- [x] Add GTM support to ab_test_script partial
- [x] Decide how we should deal with enabling this on-boarding A/B test (Should we add global feature flag context in addition to community specific feature flags?)
- [x] test with gtm

-> We went even more specific direction this case. 
-> We splitted to another branch a related feature: expose feature flags to GTM
-> This won't be merged before we need it

### DIFF
--- a/app/controllers/int_api/marketplaces_controller.rb
+++ b/app/controllers/int_api/marketplaces_controller.rb
@@ -73,7 +73,7 @@ class IntApi::MarketplacesController < ApplicationController
   end
 
   def assign_onboarding_feature_flag(community_id:)
-    if(rand(2) == 1)
+    if(rand < 0.5)
       FeatureFlagService::API::Api.features.enable(community_id: community_id, features: [:onboarding_redesign_v1])
     end
   end
--- a/app/controllers/int_api/marketplaces_controller.rb
+++ b/app/controllers/int_api/marketplaces_controller.rb
@@ -49,6 +49,8 @@ class IntApi::MarketplacesController < ApplicationController
     auth_token = UserService::API::AuthTokens.create_login_token(user[:id])
     url = URLUtils.append_query_param(marketplace[:url], "auth", auth_token[:token])
 
+    assign_onboarding_feature_flag(community_id: marketplace[:id])
+
     # TODO handle error cases with proper response
 
     render status: 201, json: {"marketplace_url" => url, "marketplace_id" => marketplace[:id]}
@@ -68,6 +70,12 @@ class IntApi::MarketplacesController < ApplicationController
   def set_access_control_headers
     # TODO change this to more strict setting when done testing
     headers['Access-Control-Allow-Origin'] = '*'
+  end
+
+  def assign_onboarding_feature_flag(community_id:)
+    if(rand(2) == 1)
+      FeatureFlagService::API::Api.features.enable(community_id: community_id, features: [:onboarding_redesign_v1])
+    end
   end
 
 end


app/controllers/int_api/marketplaces_controller.rb
```diff
@@ -49,6 +49,8 @@ def create
     auth_token = UserService::API::AuthTokens.create_login_token(user[:id])
     url = URLUtils.append_query_param(marketplace[:url], "auth", auth_token[:token])
 
+    ab_test_onboarding(community_id: marketplace[:id], alternative_flags: [:onboarding_new, :onboarding_old])
```
Gnito: This needs a global conditional ```if(featured_enabled(:onboarding_ab_test))``` or we don't merge this branch to master before it's ready.

rap1ds: What if we would just use one flag, `:onboarding`? Marketplaces that have this flag turned on would have the new onboarding flow and the once that don't have this would have the old (=current) getting started page.

app/controllers/int_api/marketplaces_controller.rb
```diff
@@ -74,4 +76,9 @@ def set_access_control_headers
     headers['Access-Control-Allow-Origin'] = '*'
   end
 
+  def ab_test_onboarding(community_id:, alternative_flags:)
+    onboarding_flag = alternative_flags[rand(alternative_flags.size)]
```
rap1ds: *(offtopic Ruby tip: instead of `alternative_flags[rand(alternative_flags.size)]` you can use `alternative_flags.sample`)*

app/views/admin/_ab_test_script.html.erb
```diff
@@ -0,0 +1,4 @@
+<script>
+  window.abTestFlag = <%= get_ab_test_flag(alternative_flags: [:onboarding_new, :onboarding_old]).to_json.html_safe %>;
```
rap1ds: I'd return all the feature flags that the marketplace has and not only limit it to A/B testing flags. I feel that this can be potentially beneficial information, but more importantly, it would make the code a lot simpler: There wouldn't be need for intersecting and that would make the `get_ab_test_flag` would be trivial \o/

app/views/layouts/_marketplace_head.haml
```diff
@@ -88,5 +88,8 @@
 
 = auto_discovery_link_tag(:atom, atom_feed_url )
 
+- if on_admin?
```
rap1ds: I'm not sure if we need to limit this only to the admin panel.

rap1ds:
> Should we add global feature flag context in addition to community specific feature flags?

I think we should merge this only when the onboarding view is ready and not to add yet another feature flag.

app/controllers/int_api/marketplaces_controller.rb
```diff
@@ -74,4 +76,10 @@ def set_access_control_headers
     headers['Access-Control-Allow-Origin'] = '*'
   end
 
+  def ab_test_onboarding(community_id:)
+    if(rand(2) == 1)
+      FeatureFlagService::API::Api.features.enable(community_id: community_id, features: [:onboarding])
```
mporkola: this should probably have a more specific name, such as `onboarding_experiment_1`

app/controllers/int_api/marketplaces_controller.rb
```diff
@@ -74,4 +76,10 @@ def set_access_control_headers
     headers['Access-Control-Allow-Origin'] = '*'
   end
 
+  def ab_test_onboarding(community_id:)
```
mporkola: I think this name should be more verb-y, because it has side-effects? Like `assign_something_test`

app/services/feature_flag_service/store.rb
```diff
@@ -10,6 +10,7 @@ class FeatureFlag
     FLAGS = [
       :export_transactions_as_csv,
       :customer_universal_analytics,
+      :onboarding_flow_v1,
```
rap1ds: flag names do not match: onboarding_flow_v1 vs onboarding_design_v1

app/controllers/int_api/marketplaces_controller.rb
Gnito: This is nicely random. Had to test 5 times on local environment before I got a marketplace which didn't have feature flag set.

app/controllers/int_api/marketplaces_controller.rb
```diff
@@ -74,4 +76,10 @@ def set_access_control_headers
     headers['Access-Control-Allow-Origin'] = '*'
   end
 
+  def assign_onboarding_feature_flag(community_id:)
+    if(rand(2) == 1)
```
Gnito: This is nicely random. I had to test 5 times on local dev environment before I got a marketplace without feature flag.

rap1ds: This is a good solution and works, no need to change.

Another option to do this is to use `rand` without arguments. It returns float between 0..1. So in this case when we want half of the people to get the feature flag we can type:

```ruby
rand < 0.5
```

The `0.5` communicates nicely that it's 50%. If we needed to change this so that 72% of people get the flag, the change would be trivial:

```ruby
rand < 0.72
```

As said, no need to change. Just wanted to share this practice as I have found it pretty useful and clear 😀

Gnito:
This branch will have to wait until we have redesigned onboarding flow ready to be tested.

PR #1856, opened 2016-03-30
Expose feature flags to GTM

Gnito:


app/views/analytics/_google_tag_manager.haml
```diff
@@ -15,7 +21,8 @@
   })(window,document,'script','dataLayer','#{APP_CONFIG.google_tag_manager_key}');
   <!-- End Google Tag Manager -->
 
-  var gtm_datalayer_push = function(data){
+  window.ST = window.ST || {}
```
mporkola: +;

app/views/layouts/application.haml
```diff
@@ -66,7 +64,9 @@
       report_analytics_event.apply(null, #{@analytics_event});
   - if @gtm_datalayer
     :javascript
-      gtm_datalayer_push(#{@gtm_datalayer.to_json});
+      if(window.ST != null && typeof window.ST.gtmPush === 'function') {
```
mporkola: wouldn't `(window.ST && typeof window.ST.gtmPush === 'function')` do the same? applies also to the previous case

mporkola:
Minor comments, go ahead and merge when you're happy with them

app/views/layouts/application.haml
Gnito: @mporkola, In this particular case, it wouldn't matter. The use case for ST is defined clearly as object literal.

However, checking just existence of variable (in if statement) is also taking into account empty string, false, 0, NaN etc. 
So, in general, it is a better practice to just check null & undefined (with "!= null"), when the purpose is to check whether it is unintentionally undefined or explicitly defined to null by devs. 
To me this is more like a best practice thing: check just the existence in 'if' statements and don't assume anything about what might be assigned inside of dynamic variables.

app/views/analytics/_google_tag_manager.haml
Gnito: Good catch. I'll add that.

PR #1857, opened 2016-03-31
Remove legacy analytics

mporkola:
As we now use Google Tag Manager ourselves, we should remove the legacy Google Analytics snippet soon. Moving tracking to GTM takes some work, so I think a deprecation notice is a good idea.

None of our customers user legacy GA accounts that don't support Universal Analytics, so it's safe to move all customers to use it. 

 - [x] Add deprecation notices for Kissmetrics and GA
 - [x] Use universal analytics for all customer tracking

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ This file follows the best practices from [keepachangelog.com](http://keepachang
 
 - Marketplace ID is removed from the Admin Settings URL [#1839](https://github.com/sharetribe/sharetribe/pull/1839)
 
+### Deprecated
+
+- Google Analytics and Kissmetrics tracking snippets are deprecated in favor of Google Tag Manager [#1857](https://github.com/sharetribe/sharetribe/pull/1857)
+
 ### Removed
 
 - Delete duplicated memberships from the database [#1838](https://github.com/sharetribe/sharetribe/pull/1838)
--- a/app/views/analytics/_customer_analytics.haml
+++ b/app/views/analytics/_customer_analytics.haml
@@ -1,17 +1,23 @@
 :plain
+  window.ST = window.ST || {};
+
   (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
   (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
   m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-  })(window,document,'script','//www.google-analytics.com/analytics.js','ga_customer');
+  })(window,document,'script','//www.google-analytics.com/analytics.js','gaCustomer');
 
-  var community_ga_key = "#{@current_community.google_analytics_key}"
-  var community_cookie_domain = "#{PublicSuffix.parse(request.host).domain}"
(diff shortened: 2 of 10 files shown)


app/views/analytics/_customer_analytics.haml
```diff
@@ -1,38 +1,17 @@
-- if feature_enabled?(:customer_universal_analytics)
```
rap1ds: Is the flag still used somewhere? If not, it should be removed from the Flag Store.

app/views/analytics/_customer_analytics.haml
```diff
@@ -1,38 +1,17 @@
...
-    var community_ga_key = "#{@current_community.google_analytics_key}"
-    var community_cookie_domain = "#{PublicSuffix.parse(request.host).domain}"
-    ga_customer('create', community_ga_key, 'auto', {'legacyCookieDomain': community_cookie_domain,   'allowLinker': true});
-    ga_customer('send', 'pageview');
-    var secondary_analytics_in_use = true;
+  var community_ga_key = "#{@current_community.google_analytics_key}"
+  var community_cookie_domain = "#{PublicSuffix.parse(request.host).domain}"
+  ga_customer('create', community_ga_key, 'auto', {'legacyCookieDomain': community_cookie_domain,   'allowLinker': true});
+  ga_customer('send', 'pageview');
+  var secondary_analytics_in_use = true;
```
rap1ds: Do we need to namespace these to `ST`?

config/config.defaults.yml
```diff
@@ -143,16 +143,20 @@ default: &default_settings
...
   # Google Analytics can be used to track traffic on the site
   # If set to true you need to obtain your own API key for analytics
   # and enter it below
   use_google_analytics: false
   google_analytics_key: "enter_your_key_here"
 
-  # You can also use Google Tag Manager
+  # Google Tag Manager can be used to track traffic on the site
+  # If set to true you need to obtain your own container key
+  # and enter it below
```
rap1ds: Would it make sense to log deprecation warning if these are still in use? For example, add a piece of code to some file that is loaded on during the app startup (e.g. application.rb or add a new initializer) and add a warning:

```ruby
ActiveSupport::Deprecation.warn("Kissmetrics is deprecated ... ") if APP_CONFIG.use_kissmetrics
```

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ This file follows the best practices from [keepachangelog.com](http://keepachang
 
 - Marketplace ID is removed from the Admin Settings URL [#1839](https://github.com/sharetribe/sharetribe/pull/1839)
 
+### Deprecated
+
+- Google Analytics and Kissmetrics tracking snippets are deprecated in favor of Google Tag Manager.
```
rap1ds: Good to add this to CHANGELOG :+1: Remember to add a link to the PR

config/config.defaults.yml
mporkola: That's a good idea and implementation suggestion :+1:

app/views/analytics/_customer_analytics.haml
mporkola: removed

app/views/analytics/_customer_analytics.haml
mporkola: Namespacing `ga_customer` is scary, because it's created (and possibly referenced) by asynchronously loaded analytics scripts. I'm not sure of a safe way to do that. Cleaned the other things from global namespace, though

rap1ds:
Looks good to me! I didn't test if this works so remember to test it!

PR #1858, opened 2016-03-31
Log all Person.find calls with username

rap1ds:
The default Rails `Person.find` method takes person ID as a parameter and finds the person by that ID. However, that method has been overriden so that the `Person.find` accepts also `username` as a parameter. This is very confusing and we are about to remove that and not override the method. In addition, in the future we are not able to search user by only username, because the usernames will not be unique anymore, instead, we need to add `community_id` when ever we want to search by username. So that's another reason why this needs to be changed.

In controllers, there are a lot code like this:

```
@person = Person.find(params[:person_id] || params[:id])
```

...and in practice it's impossible to say whether the params contain ID or username. That's why we thought that a good way to find out is to add some logging and deploy that to production and see the result.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -113,7 +113,7 @@ class Person < ActiveRecord::Base
   end
 
   def self.find(username_or_id)
-    by_username = self.find_by_username(username_or_id)
+    by_username = self.find_by(username: username_or_id)
 
     if by_username.nil?
       # Couldn't find by username, trying to find by id
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -112,8 +112,19 @@ class Person < ActiveRecord::Base
     username
   end
 
-  def self.find(username)
-    super(self.find_by_username(username).try(:id) || username)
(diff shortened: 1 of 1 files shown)


app/models/person.rb
```diff
@@ -112,8 +112,19 @@ def to_param
     username
   end
 
-  def self.find(username)
-    super(self.find_by_username(username).try(:id) || username)
+  def self.find(username_or_id)
+    by_username = self.find_by_username(username_or_id)
```
ovan: Where's the find_by_username coming from? I couldn't find it in this file.

rap1ds: Rails magic 😅

Dynamic finders (`.find_by_xxx`) work for each attribute. And you can even use `and` in the method name, i.e. `.find_by_id_and_username(id, username)`. Cool, isn't it? 😜

But seriously, good catch, I'll change it to `find_by`. In Rails 5 dynamic finders will be removed and instead of `find_by_username(username)` we should use `find_by(username: username)`

PR #1861, opened 2016-04-01
Allow disabling Flying Shinx incremental indexing delayed jobs

bladealslayer:


config/config.defaults.yml
```diff
@@ -266,6 +266,9 @@ default: &default_settings
   external_search_url:
   external_search_apikey:
 
+  # Disable Flying Sphinx incremental indexing delayed jobs
+  disable_flying_sphinx_indexing: false
```
ovan: This should be called `disable_thinking_sphinx_indexing`. Flying sphinx is the cloud service but thinking sphinx is the gem whose actions are controlled by this.

mporkola:
I find the negations confusing, we usually have `enable_xxx` and I believe it would be better also in this case.

PR #1863, opened 2016-04-01
Autologout non-members

rap1ds:
This PR cherry-picks the three commits from `community-specific-user-migrations` and adds one, which changes `person.community_id` to `person.communities`, since the `community_id` doesn't exist at this point.

In the first place, I felt that this piece of code should not be in `community-specific-user-migrations`. However, I put it there because I though I need to use the `person.community_id` but `person.communities` works as well atm.


rap1ds:
After this is merged, I will rebase the `community-specific-user-migrations` so that these commit will be removed from that branch.

PR #1865, opened 2016-04-04
Remove Person.find_by_username invocations

lyyder:
With the community specific user migration, user information has to be fetched by using user and community identifiers.


rap1ds:
As soon as the tests pass, this can be merged

PR #1867, opened 2016-04-04
Fix id and username lengths

lyyder:
Set length range for usernames and an exact length for an ID for
fetching people resources.


lyyder:
This fixes id to username redirect in a cases where user has an ID containing no upper case letters.

PR #1868, opened 2016-04-04
Change people param

lyyder:
- Change people identifier to username (`routes.rb` and all calls to `person_path`).
 - Fix length requirements for people id and username parameters (`routes.rb`)
 - Add a redirect for fetching people with id (`routes.rb`)
 - Fetch a person by username and community (`people_controller.rb`)

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -21,7 +21,6 @@ class PeopleController < Devise::RegistrationsController
   def show
     @person = Person.find_by_username_and_community_id!(params[:username], @current_community.id)
     raise PersonDeleted if @person.deleted?
-    PersonViewUtils.ensure_person_belongs_to_community!(@person, @current_community)
 
     redirect_to root and return if @current_community.private? && !@current_user
     @selected_tribe_navi_tab = "members"
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -350,11 +350,6 @@ class PeopleController < Devise::RegistrationsController
     params[:closed] && params[:closed].eql?("true")
   end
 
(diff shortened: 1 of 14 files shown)


config/routes.rb
```diff
@@ -292,21 +292,19 @@
       get "/signup" => "people#new", :as => :sign_up
       get '/people/auth/:provider/setup' => 'sessions#facebook_setup' #needed for devise setup phase hook to work
 
-      resources :people, :path => "", :only => :show, :constraints => { :id => /[_a-z0-9]+/ }
+      resources :people, param: :username, :path => "", :only => :show, :constraints => { :username => /[_a-z0-9]{3,20}/ }
```
kpuputti: Should these regexes have form `/^ ... $/` to match the full argument?

lyyder: Good question. And the answer is no.
From the Rails [documentation](http://guides.rubyonrails.org/routing.html#segment-constraints):

> :constraints takes regular expressions with the restriction that regexp anchors can't be used.
and
> However, note that you don't need to use anchors because all routes are anchored at the start.

app/controllers/people_controller.rb
```diff
@@ -19,7 +19,7 @@ class PersonDeleted < StandardError; end
   helper_method :show_closed?
 
   def show
-    @person = Person.find(params[:person_id] || params[:id])
+    @person = Person.find_by_username_and_community_id!(params[:username], @current_community.id)
     raise PersonDeleted if @person.deleted?
     PersonViewUtils.ensure_person_belongs_to_community!(@person, @current_community)
```
rap1ds: This can be now removed because `Person.find_by_username_and_community_id!` already ensures that the person belongs to community.

PR #1869, opened 2016-04-04
Fix language fallback handling

mporkola:


rap1ds:
Code looks good! Remember to test that it works!

PR #1870, opened 2016-04-04
Remove ensure_authorized method

lyyder:
Remove ensure_authorized method from application_controller as it's
obsoleted by the use of @current_user variable.

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -6,6 +6,8 @@ class MessagesController < ApplicationController
     controller.ensure_logged_in t("layouts.notifications.you_must_log_in_to_send_a_message")
   end
 
+  ensure_can_access_person :person_id
+
   def create
     unless is_participant?(@current_user, params[:message][:conversation_id])
       flash[:error] = t("layouts.notifications.you_are_not_authorized_to_do_this")
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -6,10 +6,6 @@ class MessagesController < ApplicationController
     controller.ensure_logged_in t("layouts.notifications.you_must_log_in_to_send_a_message")
   end
 
-  before_filter do |controller|
-    controller.ensure_authorized t("layouts.notifications.you_are_not_authorized_to_do_this")
-  end
-
   def create
     unless is_participant?(@current_user, params[:message][:conversation_id])
       flash[:error] = t("layouts.notifications.you_are_not_authorized_to_do_this")
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -4,6 +4,8 @@ class PeopleController < Devise::RegistrationsController
   skip_before_filter :verify_authenticity_token, :only => [:creates]
   skip_before_filter :require_no_authentication, :only => [:new]
 
+  ensure_can_access_person :id, only: [:update, :destroy]
+
   before_filter :ensure_is_admin, :only => [ :activate, :deactivate ]
 
   skip_filter :check_email_confirmation, :only => [ :update]
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -4,10 +4,6 @@ class PeopleController < Devise::RegistrationsController
   skip_before_filter :verify_authenticity_token, :only => [:creates]
   skip_before_filter :require_no_authentication, :only => [:new]
 
-  before_filter :only => [ :update, :destroy ] do |controller|
-    controller.ensure_authorized t("layouts.notifications.you_are_not_authorized_to_view_this_content")
-  end
-
   before_filter :ensure_is_admin, :only => [ :activate, :deactivate ]
 
   skip_filter :check_email_confirmation, :only => [ :update]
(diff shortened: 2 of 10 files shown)


app/controllers/messages_controller.rb
```diff
@@ -6,10 +6,6 @@ class MessagesController < ApplicationController
     controller.ensure_logged_in t("layouts.notifications.you_must_log_in_to_send_a_message")
   end
 
-  before_filter do |controller|
-    controller.ensure_authorized t("layouts.notifications.you_are_not_authorized_to_do_this")
-  end
-
```
ovan: When this filter is now dropped from all these controllers what is now taking care of this action (ensuring only authorized person can view the content and show error message in case of unauthorized access)?

lyyder: In ``MessagesController`` it's the `ìs_participant?``on row 14. In ``PeopleController`` only ``@current_user`` is referenced in the resource methods that the filter used to apply to.

ovan: Ok, seems that in MessagesController the same error message is shown but in PeopleController it's then not. Instead the actions are just silently ignored. Do I interpret this correctly? Is this deliberate?

app/controllers/people_controller.rb
```diff
@@ -218,7 +214,7 @@ def update
       redirect_to :back and return
     end
 
-    @person.set_emails_that_receive_notifications(params[:person][:send_notifications])
+    @current_user.set_emails_that_receive_notifications(params[:person][:send_notifications])
```
ovan: Should there be an if check here now that the filter is removed?

app/controllers/settings_controller.rb
```diff
@@ -16,8 +12,8 @@ def show
 
   def account
     @selected_left_navi_link = "account"
-    @person.emails.build
-    marketplaces = @person.community_memberships
+    @current_user.emails.build
```
ovan: Same here, does this need an if check when filter is removed?

app/controllers/messages_controller.rb
rap1ds: The main problem here is that the URL schema is a bit wrong. For example, in my opinion, instead of /username/settings/, the settings page should live in /settings. That's for example how Facebook does it: www.facebook.com/username is the URL for the profile page (same as we do) but then www.facebook.com/settings is where you can access settings (not www.facebook.com/username/settings). So in the future, I'd like to change the URL schema to that direction. 

So what we currently do with the username in the URL? We don't care. We update the settings of the current_user, even if I send a POST request to /ovan/settings

app/controllers/settings_controller.rb
```diff
@@ -4,9 +4,7 @@ class SettingsController < ApplicationController
     controller.ensure_logged_in t("layouts.notifications.you_must_log_in_to_view_your_settings")
   end
 
-  before_filter :except => :unsubscribe do |controller|
-    controller.ensure_authorized t("layouts.notifications.you_are_not_authorized_to_view_this_content")
-  end
+  ensure_can_access_person :person_id, except: [:unsubscribe]
```
ovan: This is confusing that ensure_can_access_person expects a parameter whose value resolves to *username* and it gets passed param name called `:person_id`. With quick search I see this parameter name being referenced in only one other place so why not just rename the route parameter to avoid further confusion?

app/controllers/messages_controller.rb
```diff
@@ -6,9 +6,7 @@ class MessagesController < ApplicationController
     controller.ensure_logged_in t("layouts.notifications.you_must_log_in_to_send_a_message")
   end
 
-  before_filter do |controller|
-    controller.ensure_authorized t("layouts.notifications.you_are_not_authorized_to_do_this")
-  end
+  ensure_can_access_person :person_id
```
ovan: This is confusing that ensure_can_access_person expects a parameter whose value resolves to *username* and it gets passed param name called `:person_id`. With quick search I see this parameter name being referenced in only one other place so why not just rename the route parameter to avoid further confusion?

app/controllers/people_controller.rb
rap1ds: Added `ensure_can_access_person`

app/controllers/settings_controller.rb
rap1ds: Hmmm... I guess you mean it's used in only one other place in *this controller*?

I too think that this parameter name should be changed. However, it's not a low hanging fruit. In `routes.rb` the `people` resource contains a lot of nested resources inside of it:

```
resources :people do
  resources :messages
  resources :settings
  resources :listings
  resources :testimonials
  ...
end
```

So if we change the param name from `person_id` to `person_username` by changing the people resource to:

```
resources :people, param: :username do
```

...then it will change the param name for *all the nested resources*, meaning that we would need to go through all the controllers accessing this parameter name, and all the code that creates URLs pointing to these resources (e.g. `person_settings_path(person_id: ... )` needs to be changed to `person_settings_path(person_username: ... )`).

With a quick search I found 60+ places where renaming is needed.

lyyder:
Closing this PR as we are going to take a different direction with this change.

PR #1873, opened 2016-04-05
Remove find by username invocation from listings controller

lyyder:
Change Person fetch so that username and community id are used in ListingsController.

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -34,9 +34,15 @@ class ListingsController < ApplicationController
     respond_to do |format|
       # Keep format.html at top, as order is important for HTTP_ACCEPT headers with '*/*'
       format.html do
-        if request.xhr? && params[:person_id] # AJAX request to load on person's listings for profile view
-          @person = Person.find(params[:person_id])
-          PersonViewUtils.ensure_person_belongs_to_community!(@person, @current_community)
+
+        # Username is passed in person_id parameter for historical reasons.
+        # In the future the actual parameter name should be changed to better
+        # express its purpose. However this is a large change as there are quite a few
+        # resources nested under the people resource.
+        username = params[:person_id]
+
+        if request.xhr? && username # AJAX request to load on person's listings for profile view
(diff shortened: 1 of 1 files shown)


app/controllers/listings_controller.rb
```diff
@@ -34,9 +34,15 @@ def index
...
-          PersonViewUtils.ensure_person_belongs_to_community!(@person, @current_community)
+
+        # Username is passed in person_id parameter for historical reasons.
+        # In the future the actual parameter name should be changed to better
+        # express its purpose. However this is a large change as there are quite a few
+        # resources nested under the people resource.
+        username = params[:person_id]
+
+        if request.xhr? && username # AJAX request to load on person's listings for profile view
+          @person = Person.find_by_username_and_community_id!(username, @current_community.id)
```
sktoiva: We should already start moving away from the magic dynamic `find_by_*` methods and instead use the format that Rails 5 will enforce:
`Person.find_by!(username: username, community_id: @current_community.id)`

lyyder: In general yes.

In this case we are planning to deploy this change before the Hydra data migrations. So actually it's not a magic dynamic method but an implemented one from ``Person.rb`` :
````
  def self.find_by_username_and_community_id(username, community_id)
    Person
      .joins(:community_memberships)
      .where(username: username, community_memberships: { community_id: community_id })
      .first
  end

  def self.find_by_username_and_community_id!(username, community_id)
    person = Person.find_by_username_and_community_id(username, community_id)

    if person
      person
    else
      raise ActiveRecord::RecordNotFound.new("Can not find person with username #{username} and community id #{community_id}")
    end
  end
````

sktoiva: Ah, ok. Somehow it feels bad that there's a similarly named dynamic method as I already did the mistake in assuming that the method refers to that. But I don't have a strong opinion on should this be fixed by renaming the methods somehow or leave it as is.

lyyder: Plan is to remove the method once we run the migrations and introduce the ``community_id`` field to person. Or at latest when the ``community_memberships`` table is dropped.

PR #1874, opened 2016-04-05
Email.find_by_address_and_community_id

rap1ds:
Currently email addresses are globally unique. However, this will change soon, and emails will be unique per community. That's why we can't use `Email.find_by_address` anymore, instead we need to use `Email.find_by_address_and_community_id`.

Some controllers that are related to login/signup or email availability still use `find_by_address`, because these will change in the #1753 PR.

One controller (Amazon Bounce) needs to find all emails by address, so `find_by_address` is changed to `where(address: address)

### DIFF
--- a/app/models/email.rb
+++ b/app/models/email.rb
@@ -45,10 +45,6 @@ class Email < ActiveRecord::Base
     self.save
   end
 
-  def self.confirmed?(email)
-    Email.find_by_address(email).confirmed_at.present?
-  end
-
   # Email already in use for current user or someone else
   def self.email_available?(email)
     !Email.find_by_address(email).present?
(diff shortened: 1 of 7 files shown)


app/models/email.rb
```diff
@@ -63,4 +59,11 @@ def self.email_available_for_user?(user, address)
   def self.send_confirmation(email, community)
     MailCarrier.deliver_later(PersonMailer.email_confirmation(email, community))
   end
+
+  def self.find_by_address_and_community_id(address, community_id)
+    Email
+      .joins(:community_memberships)
```
rap1ds: I guess this will not work without explicitly defining the foreign key (community_memberships doesn't have `email_id`)

PR #1876, opened 2016-04-07
In the map listing bubble, use icon_tag instead of font's classname

### LINKED ISSUE #1871: Multiple Listings from the same geo sensed location showing up but not able to browse
##### Sharetribe Version:

Stable 5.6.0 & Master

##### Environment:

Ubuntu, AWS, S3, RDS

##### Server mode:

Production

##### Description:

When multiple listings are made from a same geo location the listings when seen in map view are seen as the number ; but on the clicking the location icon generated ... it is not possible to browse the multiple listings from the geo place. 

Only one listing is seen 

##### Steps To Reproduce:

Post multiple listings without by automatically generated geo location

Expected result: A forward and back arrow /marker for browsing to see multiple listings posted from the same place

Actual result: Only one result is seen although multiple listings are shown

##### Additional information

Map View

On Clicking : No Arrow for mov etc

rap1ds:
Fixes #1871

Before:

screen shot 2016-04-07 at 09 02 17

After:

screen shot 2016-04-07 at 09 00 19

### DIFF
--- a/app/views/homepage/_listing_bubble_multiple.haml
+++ b/app/views/homepage/_listing_bubble_multiple.haml
@@ -1,7 +1,7 @@
 .bubble-navi-container
   .bubble-navi
     %a
-      %i.bubble-navi-left.ss-navigateleft
+      = icon_tag("directleft", ["bubble-navi-left"])
     .bubble-navi-header
       %span.buble-navi-selected-item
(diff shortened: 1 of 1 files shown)


pcm211:
@rap1ds . Yes it does

PR #1878, opened 2016-04-07
Remove find by username from application controller

lyyder:
Remove the Person.find(username) calls from ApplicationController. This PR aims to apply the lessons learnt from https://github.com/sharetribe/sharetribe/pull/1870.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -49,6 +49,8 @@ class ApplicationController < ActionController::Base
 
   helper_method :root, :logged_in?, :current_user?
 
+  attr_reader :current_user
+
   def redirect_removed_locale
     if params[:locale] && Kassi::Application.config.REMOVED_LOCALES.include?(params[:locale])
       fallback = Kassi::Application.config.REMOVED_LOCALE_FALLBACKS[params[:locale]]
@@ -178,18 +180,6 @@ class ApplicationController < ActionController::Base
     session[:return_to] = request.fullpath
     flash[:warning] = warning_message
     redirect_to login_path and return
-  end
-
-  # A before filter for views that only authorized users can access
-  def ensure_authorized(error_message)
-    if logged_in?
-      @person = Person.find(params[:person_id] || params[:id])
-      return if current_user?(@person)
-    end
-
-    # This is reached only if not authorized
-    flash[:error] = error_message
-    redirect_to root and return
   end
 
   def logged_in?
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -6,9 +6,7 @@ class MessagesController < ApplicationController
     controller.ensure_logged_in t("layouts.notifications.you_must_log_in_to_send_a_message")
   end
 
-  before_filter do |controller|
-    controller.ensure_authorized t("layouts.notifications.you_are_not_authorized_to_do_this")
-  end
+  before_filter EnsureCanAccessPerson.new(:person_id, error_message_key: "layouts.notifications.you_are_not_authorized_to_do_this")
 
   def create
     unless is_participant?(@current_user, params[:message][:conversation_id])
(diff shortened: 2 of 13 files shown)


app/controllers/messages_controller.rb
```diff
@@ -1,3 +1,5 @@
+require 'ensure_can_access_person'
```
rap1ds: This can be removed. Rails autoloading takes care of this.

app/controllers/settings_controller.rb
```diff
@@ -4,63 +4,60 @@ class SettingsController < ApplicationController
...
   end
 
-  before_filter :except => :unsubscribe do |controller|
-    controller.ensure_authorized t("layouts.notifications.you_are_not_authorized_to_view_this_content")
-  end
+  before_filter EnsureCanAccessPerson.new(:person_id, error_message_key: "layouts.notifications.you_are_not_authorized_to_view_this_content"), except: :unsubscribe
 
   def show
+    target_user = Person.find_by_username_and_community_id(params[:person_id], @current_community.id)
+    target_user = add_location_to_person(target_user)
```
rap1ds: This could be changed to `add_location_to_person!(target_user)` and not return the person, instead mutate it in the method.

app/controllers/application_controller.rb
```diff
@@ -49,6 +49,8 @@ def default_url_options
 
   helper_method :root, :logged_in?, :current_user?
 
+  attr_reader :current_user, :current_community
```
rap1ds: `current_community` is not used anywhere, so it could be removed.

app/views/settings/unsubscribe.haml
```diff
@@ -1,8 +1,8 @@
 .centered-section
-  - if @unsubscribe_successful
+  - if unsubscribe_successful
     %h2
       = t("settings.notifications.unsubscribe_succesful")
-    = t("settings.notifications.unsubscribe_info_text", :settings_link => link_to(t("settings.notifications.settings_link"), notifications_person_settings_path(:person_id => @person_to_unsubscribe.id)), :homepage_link => link_to(t("settings.notifications.homepage_link"), root_path) ).html_safe
+    = t("settings.notifications.unsubscribe_info_text", :settings_link => link_to(t("settings.notifications.settings_link"), notifications_person_settings_path(:person_id => target_user.id)), :homepage_link => link_to(t("settings.notifications.homepage_link"), root_path) ).html_safe
```
rap1ds: `target_user.id` should be `target_user.username` shouldn't it?

app/views/settings/account.haml
```diff
@@ -128,7 +128,7 @@
     %p
       = t("settings.account.delete_information_others_involved")
 
-    = form_tag url_for(:controller => :people, :action => :destroy, :id => @current_user.id), :method => :delete do
+    = form_tag url_for(:controller => :people, :action => :destroy, :id => target_user.id), :method => :delete do
```
rap1ds: This should be `target_user.username`

app/controllers/settings_controller.rb
```diff
@@ -4,63 +4,60 @@ class SettingsController < ApplicationController
...
+    target_user = Person.find_by_username_and_community_id(params[:person_id], @current_community.id)
     @selected_left_navi_link = "notifications"
-  end
-
-  def payments
-    @selected_left_navi_link = "payments"
+    render locals: {target_user: target_user}
   end
 
   def unsubscribe
```
ovan: I noticed this method is omitted from the actions that the EnsureCanAccessPerson filter applies to and started to look where is it checked who has authority to do this operation on behalf of a person but can't find any check. Is there any?

rap1ds: Good catch! Actually, to my understanding this method is not used. The `payments` link points to either PayPal controller or to Braintree controller, but never to the Settings controller. So, I think this method should be removed.

ovan: I actually meant the unsubscribe action.

rap1ds: Oh, sorry. Well good catch about the `payments` method anyway! 😀

Line 39 has this: `@person_to_unsubscribe.username == params[:person_id]`. It should take care of this.

PR #1880, opened 2016-04-07
Remove all the code that references to old "Kaapo" figures

rap1ds:
Remove old code that has a Kaapo reference.

TODO:

- [ ] Remove feedback description file from WTI


mporkola:
I would've preferred more Kaapo, but this is also ok

PR #1881, opened 2016-04-07
Remove find by username from person messages controller

lyyder:
Remove find by username person fetch.


rap1ds:
Code looks good :+1: Have you tested this by sending a message to other user by clicking the "Contact" button in the profile? If you have and this works, then ok to merge.

lyyder:
@rap1ds have tested, will merge now.

PR #1884, opened 2016-04-08
On boarding guide backend

ovan:
This design intentionally puts all logic in one place and operates directly on existing AR models. This has the downside that it couples the implementation tightly to domain classes in a non-obvious way. There's couple of reasons to this choice:

* Make it easy to change the wizard logic, add steps, remove steps, or even remote the entire wizard. It's now an experimental feature
* Minimize the impact on controllers; it's just one event call here and there, no need to care what it does.
* There's no good other place to put it or other way to do it. Preferrably this should be coupled with the domain logic but since we don't have the service layer in place for the admin actions to configure marketplace the his domain logic is scattered across several controllers. I don't want to spread the wizard checks across controllers because those are not correct places in any scenario and that would eliminate the easy of change & delete for on boarding wizard logic.

### DIFF
--- a/app/controllers/admin/communities_controller.rb
+++ b/app/controllers/admin/communities_controller.rb
@@ -231,8 +231,10 @@ class Admin::CommunitiesController < ApplicationController
     update(@current_community,
            community_params.merge(stylesheet_needs_recompile: regenerate_css?(params, @current_community)),
            edit_look_and_feel_admin_community_path(@current_community),
-           :edit_look_and_feel) {
+           :edit_look_and_feel) { |community|
       Delayed::Job.enqueue(CompileCustomStylesheetJob.new(@current_community.id), priority: 3)
+      Admin::OnboardingWizard.new(community.id)
+        .update_from_event(:community_updated, community)
     }
   end
 
@@ -343,7 +345,7 @@ class Admin::CommunitiesController < ApplicationController
   def update(model, params, path, action, &block)
     if model.update_attributes(params)
       flash[:notice] = t("layouts.notifications.community_updated")
-      yield if block_given? #on success, call optional block
+      block.call(model) if block_given? #on success, call optional block
       redirect_to path
     else
       flash.now[:error] = t("layouts.notifications.community_update_failed")
--- a/app/controllers/admin/communities_controller.rb
+++ b/app/controllers/admin/communities_controller.rb
@@ -232,7 +232,7 @@ class Admin::CommunitiesController < ApplicationController
            community_params.merge(stylesheet_needs_recompile: regenerate_css?(params, @current_community)),
            edit_look_and_feel_admin_community_path(@current_community),
            :edit_look_and_feel) { |community|
-      Delayed::Job.enqueue(CompileCustomStylesheetJob.new(@current_community.id), priority: 3)
+      Delayed::Job.enqueue(CompileCustomStylesheetJob.new(community.id), priority: 3)
       Admin::OnboardingWizard.new(community.id)
         .update_from_event(:community_updated, community)
     }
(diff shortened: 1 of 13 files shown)


app/controllers/admin/communities_controller.rb
```diff
@@ -231,8 +231,10 @@ def update_look_and_feel
     update(@current_community,
            community_params.merge(stylesheet_needs_recompile: regenerate_css?(params, @current_community)),
            edit_look_and_feel_admin_community_path(@current_community),
-           :edit_look_and_feel) {
+           :edit_look_and_feel) { |community|
       Delayed::Job.enqueue(CompileCustomStylesheetJob.new(@current_community.id), priority: 3)
```
rap1ds: Now that the block gets the `community` parameter, this could be also changed.

ovan: Yeah, makes sense, I'll do it.

app/services/admin/onboarding_wizard.rb
```diff
@@ -0,0 +1,164 @@
...
+    # Get the status as a SetupStatus hash
+    def setup_status
+      to_setup_status(load_setup_steps(@community_id))
+    end
+
+    # Imperative shell. Process the given event_type with *args
+    # arguments. If the event leads to a state change apply it and
+    # return true. Otherwise return false.
+    def update_from_event(event_type, *args)
+      setup_status = to_setup_status(load_setup_steps(@community_id))
```
rap1ds: This seems to shadow the `setup_status` method. Is that intentional?

ovan: Good catch. I renamed something and missed this. But actually I could just get rid of the whole line. Thanks.

app/services/admin/onboarding_wizard.rb
```diff
@@ -0,0 +1,164 @@
...
+      m
+    end
+
+    def update_completed(community_id, status)
+      unless KNOWN_STATUSES.include?(status)
+        raise ArgumentError.new("Unkown status: #{status}")
+      end
+
+      m = load_setup_steps(community_id)
+      m.update(status => true)
```
rap1ds: `status: true` ?

ovan: Nope, status is a dynamic value, I don't want symbol :status.

app/services/admin/onboarding_wizard.rb
```diff
@@ -0,0 +1,164 @@
...
+      if !setup_status[:slogan_and_description] &&
+         community_customizations.all? { |c| c.slogan.present? } &&
+         community_customizations.all? { |c| c.description.present? }
+        :slogan_and_description
+      end
+    end
+
+    def community_updated(setup_status, community)
+      if !setup_status[:cover_photo] &&
+         Maybe(community).map { |c| c.cover_photo_file_name.present? }.or_else(false)
```
rap1ds: Why would the `community` ever be null?

ovan: Not sure but just to be safe. I don't want the on boarding wizard to crash things easily since it's a non-critical experimental feature for now.

app/services/admin/onboarding_wizard.rb
```diff
@@ -0,0 +1,164 @@
...
+
+    private
+
+    def process_event(event_type, setup_status, args)
+      unless EVENT_TYPES.include?(event_type)
+        raise ArgumentError.new("Unkown event type: #{event_type}")
+      end
+
+      # Dispatch to event handler method of same name as event_type
+      method(event_type).call(setup_status, *args)
```
rap1ds: I was thinking, would it make sense to separate the handler methods from other helpers by putting them into their own module or class...?

PR #1893, opened 2016-04-08
Fix cucumber tests

lyyder:


rap1ds:
If the tests pass, this is good to be merged

PR #1895, opened 2016-04-11
Add generic migration helpers for common migration patterns

rap1ds:
This PR adds a generic migration helper that can handle the most common user data migration patterns. The idea is to use this helper to all the places where this type of migration pattern is used (I counted 13 places). The benefit of this generalisation is that once we have verified that this works, we can be much more confident that all 13 migrations work and there are no typos in them.

TODO

- [X] Fix feedback migration so that it doesn't loop
- [X] Implement first draft of the generic helpers
- [x] Mid-review, does this make sense?
- [x] Use these helpers in all 13 migrations where this pattern applies
- [x] Final review


lyyder:
Looking good to me. I think this could be applied to other tests alike.

app/models/person.rb
```diff
@@ -12,6 +12,7 @@
 #  active_days_count                  :integer          default(0)
 #  last_page_load_date                :datetime
 #  test_group_number                  :integer          default(1)
+#  active                             :boolean          default(TRUE)
```
sktoiva: I guess you have new fields in database that these migrations don't touch? Seems a bit odd that there's new fields added to annotations even though they are not present in the migrations themselves.

app/models/person.rb
```diff
@@ -50,7 +51,7 @@
 #  index_people_on_facebook_id_and_community_id  (facebook_id,community_id) UNIQUE
 #  index_people_on_id                            (id)
 #  index_people_on_reset_password_token          (reset_password_token) UNIQUE
-#  index_people_on_username_and_community_id     (username,community_id) UNIQUE
+#  index_people_on_username                      (username)
```
sktoiva: Same here, the migrations don't touch to indexes, though there's a change here. But I guess this also fixes the state the index is.

PR #1899, opened 2016-04-11
Remove livereload

kpuputti:
By default having livereload injected to the page generates errors and
since few are actually using it, there's no need to keep it.


rap1ds:
LGMT

PR #1901, opened 2016-04-11
Remove listing_images migration

lyyder:
Remove `listing_images` migration for two reasons:

 - author_id values in listing images don't always match with the author_id of the corresponding listing as the image might have been added/modified by the community admin
 - the listing image author_id value is never accessed in the application after it is set


rap1ds:
I'd say we don't need to merge this now. #1895 takes care of this.

PR #1902, opened 2016-04-11
Optimize clone people query

sktoiva:


rap1ds:
Code looks nice and clean! How can we test this?

PR #1904, opened 2016-04-12
Faster commnity memberships

rap1ds:
In this PR:

- [X] Include `community_id` in the query that duplicates users based on users' memberships
- [X] Add `community_memberships` table to the long `update_person_ids...` migration that uses the generic migration helper
- [X] Delete the old community memberships migration
- [X] Rename `AddCommunityIdValuesToPeople` to `AddMissingCommunityIdValuesToPeople`
- [X] Rename `AddCommunityIdValuesToEmail` to `AddMissingCommunityIdValuesToEmail`
- [X] Move `update_person_ids...` before `AddMissingCommunityIdValuesToPeople` and `AddMissingCommunityIdValuesToEmail` migrations. Those two migrations rely on `community_memberships` so it needs to be updated first.
- [x] Squash before merge


db/migrate/20160301141911_duplicate_people_by_community_membership.rb
```diff
@@ -1,44 +1,52 @@
...
+            "FROM people AS p",
+            "LEFT OUTER JOIN community_memberships AS cm ON p.id = cm.person_id",
+            "GROUP BY p.id",
+            "HAVING count(cm.person_id) > 1)"
+        ].join(" ")
+      ).group_by { |p|
+        p['id']
+     }.flat_map { |_, people|
+        people.drop(1) # drop the first one (that's the original)
+      }.each_slice(1000) { |batch|
```
sktoiva: Indent is inconsistent.

PR #1909, opened 2016-04-12
Optimize follower relationships migration

lyyder:
Single SQL statments executed in loops have been replaced with mass
statements.

### DIFF
--- a/db/migrate/20160408070080_update_follower_relationships_based_on_cloned_from.rb
+++ b/db/migrate/20160408070080_update_follower_relationships_based_on_cloned_from.rb
@@ -9,9 +9,13 @@ class UpdateFollowerRelationshipsBasedOnClonedFrom < ActiveRecord::Migration
                                     (SELECT DISTINCT cloned_from FROM people
                                     WHERE cloned_from IS NOT NULL); ").to_ary
 
+      insert_values = Array.new
+      delete_ids = Array.new
+
       relationships.each { |r|
         people = expand_person(r['person_id'])
         followers = expand_person(r['follower_id'])
+
         delete =
           people.inject(true) { |del, person|
             follower = followers.find { |f| f['community_id'] == person['community_id'] }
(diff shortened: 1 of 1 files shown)


db/migrate/20160309073823_update_follower_relationships_based_on_cloned_from.rb
```diff
@@ -9,9 +9,13 @@ def up
                                     (SELECT DISTINCT cloned_from FROM people
                                     WHERE cloned_from IS NOT NULL); ").to_ary
 
+      insert_values = Array.new
+      delete_ids = Array.new
```
rap1ds: `[]` looks more Ruby 😉

db/migrate/20160309073823_update_follower_relationships_based_on_cloned_from.rb
```diff
@@ -69,8 +84,17 @@ def expand_person(person_id)
...
   end
 
-  def create_followers_query(person_id, follower_id, created_at, updated_at)
-    "INSERT INTO follower_relationships (person_id, follower_id, created_at, updated_at)
-     VALUES(#{quote(person_id)}, #{quote(follower_id)}, #{quote(created_at.to_s(:db))}, #{quote(updated_at.to_s(:db))})"
+  def create_insert_values(person_id, follower_id, created_at, updated_at)
+    "(#{quote(person_id)}, #{quote(follower_id)}, #{quote(created_at.to_s(:db))}, #{quote(updated_at.to_s(:db))})"
+  end
+
+  def create_insert_statement(insert_values, ignore)
```
rap1ds: As a general comment (no need to change this now), readability of the method signature can be improved by using `opts` Hash. For example, if someone reads this:

```
create_insert_statement(batch, false)
```

...it's a difficult to know what does the last boolean argument do without looking at the method implementation. This improves it a bit:

```
create_insert_statement(batch, ignore: false)
```

In the implementation, you can then change:

```
def create_insert_statement(insert_values, ignore)
  "INSERT#{ignore ? " IGNORE " : " "}INTO ..."
end
```

to

```
def create_insert_statement(insert_values, opts = {})
  "INSERT#{opts[:ignore] ? " IGNORE " : " "}INTO ..."
end
```

rap1ds:
Couple minor comments to the code (no need to change) but otherwise this looks good! I didn't test this though. Seems that this needs rebase.

PR #1910, opened 2016-04-12
Add missing indexies and move indecies to own migrations

rap1ds:


db/migrate/20160408069990_add_cloned_from_and_indecies_for_migrations.rb
```diff
@@ -0,0 +1,22 @@
+class AddClonedFromAndIndeciesForMigrations < ActiveRecord::Migration
+  def change
+    # Add column
+    add_column :people, :cloned_from, :string, { limit: 22 }
+
+    # Add indecies
```
sktoiva: It's indices.

db/migrate/20160408069990_add_cloned_from_and_indecies_for_migrations.rb
```diff
@@ -0,0 +1,22 @@
+class AddClonedFromAndIndeciesForMigrations < ActiveRecord::Migration
+  def change
+    # Add column
+    add_column :people, :cloned_from, :string, { limit: 22 }
```
sktoiva: Should we add the `community_id` here also?

We could also add the columns direcly by sql using `ALTER TABLE` syntax:

```mysql
ALTER TABLE people
ADD COLUMN cloned_from VARCHAR(22),
ADD COLUMN community_id INTEGER;
```

But I'm not sure if that's worth the trouble.

PR #1911, opened 2016-04-13
Add new filter cannot_access_without_confirmation

rap1ds:
The old code (the before filter `cannot_access_without_joining`) redirected user that didn't have accepted membership to `new_tribe_membership_path`, which points to `CommunityMemberships#new`. This method had some logic which checked if the user account was confirmed and if not, it redirected the user to the correct page. When `cannot_access_without_joining` was removed, this logic was also removed. This PR brings it back.


lyyder:
Code looks good.

PR #1912, opened 2016-04-13
Optimize follower relationships migration

lyyder:
Migration logic has been moved from code to SQL statements.

### DIFF
--- a/db/migrate/20160408070080_update_follower_relationships_based_on_cloned_from.rb
+++ b/db/migrate/20160408070080_update_follower_relationships_based_on_cloned_from.rb
@@ -24,6 +24,19 @@ class UpdateFollowerRelationshipsBasedOnClonedFrom < ActiveRecord::Migration
 
   def down
     ActiveRecord::Base.transaction do
+
+      # New relationships created in the migration are converted
+      # into old ones by replacing the cloned person IDs with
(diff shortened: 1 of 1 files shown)


db/migrate/20160408070080_update_follower_relationships_based_on_cloned_from.rb
```diff
@@ -1,101 +1,69 @@
...
+              follower.cloned_from IS NOT NULL
+          UNION ALL
+            SELECT person.cloned_from, follower.id, fr.created_at, fr.updated_at
+            FROM follower_relationships AS fr
+            LEFT JOIN people AS person ON fr.person_id = person.id
+            LEFT JOIN people AS follower ON fr.follower_id = follower.id
+            WHERE
+              person.cloned_from IS NOT NULL AND
+              follower.cloned_from IS NULL
+      ")
```
lyyder: New relationships are replaced in 3 steps:

1. relationships where both person_id and follower_id are from cloned users
2. relationships where person_id is from an original user and follower_id is from a cloned one
3. relationships where person_id is from a cloned user and follower_id is from an original one

sktoiva: This could be an actual comment in the code.

PR #1915, opened 2016-04-14
Clean up Travis scripts

kpuputti:
- Use `set -e` in scripts to stop executing immediately after a failure.
 - Check `$SUITE` only in the first script. If that fails, the whole build is stopped.
 - Combine steps when several SUITEs run the same commands for less copy-paste duplication

### DIFF
--- a/travis/before_install.sh
+++ b/travis/before_install.sh
@@ -1,38 +1,27 @@
 #!/bin/bash
 
+set -e
+
 echo "Running before_install"
-echo "SUITE: ${SUITE}"
 
-"echo 'gem: --no-ri --no-rdoc' > ~/.gemrc"
+# Assert that the given SUITE is a known one
+case "$SUITE" in
+    rspec|rubocop|cucumber)
+        echo "SUITE: ${SUITE}"
+        ;;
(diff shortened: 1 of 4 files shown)


travis/before_install.sh
```diff
@@ -1,38 +1,25 @@
...
-echo "SUITE: ${SUITE}"
 
-"echo 'gem: --no-ri --no-rdoc' > ~/.gemrc"
+# Assert that the given SUITE is a known one
+case "$SUITE" in
+    rspec|rubocop|cucumber)
+        echo "SUITE: ${SUITE}"
+        ;;
+    *)
+        echo -e "Error: empty or unknown SUITE: $SUITE\n"
```
bladealslayer: You can ditch the `-e` and remove the `\n` in the end of the string. By default `echo` prints a newline.

bladealslayer:
Looks great!

travis/before_install.sh
kpuputti: Ah yeah, good catch. Didn't think of it since the line was already there previously and I just changed the content.

PR #1916, opened 2016-04-14
Setup nodejs in travis

kpuputti:


bladealslayer:
Looking good! Do we need to add e.g. `~/.nvm` to the cached directories in `.travis.yml`?

kpuputti:
@bladealslayer Not really sure, nvm comes preinstalled in Travis, so we probably shouldn't cache the directory? I'm not sure what versions of Node.js are already installed there since we have to download the one defined in .nvmrc if it doesn't exist yet. But I didn't notice any long downloads, so I guess we're pretty safe.

PR #1917, opened 2016-04-15
Onboarding wizard guide

Gnito:
TODO:
- [x] basic page change functionality
- [x] add initial translations
- [x] style all pages

Currently, all the backend generated props are with snake_case (which probably looks strange)
Screenshots & copy texts are checked later


app/controllers/styleguide/pages_controller.rb
```diff
@@ -14,10 +14,28 @@ class Styleguide::PagesController < ApplicationController
   private
 
   def data
+    path_parts = request.env['PATH_INFO'].split("/getting_started_guide")
+    has_deep_path = !(path_parts.count == 1 || path_parts == "")
+    path_string = has_deep_path ? path_parts[1] : "";
+
     # This is the props used by the React component.
     @app_props_server_render = {
       helloReduxData: {
```
mporkola: this can be removed now, I think

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import scss from './styles.scss';
```
kpuputti: Should we import the styles as `css` as a convention?

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
```diff
@@ -0,0 +1,48 @@
...
+      </a>
+
+      <h2 className={ scss.title } >{ props.t('title') }</h2>
+
+      <p className={ scss.description } >
+        { props.t('description') }
+      </p>
+
+      <div className={ scss.sloganImageContainer } >
+        <img src={ props.pageData.infoImage } className={ scss.sloganImage } />
```
kpuputti: Is this an image with some slogan text? In any case, there should probably be an alt text here.

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
```diff
@@ -0,0 +1,48 @@
...
+      </div>
+
+      <div className={ scss.infoTextContainer } >
+        <div className={ scss.infoTextIcon } >
+          <i className="ss-info"></i>
+        </div>
+        <div className={ scss.infoTextContent } >
+          { props.t('advice_p1') }
+          <br />
```
kpuputti: Maybe two `<p>` tags instead of a single `<div>` with a `<br/>` in it.

PS: and I know old habits die hard, but there is really no need to have the space before the self-closing tag `/` anymore ;)

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+      <a href={ props.pageData.link }
+        className={ scss.nextButton } >
+        { props.t('add_your_own') }
+      </a>
+    </div>
+  );
+};
+
+export default GuideCoverPhotoPage;
```
kpuputti: Missing newline in the end in several of these files.

client/app/components/onboardingGuide/GuideFilterPage.jsx
```diff
@@ -0,0 +1,49 @@
...
+import scss from './styles.scss';
+
+const GuideFilterPage = (props) => {
+
+  const handleClick = function(e, path) {
+    e.preventDefault();
+    props.changePage(path);
+  }
+
+  return (
```
kpuputti: This might be a subjective/stylistic question, but do we really need a className for every single element?

client/app/components/onboardingGuide/GuideStatusPage.jsx
```diff
@@ -0,0 +1,72 @@
...
+        { props.t('description_p1') }
+      </p>
+      <p className={ scss.description } >
+        { props.t('description_p2') }
+      </p>
+    </div>
+
+  const done_desc_partial =
+    <p className={ scss.description }
+      dangerouslySetInnerHTML={{ __html: props.t('congratulation') }} ></p>
```
kpuputti: Why is `dangerouslySetInnerHTML` used here? If it avoids the extra `<span>` elements within the `<p>` tag, rebasing to the `react-on-rails` branch updates React to 15.0.1 which doesn't generate the extra spans anymore.

client/app/components/onboardingGuide/GuideStatusPage.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+      <ul className={ scss.stepList } >
+        <li className={ scss.stepListItemDone }>
+          <span className={ scss.stepListLink } >
+            <span className={ scss.stepListCheckbox }></span>
+            Create marketplace
+          </span>
+        </li>
+        { Object.keys(onboarding_data).map(function (key) {
+            let stepListItem = onboarding_data[key].status ? scss.stepListItemDone : scss.stepListItem;
```
kpuputti: We should really agree on a convention about when to use var/let/const.

client/app/components/onboardingGuide/GuideStatusPage.jsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import scss from './styles.scss';
+
+const GuideStatusPage = (props) => {
+
```
kpuputti: No PropTypes for this component? Is it even possible to add those to a stateless function component? If not, I'd vote for using classes everywhere with proper prop type annotations.

client/app/components/onboardingGuide/OnboardingGuide.jsx
```diff
@@ -0,0 +1,183 @@
...
+// Get link and title of next recommended onboarding step
+const nextStep = function(data, translate, initialPath) {
+
+  const stepOrder = [
+    'slogan_and_description',
+    'cover_photo',
+    'filter',
+    'paypal',
+    'listing',
+    'invitation'
```
kpuputti: These step keys should probably be inferred from the data itself.

client/app/components/onboardingGuide/OnboardingGuideContainer.jsx
```diff
@@ -0,0 +1,32 @@
...
+const OnbardingGuideContainer = ({ actions, data, railsContext }) => {
+  return (
+    <OnboardingGuide {...{actions, data, railsContext}} />
+  );
+}
+
+OnbardingGuideContainer.propTypes = {
+  actions: PropTypes.object.isRequired,
+  data: PropTypes.object.isRequired,
+  railsContext: PropTypes.object.isRequired
```
kpuputti: Maybe be more precise than `PropTypes.object` in the types?

client/app/components/onboardingGuide/styles.scss
```diff
@@ -0,0 +1,131 @@
...
+}
+
+.sloganImageContainer {
+  margin: 2rem 0;
+  display: block;
+  width: 100%;
+  height: auto;
+  position: relative;
+  overflow: hidden;
+  padding: 46.72% 0 0 0; /* 46.72% = 100 / (w / h) = 100 / (732 / 342) */
```
kpuputti: Can the style loaded calculate this at compile time? Would avoid the explanation comment that can become out of sync.

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
Gnito: alt! 👍

client/app/components/onboardingGuide/styles.scss
```diff
@@ -0,0 +1,131 @@
+.title {
+  font-size: 1.375rem;
+  font-weight: 600;
+  color: #171717;
+  line-height: 1.875rem;
+
```
kpuputti: Lots of values that should potentially be vars. Don't we have means to use vars in the styles atm?

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
Gnito: I had it separated, but with '<p>' there is going to be margin-bottom and that was not something our examples in drive suggested. Since we are still waiting for some improved copy texts - I'd go with this.
P.S. I like that space - and I want to use one extra space after last attribute of opening tags.

client/app/components/onboardingGuide/GuideFilterPage.jsx
Gnito: I really want to make every component independent what comes to styles. So, when we relocate components, they don't need to worry about styles (of parent-component).
It's also annoying when old code defines styles in a form of 
```
.component {
  p {
    color: black;
  }
}
```

That means that all the injected Childs are going to have that  generic `<p>` style as well.
=> I prefer adding className every time there's styles attached to that component.

client/app/components/onboardingGuide/GuideStatusPage.jsx
Gnito: It's due our translation problems. We need to think through how we are going to make interpolation. This project uses translations that are not coming from WTI, which gives us total control over writing those files. So, I added links directly to translation file.
=> translation interpolation problem still exists! It needs to be thought through before we can use WTI.

client/app/components/onboardingGuide/GuideStatusPage.jsx
Gnito: I'll change this. 
- 'var' shouldn't be used. 
- 'const' should be a default and 
- 'let' is for reassignment situations like counters in loops and stuff.
(There was a 'for' loop at some point - but I don't remember was there any real use for 'let' even then. Bad code anyways.)

client/app/components/onboardingGuide/OnboardingGuide.jsx
Gnito: There was long discussions with Olli that who should own the state of these steps. I was originally leaning towards Rails having something to say about next step (since we have separate frontend components instead of SPA), but I decided to go with this direction in the end. So, it's (web) applications job to decide the order, which means hard coded values need to exist somewhere.

client/app/components/onboardingGuide/styles.scss
Gnito: Good question. I don't know if there's anything in css-loader world that would be capable to do it. JS could do it of course, but that's bad.

client/app/components/onboardingGuide/GuideStatusPage.jsx
Gnito: It's possible - static might not be available, but that's just a shorthand to some special property magic. Anyway, I added those PropTypes.

client/app/components/onboardingGuide/OnboardingGuideContainer.jsx
Gnito: At this point I added more specific to data, which is something we construct. 
RailsContext might change with gem updates and actions is

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
Gnito: I thought that (since we are hopefully going to use PostCSS at some point) it's easier to spot files that need updating when they are using correct file extension.

client/app/components/onboardingGuide/OnboardingGuide.jsx
kpuputti: My point was more about not having those hard coded values in multiple places in the frontend. I think this wasn't the only place where those step keys were written down.

client/app/components/onboardingGuide/OnboardingGuideContainer.jsx
kpuputti: I don't understand your comment. What do you mean?

client/app/components/onboardingGuide/styles.scss
Gnito: Apparently, there is. e.g. https://www.npmjs.com/package/postcss-aspect-ratio
But since we have not yet added PostCSS support, I will do it manually. 
(But this is good to remember when we really start developing new design with PostCSS.)

app/view_utils/onboarding_view_utils.rb
```diff
@@ -0,0 +1,28 @@
+module OnboardingViewUtils
```
Gnito: This is a copy-paste from Onboarding popup branch.

app/controllers/admin/getting_started_guide_controller.rb
```diff
@@ -0,0 +1,68 @@
+class Admin::GettingStartedGuideController < ApplicationController
+
+  before_filter :ensure_is_admin
+
+  rescue_from ReactOnRails::PrerenderError do |err|
+    Rails.logger.error(err.message)
+    Rails.logger.error(err.backtrace.join("\n"))
+    redirect_to root_path, flash: { error: "Error prerendering in react_on_rails. See server logs." }
+  end
```
mporkola: We probably don't want to display stacktraces to users?

app/controllers/admin/getting_started_guide_controller.rb
```diff
@@ -0,0 +1,68 @@
+class Admin::GettingStartedGuideController < ApplicationController
+
+  before_filter :ensure_is_admin
+
+  rescue_from ReactOnRails::PrerenderError do |err|
+    Rails.logger.error(err.message)
+    Rails.logger.error(err.backtrace.join("\n"))
+    redirect_to root_path, flash: { error: "Error prerendering in react_on_rails. See server logs." }
```
mporkola: This error message doesn't mean much to a normal user. Prerender error handling is actually probably something we need to think about a bit more, also in my branch.

app/views/admin/getting_started_guide/index.haml
```diff
@@ -0,0 +1,13 @@
...
+    = "-"
+    = t("admin.communities.getting_started.getting_started")
+
+%article.page-content
+  .row
+    .col-3
+      = render :partial => "admin/left_hand_navigation", :locals => { :links => admin_links_for(@current_community) }
+    .col-9
+      - cache ['v1', 'onboarding_guide', props] do
+        = react_component("OnboardingGuideApp", props: props, prerender: true, raise_on_prerender_error: true)
```
mporkola: `prerender` and `raise_on_prerender_error` should probably be set in React on Rails initialiser, so that they'd be in line everywhere.

app/controllers/admin/getting_started_guide_controller.rb
Gnito: Good point.

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
```diff
@@ -0,0 +1,62 @@
+import React, { PropTypes } from 'react';
+import scss from './styles.scss';
```
mporkola: Agreeing with @kpuputti 's earlier comment, should be `css`. The preprocessor we may or may not use shouldn't affect this variable name here.

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+const GuideCoverPhotoPage = (props) => {
+  const handleClick = function handleClick(e, path) {
+    e.preventDefault();
+    props.changePage(path);
+  };
```
mporkola: This here is probably a good candidate to move to HOC or even better, it's own component

client/app/components/onboardingGuide/GuideListingPage.jsx
```diff
@@ -0,0 +1,60 @@
...
+      <div className={scss.sloganImageContainer} >
+        <img src={props.pageData.info_image}
+          className={scss.sloganImage}
+          alt={props.t('info_image_alt')}
+        />
+      </div>
+
+      <div className={scss.infoTextContainer} >
+        <div className={scss.infoTextIcon} >
+          <i className="ss-info"></i>
```
mporkola: Usually we shouldn't use the `ss` icons directly, has this been considered?

client/app/components/onboardingGuide/GuideSloganAndDescriptionPage.jsx
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  return (
+    <div className="container">
+      <a className={scss.backLink}
+        onClick={(e) => handleClick(e, '')}
+        href={props.initialPath}
+      >
+        ‹ {props.t('back_to_todo')}
+      </a>
```
mporkola: As this link is in many components and in most cases `handleClick` is mostly used here, making it it's own component would be a good idea. Maybe even before merging, can't say!

client/app/components/onboardingGuide/OnboardingGuide.jsx
```diff
@@ -0,0 +1,203 @@
...
+    } else if (event.state == null && typeof this.props.data.pathHistoryForward !== 'undefined') {
+      // null state means that page component's root path is reached and
+      // previous page is actually on Rails side - i.e. one step further
+      // Safari fix: if pathHistoryForward is not defined, its initial page load
+      window.history.back();
+    }
+  }
+
+  handlePageChange(path) {
+    this.props.actions.updateGuidePage(path, true);
```
mporkola: I'd prefer a navigation reducer, or something, to own the URL. Component manipulating URL (and having state affected by it?) feels bad, I think this should be done by dispatching actions.

Can be done later, I've been marking these to a list :)

client/app/components/onboardingGuide/GuideListingPage.jsx
Gnito: D'oh. This escaped my todo list. 
We can't use ss-info directly. We should use 'information' which is then mapped in Rails side to ss-pika or font-awesome - but the question is that: I'm not sure does it do that on our setup.

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
Gnito: Own component could make sense (although it would make more sense if we start to build shared components folder). But I think this is a wrong place to wrap it as HOC.

app/views/admin/getting_started_guide/index.haml
Gnito: OK. Let's do it that way.

client/app/components/onboardingGuide/OnboardingGuide.jsx
Gnito: Let's leave this decision to next project. (We need to think about routing a bit more later.)
Currently, flow goes so that: click to link calls this action -> reducer -> redux store -> props changed -> this current *root component* handles window.history related actions.

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
Gnito: And I decided to separate this Link to its own component file.

mporkola:
[selfie-0](https://github.com/thieman/github-selfies/)

PR #1918, opened 2016-04-15
Setup React on Rails project

kpuputti:


README.md
```diff
@@ -100,12 +106,27 @@ Before you get started, the following needs to be installed:
...
+  1. Static Loading of Rails Assets
+  ```bash
+  foreman start -f Procfile.static
+  ```
+
+  1. Creating Assets for Tests
+  ```bash
+  foreman start -f Procfile.spec
+  ```
+
```
ovan: As somebody not familiar with React, hot loading and react-on-rails this section is not very clear. Should I invoke all those commands one time? Every time I change something? Do they start a continuous server processes to do something every time a change is made?

client/README.md
```diff
@@ -0,0 +1,66 @@
+Please see parent directory README.md.
```
ovan: This should be a link.

README.md
```diff
@@ -55,6 +56,11 @@ Before you get started, the following needs to be installed:
   bundle install
   ```
 
+1. Install node modules
+  ```bash
+  npm install
```
ovan: This seems to take a lifetime and quickly looking at the list includes modules that are not in active use. I would suggest pruning this to minimum now and adding later more modules if needed (things like react router).

README.md
ovan: Also, it should be highlighted that standard way like `rails server` is no longer the way to go here.

Gemfile
```diff
@@ -111,6 +111,8 @@ gem 'demoji', "~> 0.0.5"
 gem 'faraday', '~> 0.9.1'
 gem 'faraday_middleware', '~> 0.10.0'
 
+gem "react_on_rails", "~> 5"
```
rap1ds: Maybe a stricter versioning here? `bundle update` may now update react_on_rails to `6`, which may not be desired.

Gemfile
```diff
@@ -153,8 +155,9 @@ group :test do
...
-  gem 'pry', '~> 0.9.12.6'
-  gem 'pry-rails', '~> 0.3.2'
-  gem 'pry-nav', '~> 0.2.3'
-  gem 'pry-stack_explorer', '~> 0.4.9.1'
+  gem 'pry-byebug'
+  gem 'awesome_rails_console', '~> 0.4.0'
 end
+
+
+gem 'therubyracer', platforms: :ruby
```
rap1ds: No version number here. Is that desired? `bundle update` can now update ruby racer to any version.

README.md
rap1ds: I agree, this could be improved a bit:

* Use imperative mood. `s/Starting the development.../Start the development...`
* Would it make sense to pick one of these and move others under "Advanced setup" paragraph and "Testing" paragraph.

client/app/actions/HelloReduxActions.jsx
```diff
@@ -0,0 +1,8 @@
+import * as actionTypes from '../constants/HelloReduxConstants';
```
rap1ds: This file looks like a normal JS file but the file extension says `.jsx`. Is that correct?

ovan:
* This should lead to 5.7 version release with proper release notes mentioning at least that building is now different and running local dev server needs the foreman command.
* Upgrade guide needs instructions what changes are required from a Heroku prod environment (that's currently our officially supported prod environment).

UPGRADE.md
```diff
@@ -25,6 +26,14 @@ A new worker is added to the Procfile to work for the new queue. If you're hosti
 
 This change doesn't require any changes if you are compiling the stylesheets synchronously using the `rake sharetribe:generate_customization_stylesheets_immediately` command during the deployment. However, if you are compiling the stylesheets asynchronously using the `rake sharetribe:generate_customization_stylesheets` command, then you need to make sure that you have at least one worker working for the "css_compile" queue.
 
+
+React on Rails build environment is added with this change. This means that production environment needs to have node set up. With Heroku this can be set with ```heroku buildpacks:set heroku/nodejs```. For other environments - see [npm instructions](https://docs.npmjs.com/getting-started/installing-node), [nvm](https://github.com/creationix/nvm), or [n](https://github.com/tj/n). In addition, production environments should have ```NODE_ENV=production`` set.
```
ovan: not production environment but build environment. Important difference because we do not require node for runtime environment, only for building.

package.json
```diff
@@ -3,5 +3,13 @@
   "license": "MIT",
   "engines": {
     "node": "5.10"
+  },
+  "scripts": {
+    "postinstall": "cd client && npm install",
+    "clean": "rm -f app/assets/webpack/*"
+  },
+  "cacheDirectories": ["client/node_modules"],
```
ovan: Is this an override or addition? If it's an override then do we intentionally skip the current node_modules?

client/README.md
```diff
@@ -1,4 +1,4 @@
-Please see parent directory README.md.
+Please see parent directory [README.md](../README.md).
```
mporkola: Maybe this could be expanded a bit?

PR #1920, opened 2016-04-15
Onboarding top bar

mporkola:
- [x] Show top bar for admins
 - [x] Populate it with live data
 - [x] Put behind A/B testing flag
 - [x] Links to actual steps


app/controllers/application_controller.rb
```diff
@@ -455,6 +458,13 @@ def maintenance_warning
     @minutes_to_maintenance = NextMaintenance.minutes_to(now)
   end
 
+  def get_onboarding_status
+    if @is_current_community_admin
+      #TODO: maybe not with global variable?
+      @onboarding_status = Admin::OnboardingWizard.new(@current_community).setup_status
```
ovan: The OnboardingWizard expects community_id, not a community object as ctor parameter. This works by accident now because ActiveRecord find_by is flexible and seems to know how to interpret an AR object as id in find calls. But that's an accidental implementation detail so this should be changed to pass in id instead of AR object.

app/controllers/application_controller.rb
```diff
@@ -455,6 +458,13 @@ def maintenance_warning
     @minutes_to_maintenance = NextMaintenance.minutes_to(now)
   end
 
+  def get_onboarding_status
+    if @is_current_community_admin
+      #TODO: maybe not with global variable?
```
ovan: This comment feels like a good idea to me. Instead of doing this with filter it's enough if get_onboarding_status is exposed as a method and called in application.haml when topbar is initialized.

client/app/components/OnboardingTopBar/OnboardingTopBar.jsx
```diff
@@ -0,0 +1,98 @@
...
+
+  nextElement() {
+    if (this.next()) {
+      return (
+        <div className={css.nextContainer}>
+          <div className={css.nextLabel}>
+            {this.t('next_step')}:
+          </div>
+          <a href={this.next().link} className={css.nextButton}>
+            <span>{this.next().title}</span>
```
ovan: next() is called 3 times here and afaik has to do the processing every time. Why not capture it in local var at the top level on this function?

client/app/components/OnboardingTopBar/OnboardingTopBar.jsx
```diff
@@ -0,0 +1,98 @@
...
+      };
+    } else if (!this.props.status.listing) {
+      return {
+        title: this.t('add_listing'),
+        link: this.props.guide_root + '/listing',
+      };
+    } else if (!this.props.status.invitation) {
+      return {
+        title: this.t('invite_users'),
+        link: this.props.guide_root + '/invitation',
```
ovan: The link building here skips the Rails so if paths are moved in router they have to manually reflected here. Also, the guide has to get these same links so I'd suggest having just one way to do this and to let Rails build the paths.

app/controllers/admin/getting_started_guide_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  def index
+    render locals: { props: data }
+  end
+
+  private
+
+  def data
+    path_parts = request.env['PATH_INFO'].split("/getting_started_guide")
+    has_deep_path = !(path_parts.count == 1 || path_parts == "")
+    path_string = has_deep_path ? path_parts[1] : "";
```
kpuputti: Could the above two lines be simplified into something like the following?

```ruby
sub_path = path_parts.count == 2 ? path_parts[1] : ""
```

app/views/styleguide/pages/getting_started_guide.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= render "header" %>
+
+<div style="margin: 20px; max-width: 732px;">
```
kpuputti: These inline styles should probably live in some css/scss file.

app/views/layouts/application.haml
```diff
@@ -6,6 +6,11 @@
     .noscript-padding
       -# Noscript content will be positioned here
 
+  - if @current_user && @current_community && @current_user.has_admin_rights_in?(@current_community)
+    - with_feature(:onboarding_redesign_v1) do
+      - props = {translations: t('admin.onboarding.topbar'), status: @onboarding_status, guide_root: getting_started_guide_admin_community_path(id: @current_community.id)}
+      - cache ['v1', 'onboarding_topbar', props] do
+        = react_component("OnboardingTopBar", props: props, prerender: true)
```
kpuputti: Is there a reason why the `raise_on_prerender_error` flag is set in some places, but not e.g. here?

client/app/components/OnboardingTopBar/OnboardingTopBar.jsx
```diff
@@ -0,0 +1,98 @@
+import React, { PropTypes } from 'react';
+
+import css from './OnboardingTopBar.scss';
+
+class OnboardingTopBar extends React.Component {
+
+  t(translationKey) {
+    return this.props.translations[translationKey];
+  }
```
kpuputti: If something like `utils.t(this.props.translations, translationKey)` is too much to type or if `f(translationKey) { return utils.t(this.props.translations, translationKey); }` still seems too much repetition, creating a [Higher Order Component](http://jamesknelson.com/structuring-react-applications-higher-order-components/) that "mixes in" the `this.t` function would be quite a clean solution to share this helper function.

client/app/components/OnboardingTopBar/OnboardingTopBar.jsx
```diff
@@ -0,0 +1,98 @@
+import React, { PropTypes } from 'react';
+
+import css from './OnboardingTopBar.scss';
+
+class OnboardingTopBar extends React.Component {
```
kpuputti: Given the fact that there isn't anything stateful in this component (which is a good thing), there are several methods that take their input from `this.props`. Could those methods be refactored as separate functions, probably within this same file (if not really useful elsewhere)?

Having several clearly pure functions would IMO help a lot when reading through what thas component does and what it depends on. This would also make it easier to port the components as stateless function components in the future if we decide so.

I know the call site of the functions/methods will have more characters, but I'd say being explicit helps in reasoning about the code (sorry for the cliché saying).

app/controllers/admin/getting_started_guide_controller.rb
Gnito: Hmm. this part comes from my branch (and it seems I have refactored an important part away - I had originally root path of component marked as "/", but it caused unnecessary history update.)
I'll change my code to be:
```
    has_sub_path = (path_parts.count == 2 && path_parts[1] != "/")
    sub_path = has_sub_path ? path_parts[1] : "";
```

app/views/styleguide/pages/getting_started_guide.html.erb
Gnito: This also comes from my branch. (which is not yet ready)

client/app/components/OnboardingTopBar/OnboardingTopBar.jsx
```diff
@@ -0,0 +1,98 @@
...
+          {this.nextElement()}
+        </div>
+      </div>
+    );
+  }
+}
+
+OnboardingTopBar.propTypes = {
+  translations: PropTypes.object,
+  status: PropTypes.object,
```
kpuputti: Using the `PropTypes.object` as a type check isn't too valuable IMO. I'd go as precise as possible, preferably with the `isRequired` validation.

Examples:

```javascript
propTypes = {
  // data1: {"key1": true, "key2": false}
  data1: PropTypes.objectOf(PropTypes.bool).isRequired,
  // data2: [{"username": "mikko", "isAdmin": false}, {"username": "maija", idAdmin: true}]
  data2: PropTypes.arrayOf(PropTypes.shape({
    username: PropTypes.string.isRequired,
    isAdmin: PropTypes.bool.isRequired
  })).isRequired
}
```

As you can see, while quite verbose, you can do very detailed type checking with the PropTypes. I think we would benefit from it in the long run.

client/app/components/OnboardingTopBar/OnboardingTopBar.scss
```diff
@@ -0,0 +1,84 @@
+.topbarContainer {
+  width: 100%;
+  max-width: 120em;
+  color: white;
+  padding-top: 1.2em;
+  padding-bottom: 1.2em;
+  background-image: -moz-linear-gradient( 180deg, rgb(89,179,162) 0%, rgb(89,165,179) 100%);
+  background-image: -webkit-linear-gradient( 180deg, rgb(89,179,162) 0%, rgb(89,165,179) 100%);
+  background-image: -ms-linear-gradient( 180deg, rgb(89,179,162) 0%, rgb(89,165,179) 100%);
```
kpuputti: Let's not forget to include autoprefixer with postcss and remove these vendor-prefixed values.

client/app/components/OnboardingTopBar/OnboardingTopBar.scss
```diff
@@ -0,0 +1,84 @@
...
+  padding: 10px 15px;
+  border: 1px solid rgba(255,255,255,0.5);
+  border-radius: 5px;
+  text-decoration: none;
+  color: white;
+  &:hover {
+    color: white;
+    border: 1px solid rgba(255,255,255,1);
+  }
+}
```
kpuputti: Missing newline.

client/app/components/OnboardingTopBar/OnboardingTopBar.scss
```diff
@@ -0,0 +1,84 @@
...
+.progressBar {
+  background-color: rgb(184, 234, 225);
+  box-shadow: 0px 5px 8px 0px rgba(0, 0, 0, 0.06);
+  height: 100%;
+  border-radius: 0.5em;
+  transition: width 2s;
+}
+
+.progressBarBackground {
+  @media (max-width: 600px) {
```
kpuputti: Quite a few inline values that could be moved into vars?

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import scss from './styles.scss';
```
kpuputti: Rename scss to css to be consistent (we might not even use scss in the future).

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+      <a href={ props.pageData.link }
+        className={ scss.nextButton } >
+        { props.t('add_your_own') }
+      </a>
+    </div>
+  );
+};
+
+export default GuideCoverPhotoPage;
```
kpuputti: Missing newline. Also in several files below.

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
```diff
@@ -0,0 +1,48 @@
...
+const GuideCoverPhotoPage = (props) => {
+
+  const handleClick = function(e, path) {
+    e.preventDefault();
+    props.changePage(path);
+  }
+
+  return (
+    <div className="container">
+      <a className={ scss.backLink }
```
kpuputti: I think links are one thing that definitely should be their own component since they will have common functionality throughout the app, and will integrate to the global routing functionality, however that is implemented.

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
```diff
@@ -0,0 +1,48 @@
...
+        onClick={ (e) => handleClick(e, "") }
+        href={ props.initialPath } >
+        ‹ { props.t('back_to_todo') }
```
kpuputti: Passing in the translation helper function to child components ties them unnecessarily to the parent component. This seems like a sign of sharing functionality, but with means that is not the most intuitive one.

client/app/components/onboardingGuide/GuideCoverPhotoPage.jsx
```diff
@@ -0,0 +1,48 @@
...
+        { props.t('description') }
+      </p>
+
+      <div className={ scss.sloganImageContainer } >
+        <div className={ scss.sloganImage } ></div>
+      </div>
+
+      <div className={ scss.infoTextContainer } >
+        <div className={ scss.infoTextIcon } >
+          <i className="ss-info"></i>
```
kpuputti: This could be a separate `<Icon name="ss-info"/>` component.

rap1ds: You **can not** use "ss-" icon classes directly. They will not work in OS installation. You need to use the IconMap and figure out the proper class, which may be ss or font-awesome.

app/views/styleguide/pages/_header.erb
```diff
@@ -1,4 +1,7 @@
 <h2>ToC</h2>
 <ul>
+  <li>
+    <%= link_to "Onboarding Guide Page", styleguide_getting_started_guide_path %>
+  </li>
 </ul>
 <br/>
```
Gnito: As a suggestion: you could add OnboardingTopBar component to its own style page too

app/views/styleguide/pages/index.html.erb
```diff
@@ -4,3 +4,14 @@
...
+  translations: {"progress_label": "Marketplace progress","next_step": "Next"},
+  status: {
+    "community_id": 501,
+    "slogan_and_description": true,
+    "cover_photo": true,
+    "filter": true,
+    "paypal": false,
+    "listing": true,
+    "invitation": true}
+  }, prerender: false) %>
```
Gnito: I added props to the styleguide's pages_controller (as we do normally on Rails side)

client/app/components/OnboardingTopBar/OnboardingTopBar.jsx
Gnito: @kpuputti @mporkola 
I have that curried version on my branch - and it might make sense to move it to some utils-file. However, HoC might make more sense, but since these translations still need more thinking I'd suggest that we don't spend time to building it yet. 
(Thinking part: how to add links and other interpolation stuff)

client/app/components/OnboardingTopBar/OnboardingTopBar.jsx
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  progress() {
+    const steps = Object.keys(this.props.status).filter((entry) =>
+      ['slogan_and_description',
+       'cover_photo',
+       'filter',
+       'paypal',
+       'listing',
+       'invitation'].includes(entry)).map((key) => [key, this.props.status[key]]);
```
Gnito: Could this be spread so that next processing steps are on their own row:
```
Object.keys(this.props.status)
  .filter((entry) =>
    [ 'slogan_and_description',
      'cover_photo',
      'filter',
      'paypal',
      'listing',
      'invitation'
    ].includes(entry))
  .map((key) => [key, this.props.status[key]]);
```

client/app/components/OnboardingTopBar/OnboardingTopBar.scss
```diff
@@ -0,0 +1,84 @@
...
+
+.nextLabel {
+  display: inline-block;
+  margin-right: 0.7em;
+}
+
+.nextButton {
+  display: inline-block;
+  font-weight: 600;
+  padding: 10px 15px;
```
Gnito: Maybe change these to ems?

app/views/layouts/application.haml
mporkola: The immediate reason is that the other places are done by Vesa, and this by me :) A better reason is that we haven't agreed on the convention. I think raising is good, I'll add it here

client/app/components/OnboardingTopBar/OnboardingTopBar.jsx
mporkola: I agree, we should do like this. I'll fix

app/views/layouts/application.haml
ovan: This can be set globally in react on rails initializer. Best set desired default behaviour there.

client/app/components/OnboardingTopBar/OnboardingTopBar.scss
mporkola: My line of thinking was to keep this as plain CSS as possible, so we can easily move to whichever postcss setup we decide. But keeping things in vars probably does make it easier, instead of more difficult.

app/controllers/styleguide/pages_controller.rb
```diff
@@ -70,5 +70,21 @@ def data
...
+        next_step: "Next",
+        slogan_and_description: "Add Slogan / Description",
+        cover_photo: "Upload cover photo",
+        filter: "Add Fields / Filters",
+        paypal: "Accept payments",
+        listing: "Add a listing",
+        invitation: "Invite users"
+      },
+      progress: 83,
+      nextStep: 'paypal',
```
Gnito: I'd change these to camel case

app/controllers/application_controller.rb
```diff
@@ -605,6 +605,19 @@ def setup_logger!(metadata)
     logger.add_metadata(metadata)
   end
 
+  def onboarding_topbar_props
+    community_id = @current_community.id
+    onboarding_status = Admin::OnboardingWizard.new(community_id).setup_status
```
Gnito: You have onboarding_status in ApplicationController

app/views/layouts/application.haml
```diff
@@ -6,6 +6,11 @@
     .noscript-padding
       -# Noscript content will be positioned here
 
+  - if @current_user && @current_community && @current_user.has_admin_rights_in?(@current_community)
+    - with_feature(:onboarding_redesign_v1) do
+      - props = onboarding_topbar_props
```
Gnito: You could use just props: onboarding_topbar_props

client/app/components/OnboardingTopBar/OnboardingTopBar.scss
```diff
@@ -0,0 +1,89 @@
...
+  padding-bottom: 1.2em;
+  // todo: remove when we get postcss running
+  background-image: -moz-linear-gradient( 180deg, $gradientStart 0%, $gradientEnd 100%);
+  background-image: -webkit-linear-gradient( 180deg, $gradientStart 0%, $gradientEnd 100%);
+  background-image: -ms-linear-gradient( 180deg, $gradientStart 0%, $gradientEnd 100%);
+  box-shadow: $topbarShadow;
+}
+
+.topbar {
+  display: flex;
```
Gnito: This probably don't work without prefixes on IE, old iOS etc.

client/app/components/OnboardingTopBar/OnboardingTopBar.scss
```diff
@@ -0,0 +1,89 @@
...
+  display: inline-block;
+  font-weight: 600;
+  font-family: sans-serif;
+  font-size: 16px;
+  flex-shrink: 0;
+  margin-right: 0.5em;
+}
+
+.progressLabelPercentage {
+  font-size: 21px;
```
Gnito: I added lots of font-sizes and other style vars to app-variables.scss

client/app/utils/i18n.js
```diff
@@ -0,0 +1,7 @@
+function translate(translations) {
```
Gnito: I'd change the name of this file to translationUtils.js or i18nUtils.js. Without path, I'd assume this would contain something completely different (like something from http://i18njs.com)

config/initializers/react_on_rails.rb
```diff
@@ -36,7 +36,7 @@
 
   config.logging_on_server = true # Default is true. Logs server rendering messags to Rails.logger.info
 
-  config.raise_on_prerender_error = false # change to true to raise exception on server if the JS code throws
+  config.raise_on_prerender_error = true # change to true to raise exception on server if the JS code throws
```
Gnito: Do we want to make this global decision?
While developing I have used:
```
= react_component("OnboardingGuideApp", props: props, prerender: true, raise_on_prerender_error: true)
```
But I'm not sure I want to keep it (or at least switch it always on)

mporkola:
All review comments should now be fixed, except for making SASS use more variables and moving them to one place

client/webpack.client.rails.build.config.js
```diff
@@ -18,14 +18,16 @@ config.module.loaders.push(
...
     ),
   },
   {
     test: /\.scss$/,
     loader: ExtractTextPlugin.extract(
       'style',
       'css?minimize&modules&importLoaders=3&localIdentName=[name]__[local]__[hash:base64:5]' +
+      '!postcss' +
       '!sass' +
       '!sass-resources'
```
kpuputti: Do we need a separate sass-loader now that postcss is in use?

client/app/components/OnboardingTopBar/OnboardingTopBar.scss
kpuputti: Autoprefixer could handle this, if the file is loaded with postcss?

client/app/components/OnboardingTopBar/OnboardingTopBar.jsx
```diff
@@ -90,9 +51,10 @@ class OnboardingTopBar extends React.Component {
...
-  translations: PropTypes.object,
-  status: PropTypes.object,
-  guide_root: PropTypes.string,
+  translations: PropTypes.objectOf(PropTypes.string).isRequired,
+  guideRoot: PropTypes.string.isRequired,
+  progress: PropTypes.number.isRequired,
+  nextStep: PropTypes.string.isRequired,
 };
 
 export default OnboardingTopBar;
```
kpuputti: This file is super good after these changes 👍

client/app/utils/i18n.js
kpuputti: IMO being inside the `utils` directory is clear enough. Any single word could be a popular JavaScript project, and those would be located in some other  directory anyways.

However, I would have a convention to import any util file such as `utils/foo.js` as `fooUtils` if the whole module is imported to a "namespace" object like `import * as fooUtils from 'util/foo'`.

app/view_utils/onboarding_view_utils.rb
```diff
@@ -69,7 +69,9 @@ def popup_locals(show_popup, guide_base_path, setup_status)
   end
 
   def progress(setup_status)
-    100 * (1 - (incomplete_steps(setup_status).count / STEPS.count.to_f))
+    total_steps = STEPS.count + 1
```
kpuputti: Could this hidden step be documented here for clarity?

client/webpack.client.rails.build.config.js
kpuputti: Also, I think the `webpack.server.rails.build.config.js` file also needs the postcss loader?

PR #1923, opened 2016-04-18
Ensure membership on account create through FB

sktoiva:


rap1ds:
LGTM

app/controllers/sessions_controller.rb
```diff
@@ -101,30 +102,42 @@ def facebook
     data = request.env["omniauth.auth"].extra.raw_info
     I18n.locale = URLUtils.extract_locale_from_url(request.env['omniauth.origin']) if request.env['omniauth.origin']
 
-    person =
-      Person.find_by_facebook_id_and_community(data.id, @current_community.id) ||
-      Person.find_by_email_address_and_community_id(data.email, @current_community.id)
+    persons = Person
+              .joins(:emails, :community_memberships)
+              .where(["people.facebook_id = ? OR emails.address = ?", data.id, data.email]).uniq
```
rap1ds: `joins` will join the table only for the SQL query and doesn't include any data to the result. Later, when we loop over `persons` and ask for `community_memberships` or `emails`, they will make SQL queries.

This will show the issue:

```ruby
Person.joins(:emails, :community_memberships).where("emails.address IS NOT NULL").map { |p| [p.emails, p.community_memberships] }
```

...and this will fix the issue:

```ruby
Person.includes(:emails, :community_memberships).references(:emails).where("emails.address IS NOT NULL").map { |p| [p.emails, p.community_memberships] }
```

app/controllers/sessions_controller.rb
```diff
@@ -101,30 +102,42 @@ def facebook
...
-      sign_in_and_redirect @person, :event => :authentication
+      sign_in_and_redirect person, :event => :authentication
     else
-
-      if data.email.blank?
-        flash[:error] = t("layouts.notifications.could_not_get_email_from_facebook")
-        redirect_to sign_up_path and return
+      # Remove also the surrounding if check when removing the feature check
+      # Afterwards, only the logic for creating user is needed
+      # rubocop:disable Style/IfInsideElse
```
rap1ds: Do we need to re-enable this? Or what's the scope of this disable rule?

PR #1924, opened 2016-04-18
Use marketplace ID as analytics identifier instead of admin email

mporkola:


ovan:
Check the rubocop failure before merging.

PR #1931, opened 2016-04-20
Custom fields act as search filters by default

ovan:
The default used to be false but this PR changes the default option to be true.

### DIFF
--- a/app/views/admin/custom_fields/form/_search_filter.haml
+++ b/app/views/admin/custom_fields/form/_search_filter.haml
@@ -4,13 +4,13 @@
       - custom_field.with_type do |type|
         - case type
         - when :date_field
-          = form.check_box :search_filter, disabled: true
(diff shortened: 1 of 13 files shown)


app/views/admin/custom_fields/form/_search_filter.haml
```diff
@@ -4,13 +4,13 @@
       - custom_field.with_type do |type|
         - case type
         - when :date_field
-          = form.check_box :search_filter, disabled: true
+          = form.check_box :search_filter, checked: false, disabled: true
```
rap1ds: Why is this needed?

ovan: It just looks more logical that when text or date fields cannot be made a filter and the box is shown as disable the checkbox is not checked.

PR #1938, opened 2016-04-21
Add POST resource for prospect emails

lyyder:
Adds a resource where prospect emails can be posted.

The old `int_api/check_email_availability` resource can be removed once the marketing site uses the `int_api/prospect_email` resource.


rap1ds:
Code looks good! Remember to test

PR #1942, opened 2016-04-22
Onboarding popup

ovan:


app/views/layouts/_lightbox.haml
```diff
@@ -1,6 +1,6 @@
 .lightbox{:id => id}
   .lightbox-content
     %a{:href => "#", :class => "close lightbox-x", :id => "close_x"}
-      .ss-delete.icon-fix
+      .ss-delete
```
mporkola: `ss-` icon classes shouldn't be used directly, because we use a different icon library in OS.
`= icon_tag('delete')` would be close to what we want instead.

PR #1943, opened 2016-04-22
Add proper indices

rap1ds:
- [X] `people.username` NOT NULL
- [X] `emails.address` NOT NULL
- [X] Unique `people.facebook_id`, `people.community_id` pair
- [X] Unique `emails.address`, `emails.community_id` pair
- [X] Unique `people.username`, `people.community_id` pair

In the next PR (needs still some work):

- [x] Set `people.community_id` NOT NULL
- [x] Set `emails.community_id` NOT NULL

### DIFF
--- a/db/migrate/20160422094212_add_index_unique_community_id_and_facebook_id.rb
+++ b/db/migrate/20160422094212_add_index_unique_community_id_and_facebook_id.rb
@@ -1,0 +1,8 @@
+class AddIndexUniqueCommunityIdAndFacebookId < ActiveRecord::Migration
+  def change
+    # Remove the old facebook_id from index because we don't
+    # do searches with facebook id only
+    remove_index :people, column: :facebook_id
+    add_index :people, [:facebook_id, :community_id], unique: true
+  end
+end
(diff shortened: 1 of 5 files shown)


db/migrate/20160422094212_add_index_unique_community_id_and_facebook_id.rb
```diff
@@ -0,0 +1,8 @@
+class AddIndexUniqueCommunityIdAndFacebookId < ActiveRecord::Migration
+  def change
+    # Remove the old facebook_id from index because we don't
+    # do searches with facebook id only
```
sktoiva: Yes we do, in sessions_controller:

```ruby
  persons = Person
              .includes(:emails, :community_memberships)
              .references(:emails)
              .where(["people.facebook_id = ? OR emails.address = ?", data.id, data.email]).uniq
```

db/migrate/20160422094212_add_index_unique_community_id_and_facebook_id.rb
```diff
@@ -0,0 +1,8 @@
...
+    remove_index :people, column: :facebook_id
```
sktoiva: See previous, this should be present.

db/migrate/20160422094212_add_index_unique_community_id_and_facebook_id.rb
rap1ds: Good point. We should now change the `sessions_controller` query to not use `community_memberships` but instead use the `community_id`. But, it's not changed yet so I'll keep the index there for a while.

PR #1946, opened 2016-04-22
Non null community id

rap1ds:
- [X] Set `-1` value to people who have `community_id` NULL
- [X] Copy that to emails
- [X] Set `people.community_id` NOT NULL
- [X] Set `emails.community_id` NOT NULL

### DIFF
--- a/db/migrate/20160422114747_copy_community_id_from_people_to_emails.rb
+++ b/db/migrate/20160422114747_copy_community_id_from_people_to_emails.rb
@@ -1,0 +1,23 @@
+class CopyCommunityIdFromPeopleToEmails < ActiveRecord::Migration
+  def up
+    exec_update(
+      ["UPDATE emails e",
+       "LEFT JOIN people p ON p.id = e.person_id",
+       "SET e.community_id = -1",
+       "WHERE p.community_id = -1"].join(" "),
+      "Set community_id -1",
+      []
+    )
+  end
+
+  def down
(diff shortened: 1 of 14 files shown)


db/migrate/20160422114747_copy_community_id_from_people_to_emails.rb
```diff
@@ -0,0 +1,23 @@
...
+  def up
+    exec_update(
+      ["UPDATE emails e",
+       "LEFT JOIN people p ON p.id = e.person_id",
+       "SET e.community_id = -1",
+       "WHERE p.community_id = -1"].join(" "),
+      "Set community_id -1",
+      []
+    )
+  end
```
lyyder: Why not just set the `NULL` `community_id`s in emails table to -1. Could copying from `people` table possibly leave some emails having `community_id` as `NULL`?

rap1ds: I was thinking that this is also a good place to clean up the `emails` table. If there are any `emails` to do not match to any `person`, those emails can be removed. So after this migration is run, we can have a look at the DB if there are any orphan emails (`community_id` still `NULL`) and remove them.

lyyder: Sounds good!

rap1ds:
Tests are broken, still work to do

PR #1952, opened 2016-04-25
Check super admin status on email availability query

sktoiva:


app/models/email.rb
```diff
@@ -50,7 +50,7 @@ def confirm!
   def self.email_available?(email, community_id)
    !Email
      .joins(person: :community_memberships)
-     .where("address = ? AND community_memberships.community_id = ?", email, community_id)
+     .where("address = :email AND (is_admin = '1' OR community_memberships.community_id = :cid)", email: email, cid: community_id)
```
rap1ds: The `joins` will create an `INNER JOIN`. Is that a problem?

rap1ds:
Code looks good, remember to test!

PR #1953, opened 2016-04-25
Include rows when membership is missing

sktoiva:
If superadmin is not part of any community, we still want to check the
username availability.


rap1ds:
LGTM, remember to test

PR #1954, opened 2016-04-25
Enable password recovery for superadmins

sktoiva:


rap1ds:
Code looks good to me, remember to test!

PR #1958, opened 2016-04-25
Community specific blacklist check for usernames

lyyder:


rap1ds:
The code looks good to me. I didn't test this so remember to test this.

rap1ds:
Now that the creation of `Person` and `CommunityMembership` are in the same method, you could also wrap them in transaction, so that is e.g. CommunityMembership creation fails, the code will not create an orphan `Person`, but rollback instead.

PR #1959, opened 2016-04-25
Unique community_memberships.person_id

rap1ds:
Person and CommunityMemberhips have 1-to-1 relationship. Thus, the `person_id`in `community_memberships` should be unique globally, not only per community.

- [X] Add unique index for `community_memberships.person_id`
- [X] Fix tests where user belongs to multiple communities
- [X] Remove step "user is a member of community". Instead, create the membership during the person creation.

### DIFF
--- a/app/models/community_membership.rb
+++ b/app/models/community_membership.rb
@@ -3,7 +3,7 @@
 # Table name: community_memberships
 #
 #  id                  :integer          not null, primary key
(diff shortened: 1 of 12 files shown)


rap1ds:
Tests are failing. I need to fix those before review.

mporkola:
ok!

PR #1960, opened 2016-04-25
Cleanup CommunityMembershipsController

rap1ds:
This PR heavily refactors the `CommunityMemberships` controller.

Previously, this controller had two methods: `new` and `create`. There were two paths when user was redirected to these methods: 1) When user **joins** another community with existing account and 2) when user creates a new account with **Facebook**. Nowadays we don't have joining anymore, so the path 2 is the only option to get to this controller.

The controller is still quite complex and has many responsibilities:

- Check that the user accepted terms
- Check that the invitation code is valid if community requires invitation
- Check that the email address is valid if the community requires specific email addresses

What has been done in this PR:

- Rename methods `new -> pending_consent` and `create -> consent_given`
- Remove all the "joining" logic
- Pass the data that is needed to render the form in `locals` instead of instance variables
- Get rid of extensive `if-else` branching, and use `Result` object to make a easy to follow logic chain.
- Isolate side-effects (DB writes) to one place only, and wrap them in transaction
- Remove resourceful routes and use instead non-resourceful routes


app/controllers/community_memberships_controller.rb
```diff
@@ -9,102 +9,185 @@ class CommunityMembershipsController < ApplicationController
...
+      Result::Error.new("Updating membership failed", reason: :update_failed, errors: membership.errors.full_messages)
+    end
+  end
 
-      Delayed::Job.enqueue(CommunityJoinedJob.new(@current_user.id, @current_community.id))
-      Delayed::Job.enqueue(SendWelcomeEmail.new(@current_user.id, @current_community.id), priority: 5)
-      flash[:notice] = t("layouts.notifications.you_are_now_member")
+  def ensure_pending_consent
+    if membership.nil?
+      report_missing_membership(@current_user, @current_community) if membership.nil?
```
lyyder: Maybe the second `if membership.nil?` could be dropped?

app/controllers/community_memberships_controller.rb
```diff
@@ -9,102 +9,210 @@ class CommunityMembershipsController < ApplicationController
...
+
+      Result::Success.new(membership)
+    rescue
+      Result::Error.new("Updating membership failed", reason: :update_failed, errors: membership.errors.full_messages)
+    end
+  end
+
+  def ensure_pending_consent
+    if membership.nil?
+      report_missing_membership(@current_user, @current_community) if membership.nil?
```
lyyder: Maybe the second `if membership.nil?` could be dropped?

config/routes.rb
```diff
@@ -227,11 +236,8 @@
         post :join
       end
     end
-    resources :community_memberships, :as => :tribe_memberships do
-      collection do
-        get :access_denied
-      end
-    end
+
+    # TODO Change to not restful resource
```
lyyder: Does this apply to listings resource or could the comment be removed?

lyyder:
LGTM. Readability of the `CommunityMembershipsController` has improved a lot!

Maybe still a second look from someone else could come in handy in a PR extensive as this.

config/routes.rb
rap1ds: Damn, forgot to remove the comment. Good catch

app/controllers/community_memberships_controller.rb
```diff
@@ -9,102 +9,210 @@ class CommunityMembershipsController < ApplicationController
...
-        # abort user creation if invitation is not usable.
-        # (This actually should not happen since the code is checked with javascript)
-        # This could happen if invitation code is coming from hidden field and is wrong/used for some reason
-        session[:invitation_code] = nil # reset code from session if there was issues so that's not used again
-        ApplicationHelper.send_error_notification("Invitation code check did not prevent submiting form, but was detected in the CommunityMembershipsController", "Invitation code error")
-
-        # TODO: if this ever happens, should change the message to something else than "unknown error"
-        flash[:error] = t("layouts.notifications.unknown_error")
-        render :action => :new and return
+  def consent_given
```
sktoiva: Should this be `give_consent`? All the other actions that include `create` or `update` are in imperative.

app/controllers/community_memberships_controller.rb
```diff
@@ -9,102 +9,210 @@ class CommunityMembershipsController < ApplicationController
...
 
-  def access_denied
+  def report_missing_membership(user, community)
+    ArgumentError.new("User doesn't have membership. Don't know how to continue. person_id: #{user.id}, community_id: #{community.id}")
+  end
+
+  def membership
+    # TODO This should be changed to @current_user.community_membership when the relation is
+    # changed to has_one instead of has_many
+    @membership ||= @current_user.community_memberships.find_by(community_id: @current_community.id)
```
sktoiva: This is already in place in master. `community_membership` can be called directly and the `TODO` can be removed.

sktoiva:
Huge improvement to readability! Couple of comments.

PR #1961, opened 2016-04-25
Correctly reset legacy password and salt

sktoiva:
If legacy password and salt is present, password recovery works but logging in with the new password doesn't after that. 

This correctly resets the legacy password and salt in order for the new password to work.

Steps to reproduce:

1. Have a user with legacy_encrypted_password in database (not logged in)
2. Click forgot my password
3. Enter email address and change password according to instructions
4. Try to log in with new password

Other option:

1. Have a user with legacy_encrypted_password in database (logged in, with persistent session, don't use the log in functionality)
2. Go to profile
3. Change password
4. Log out
5. Try to log in with new password

`reset_password!` is not called and thus the old passwords are not reset properly.


rap1ds:
Was this issue in the previous release? If it was, then we could add a CHANGELOG entry.

PR #1962, opened 2016-04-26
Deprecate communities and community_memberships

sktoiva:
Deprecate has_many relationships, these will be transformed to has_one
or replaced with person.community_id.

This pull request also introduces a common method for deprecating methods inside the app. Calling a deprecated method will produce following log message:
`DEPRECATION WARNING: community_memberships is deprecated and will be removed in future. | Use community_membership instead. (called from <main> at (pry):11)`

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -112,6 +112,10 @@ class Person < ActiveRecord::Base
 
   has_and_belongs_to_many :followed_listings, :class_name => "Listing", :join_table => "listing_followers"
 
+  deprecate communities: "Use community instead.",
+            community_memberships: "Use community_membership instead.",
+            deprecator: MethodDeprecator.new
+
   def to_param
     username
   end
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -103,6 +103,8 @@ class Person < ActiveRecord::Base
   has_many :authored_comments, :class_name => "Comment", :foreign_key => "author_id", :dependent => :destroy
   has_many :community_memberships, :dependent => :destroy
   has_many :communities, -> { where("community_memberships.status = 'accepted'") }, :through => :community_memberships
+  has_one  :community_membership, :dependent => :destroy
+  has_one  :accepted_community, -> { where("community_memberships.status= 'accepted'") }, through: :community_membership, source: :community
   has_many :invitations, :foreign_key => "inviter_id", :dependent => :destroy
   has_many :auth_tokens, :dependent => :destroy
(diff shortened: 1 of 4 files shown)


rap1ds:
The deprecator looks great!

Some kind of documentation about the usage would be great. Maybe just add some comment how to use it? At least one question that came to my mind is that where can I use this `deprecate` method? Is it only for models, only for Rails classes or for all classes? Can I use this in modules?

I'm also thinking about the naming... I guess this is meant for deprecating methods only, so should it be `SharetribeMethodDeprecator`? Also, I'm not sure about the `Sharetribe` prefix. Do we need it? I know that the logger is named `SharetribeLogger` but I'm not sure if I like that or not... I don't know answers to these, just something that came to my mind.

Last thing, the message says "Use community instead", however, this PR doesn't include the `has_one :community`. Would it make sense to add it here? Or optionally, only include the common method for deprecating methods in this PR and add the deprecation later.

app/models/person.rb
```diff
@@ -103,6 +103,8 @@ class Person < ActiveRecord::Base
   has_many :authored_comments, :class_name => "Comment", :foreign_key => "author_id", :dependent => :destroy
   has_many :community_memberships, :dependent => :destroy
   has_many :communities, -> { where("community_memberships.status = 'accepted'") }, :through => :community_memberships
+  has_one  :community_membership, :dependent => :destroy
+  has_one  :community, -> { where("community_memberships.status= 'accepted'") }, through: :community_membership
```
rap1ds: How do you feel about this? Should we follow the same path as the existing implementation (community through membership where status is accepted), or should we just simply say `has_one :community` and use the `person.community_id` as the foreign key? I think that the latter would be a bit simpler, but on the other hand that's then changing the logic (status doesn't need to be accepted). If we decide to change the logic, then that needs to be taken into account when changing the deprecated `communities` to `community`.

sktoiva: What if we go through a middle ground here, where wouldn't change the logic, but rename the scope `community` to `accepted_community`? That way we would only have to rename `communities` to `accepted_community` and remove the iteration of the collections. 

I think that this would better tell what's the purpose of the member.

sktoiva: Or actually, some other, better, name that would convey the meaning of being the one where user is a member with citizenship.

rap1ds: Sounds like a good plan. `accepted_community` describes the meaning well. It might be that we need to also add plain `community`, but we can add that later if we need to.

PR #1964, opened 2016-04-26
Facebook email unconfirmed

rap1ds:
`emails.community_id` and `emails.address` pair should be unique. The same email address can not be used twice in the same community (because address is an identifier we use for logging people in).

With Facebook login, it's possible to create duplicated emails.

Steps to reproduce:

1. Make sure there are no user in the DB that uses the `facebook_id` of your account.
1. Make sure there are no email address in the DB that uses the email address of your Facebook account.
1. Create a new user with the email account that you have in your Facebook account.
1. Don't confirm the email address
1. Log out
1. Create a new user with your Facebook account

Expected: Error message, that says that the email address is already been taken but it's not confirmed.

Actual: New account successfully created. There are 2 email rows with the same address in the DB.


sktoiva:
This is a bit different now from how we regularly handle email confirmations, but on the other hand there might be some security implications if we'd log the user in and show the regular pending email confirmation page. 

The code looks good otherwise!

PR #1966, opened 2016-04-27
Fix missing legacy password data

sktoiva:
When we duplicated users based on community membersihps, we left out legacy encrypted password from the duplicates by accident.

This adds the data for users that haven't yet changed their password.


rap1ds:
Code looks clean! If you want to still improve a bit, you can use `exec_update` instead of `execute`. It will print the number of rows affected.

PR #1976, opened 2016-04-28
Move confirmation_pending to CommunityMemberships controller

rap1ds:
Merge this only after #1960 

This PR moves `confirmation_pending` action from `SessionsController` to `CommunityMemberships` controller. `pending_email_confirmation` is one of the membership statuses, so it makes sense to have the action in `CommunityMemberships` controller. It has nothing to do with session creation, because session has been created already before user sees this page.

This PR also fixes one small but super annoying bug in the app:

Steps to reproduce:

1. Sign up
2. When you see the confirmation page, open another tab and go to your webmail
3. Click the link in the confirmation email
4. Go back to the tab where the confirmation page is open
5. Refresh

Expected result: You are redirected to front-page. You see a message which says that you are already a member.

Actual result: No redirect. The page looks a bit broken because the email is missing.


ovan:
Looks good.

sktoiva:
LGTM

PR #1994, opened 2016-04-29
Show onboarding for super admins

rap1ds:
- [X] Check if user is super admin
- [X] Change the logic a bit: Show if the admin status is accepted (previously the logic was: show if status is not "pending email confirmation")


mporkola:
[selfie-2](https://github.com/thieman/github-selfies/)

PR #1995, opened 2016-04-29
Update CHANGELOG and UPGRADE

rap1ds:


UPGRADE.md
```diff
@@ -18,22 +18,50 @@ RAILS_ENV=production rake sharetribe:generate_customization_stylesheets_immediat
...
 npm install
 ```
 
+### User account migrations
+
+You can ignore this section if you are running only one marketplace in your Sharetribe instance.
+
+This release removes the ability for one user to belong to multiple marketplaces. From now on one user belongs to one and only one marketplace.
+
+Because of that this release contains quite a few migrations which will duplicate existing user accounts if they belong to multiple communities. For example, if one user belongs to three communities, two new users will be created so that each user belongs to only one community.
```
bladealslayer: Should there be comma after `Because of that` and before the `if`?

UPGRADE.md
```diff
@@ -18,22 +18,50 @@ RAILS_ENV=production rake sharetribe:generate_customization_stylesheets_immediat
...
+After bundle install, you should also install `npm` packages:
 
-After bundle install, you should also install npm packages:
 ```bash
 npm install
 ```
 
+### User account migrations
+
+You can ignore this section if you are running only one marketplace in your Sharetribe instance.
```
lyyder: Does the current OS version support multiple marketplaces in one instance? And if it does, has it been documented? Should we leave this chapter out if having multiple marketplaces is not an _official_ feature?

lyyder:
Added one comment, otherwise looks good to me!

UPGRADE.md
rap1ds: Well, it kind of it. But true, it's not documented. I'd say it is an "official" feature because the hosted version does that. And if I remember right, there have been some people asking questions about running multiple marketplaces in one instance, so I guess there are some people (not many) who are running multiple marketplaces in one instance.

It would require some code changes to make it fully supported for OS version (the `IntApi::MarketplaceController` currently adds `preauthorize` payment process, which is the default for hosted version but not supported for OS version, because of PayPal).

So how should we proceed? Should we make the multiple marketplaces "official" by doing the code changes and documenting it (is that a priority right now?). Or should we leave this out even though there might be someone running multiple marketplaces? Or should we leave this here and not make the effort to make to multiple marketplaces official? Any ideas?

PR #1996, opened 2016-05-02
Migrate from old session cookies to new

### LINKED ISSUE #1192: Login fails on fresh sharetribe install
I just installed sharetribe on EC2, you can see it working [here](http://54.191.173.8:8080/en).

I can create users and they show on the database, but I can't log in with any of them.
When I log in, it seems that it logs the user successfully, and then the user is redirected to `app/controllers/community_memberships_controller.rb` where it checks again if the user is logged in, and there it is not logged in anymore.

I can reproduce it on a fresh development install of sharetribe as well.

Any idea why this could be happening? Thanks!

sktoiva:
Prevents cookies from leaking to subdomains, and dynamically renames old cookies with
new key so that sessions are not invalidated after the change.

Removes old CustomDomainCookie and relies on Rails defaults.

Testing steps:

  1. Checkout master
  2. Run rails server
  3. Log in to some marketplace
  4. Checokut this branch and restart server
  5. Session should stay on, and visiting other subdomains shouldn't log you out, and the sessions should be separate.

Fixes #1192


lib/rack_middleware/custom_domain_cookie.rb
```diff
@@ -12,16 +12,11 @@ def initialize(app, default_domain)
...
     return @app.call(env)
   end
-
-  def custom_domain?(host)
-    return false if host.nil?
-    host !~ /#{@default_domain.sub(/^\./, '')}/i
-  end
-end
\ No newline at end of file
+end
```
rap1ds: I think we can now completely remove this file, can we?

UPGRADE.md
```diff
@@ -34,6 +34,8 @@ After bundle install, you should also install npm packages:
 npm install
 ```
 
+This release changes how cookies are handled. Previously, all cookies were shared between different subdomains. Now these are separated by subdomain. If the default `session_key` has been changed in `config.defaults.yml` this requires actions. The `session_key` config variable should be renamed to `old_session_key`, and `session_key` should contain a differing key.
```
ovan: I read this twice but can't say fully get it. So if I have session_key set I must generate myself a new key and update the value of the session_key config variable? Why is this? What is the effect of renaming current session_key to old_session_key and why must that be done?

UPGRADE.md
```diff
@@ -34,6 +34,10 @@ After bundle install, you should also install npm packages:
 npm install
 ```
 
+This release changes how cookies are handled. Previously, all cookies were shared between different subdomains i.e. domain of cookies was sharetribe.com. Therefore, it was possible to log in to foo.sharetribe.com and log in automatically also to bar.sharetribe.com. Recently, the domain model was updated so that person can belong only to one marketplace. This caused an issue where logging in to foo.sharetribe.com and then visiting in bar.sharetribe.com logged user out of foo.sharetribe.com, because session was shared but the logged in user wasn't a member of bar.sharetribe.com.
+
+This release separates cookies by subdomain so that foo.sharetribe.com and bar.sharetribe.com have now separate session cookies. In order to migrate old sessions as smoothly as possible the default `session_key` has been changed in `config.defaults.yml` and a new key `old_session_key` is added. If the default `session_key` config variable has been changed, it should be renamed to `old_session_key` and `session_key` should contain a new, differing key.
```
ovan: This is now more clear. s/should/must ?

UPGRADE.md
```diff
@@ -34,6 +34,10 @@ After bundle install, you should also install npm packages:
 npm install
 ```
 
+This release changes how cookies are handled. Previously, all cookies were shared between different subdomains i.e. domain of cookies was sharetribe.com. Therefore, it was possible to log in to foo.sharetribe.com and log in automatically also to bar.sharetribe.com. Recently, the domain model was updated so that person can belong only to one marketplace. This caused an issue where logging in to foo.sharetribe.com and then visiting in bar.sharetribe.com logged user out of foo.sharetribe.com, because session was shared but the logged in user wasn't a member of bar.sharetribe.com.
```
ovan: Even though the description is very nice I think it's unnecessary in upgrade guide. The paragraph below is to the point and describes the actions that need to be taken. In addition, multi marketplace setup is not part of open source target so no need to go into detail to explain it. Sorry if I was unclear in my earlier comment.

lib/rack_middleware/custom_cookie_renamer.rb
```diff
@@ -0,0 +1,14 @@
+class CustomCookieRenamer
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    if env["HTTP_COOKIE"] && !(env["HTTP_COOKIE"] =~ /\b#{APP_CONFIG.session_key}/)
+      env["HTTP_COOKIE"].sub!(/\b#{APP_CONFIG.old_session_key}=/, "#{APP_CONFIG.session_key}=")
```
ovan: Should this check the case if old_session_key was being used? I guess the sub! does nothing if the cookie doesn't match old_session_key. Is that ok behavior?

rap1ds:
@sktoiva This will also solve the issue #1192, so you could add a comment about that to CHANGELOG.

config/config.defaults.yml
```diff
@@ -193,6 +193,7 @@ default: &default_settings
   # If you need for some reason change the session key used in cookies, you can do it here.
   # Normally no need to touch this.
   session_key: _sharetribe_session
```
rap1ds: I would add a deprecation comment here, saying that this variable is only needed to upgrade from 5.6.0 to 5.7.0 and will be removed soon.

PR #1997, opened 2016-05-02
Setup custom ESLint configuration

kpuputti:
Previously the linting setup was based on the boilerplate from react-on-rails, and it used the AirBnB setup as the base. However, the AirBnB setup is based on the default Espree parser in ESLint, and the config used a different babel-eslint parser and a bunch of plugins. It was also quite unclear what the actual result configuration of the project was.

This PR redoes the configuration from scratch, ensuring that only Babel is used for parsing (as it is used also for making the actual Webpack bundle) and each relevant rule is configured separately in this project.

The base for the configuration is the `eslint:recommended` ruleset from ESLint, as well as the `plugin:react/recommended` ruleset from eslint-plugin-react. This configuration is quite strict, but can be loosened if needed in the future. JSX configuration didn't see too much love since it is expected to be removed anyways.


client/.eslintignore
```diff
@@ -1 +0,0 @@
-node_modules
```
kpuputti: This is already ignored by default.

Gnito:
Btw. are there anyone else than me & olli who prefer comma first approach?
:)

Gnito:
Looks good to me

PR #1998, opened 2016-05-02
Redirect to new onboarding straight from email confirmation

mporkola:
Multiple redirects lose session data, which messes up analytics.

### DIFF
--- a/app/controllers/confirmations_controller.rb
+++ b/app/controllers/confirmations_controller.rb
@@ -61,9 +61,9 @@ class ConfirmationsController < Devise::ConfirmationsController
       if @current_user && @current_user.has_admin_rights?
         report_to_gtm({event: "admin_email_confirmed"})
         if feature_enabled?(:onboarding_redesign_v1)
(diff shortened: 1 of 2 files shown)


app/controllers/confirmations_controller.rb
```diff
@@ -60,7 +60,11 @@ def show
 
       if @current_user && @current_user.has_admin_rights?
         report_to_gtm({event: "admin_email_confirmed"})
-        redirect_to getting_started_admin_community_path(:id => @current_community.id) and return
+        if feature_enabled?(:onboarding_redesign_v1)
+          redirect_to getting_started_guide_admin_community_path(@current_community)
```
rap1ds: Missing `and return`

Gnito: Maybe we should use here (in both cases explicit param "id: @current_community.id"
```
redirect_to getting_started_guide_admin_community_path(id: @current_community.id) and return
```
use the same format also in Admin::CommunitiesController

PR #2002, opened 2016-05-02
Remove community from getting_started_guide path

Gnito:


config/routes.rb
```diff
@@ -172,6 +173,10 @@
           # See the above :admin_settings routes, outside of :communities resource
           get :settings,       to: redirect("/admin/settings")
           put :update_settings # PUT request, no redirect
+          get "getting_started_guide(/*all)" => redirect {|p, req|
+            sub_path = p[:all] == "" ? "" : "/#{p[:all]}"
+            "#{p[:locale]}/admin/getting_started_guide#{sub_path}"
+          }
```
Gnito: I'm not sure if this is worth the trouble since we just launched this feature. 
(Or if there's more elegant way to do this kind of redirection including sub paths.)

config/routes.rb
```diff
@@ -172,6 +173,10 @@
           # See the above :admin_settings routes, outside of :communities resource
           get :settings,       to: redirect("/admin/settings")
           put :update_settings # PUT request, no redirect
+          get "getting_started_guide(/*all)", to: redirect {|p, req|
+            sub_path = p[:all] == "" ? "" : "/#{p[:all]}"
+            "#{p[:locale]}/admin/getting_started_guide#{sub_path}"
+          }
```
Gnito: I'm not sure if this is worth the trouble since we just launched this feature. 
(Or if there's more elegant way to do this kind of redirection including sub paths.)

rap1ds: What about just listing all the different routes here? That would be simple, super easy to understand, implicit and less magical. Yeah, a bit more repetition, but I'm not sure if that's so bad thing here.

config/routes.rb
```diff
@@ -142,10 +142,11 @@
       get   "/settings" => "communities#settings",        as: :settings
       patch "/settings" => "communities#update_settings", as: :update_settings
 
+      get "getting_started_guide(/*all)" => "getting_started_guide#index", as: :getting_started_guide
```
rap1ds: Seems that the route name `:getting_started_guide` is now clashing with the one route in the styleguide. It's no not clear which route will `getting_started_guide_path` URL helper generate. Maybe the styleguide name could be renamed to something else, like `getting_started_guide_styleguide` etc.?

Gnito: They are in different namespace (:admin vs. :styleguide) - isn't that enough?
It would actually be more convenient that they have the same name.

rap1ds: To my understanding that is not enough.

If you give a name to a path, e.g. `as: :getting_started_guide`, Rails will generate helpers such as `getting_started_guide_path` and `getting_started_guide_url`. Calling for example the path helper `getting_started_guide_path` will return a path (i.e. /admin/getting_started_guide) to the named route. Now the problem is, if there are 2 routes with the same name, which path (/admin/getting_started_guide vs /style_guide/getting_started_guide) the helper will return?

Gnito: roger

config/routes.rb
Gnito: Repeating is possible alternative, but do we need this at all? 
This thing has been live for less than 1 week and we could probably survive without this. (or just redirection to guide root ```get "getting_started_guide(/*all)", to: redirect("/admin/getting_started_guide")```

config/routes.rb
Gnito: if I run "rake routes | grep getting_started_guide" I get
```
styleguide_getting_started_guide GET      /styleguide/getting_started_guide(/*all)(.:format)                                                                           admin_getting_started_guide GET (/:locale)/admin/getting_started_guide(/*all)(.:format)
````

However, if I rename those "as: :styleguide_getting_started_guide" and "as: :admin_getting_started_guide", I get:
```
styleguide_styleguide_getting_started_guide GET      /styleguide/getting_started_guide(/*all)(.:format)
admin_admin_getting_started_guide GET      (/:locale)/admin/getting_started_guide(/*all)(.:format)  
```

I'd like to avoid that double admin_admin_route_name thing

rap1ds: Ahaa ok! Sorry about my latest comment, apparently the namespace is enough as you said.

PR #2008, opened 2016-05-03
Disabling payments for an order type counts as configuring payments

mporkola:
Our onboarding process has a step "configure payments", but not every marketplace needs payments.

 - [x] other things
 - [x] popup

### DIFF
--- a/app/controllers/admin/listing_shapes_controller.rb
+++ b/app/controllers/admin/listing_shapes_controller.rb
@@ -12,6 +12,11 @@ class Admin::ListingShapesController < ApplicationController
     category_count = @current_community.categories.count
     template_label_key_list = ListingShapeTemplates.new(process_summary).label_key_list
 
+    onboarding_popup_locals = OnboardingViewUtils.popup_locals(
+      flash[:show_onboarding_popup],
+      getting_started_guide_admin_community_path(@current_community),
+      Admin::OnboardingWizard.new(@current_community.id).setup_status)
+
     render("index",
            locals: {
              selected_left_navi_link: LISTING_SHAPES_NAVI_LINK,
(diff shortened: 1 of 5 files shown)


app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -78,6 +78,18 @@ def update
 
     if update_result.success
       flash[:notice] = t("admin.listing_shapes.edit.update_success", shape: pick_translation(shape[:name]))
+
+      # Onboarding wizard step recording
+      state_changed = Admin::OnboardingWizard.new(@current_community.id)
+        .update_from_event(:listing_shape_updated, update_result)
```
ovan: We only fire this event from the success branch so we might as well pass directly `update_result.data` in to event handler so it wouldn't have to deal with the Result-object wrapping.

app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -78,6 +78,18 @@ def update
...
+      if state_changed
+        report_to_gtm([{event: "km_record", km_event: "Onboarding payments setup"},
+                       {event: "km_record", km_event: "Onboarding payment disabled"}])
+
+        with_feature(:onboarding_redesign_v1) do
+          flash[:show_onboarding_popup] = true
```
ovan: To make the popup show this needs the redirect target action to check for this parameter and put the popup locals in place (using the view util). Plus then the view code needs to include the popup partial. The view code can be simply copy&pasted from some other view that shows the popup.

Gnito:
@mporkola I created guide changes on top of this one
https://github.com/sharetribe/sharetribe/pull/2010

PR #2012, opened 2016-05-04
Unidirectional header bar dataflow

rap1ds:
The purpose of this PR is to make the dataflow in the header bar UI component fully (with some exceptions) unidirectional, which means that all the data needed to render the view are passed to the `global_header` partial in `locals`.

- [X] User data
- [X] Marketplace data
- [X] Paths
- [x] Icons
- [ ] Translations


app/controllers/application_controller.rb
```diff
@@ -644,6 +644,51 @@ def onboarding_topbar_props
...
+        [
+          "dropdown",
+          "mail",
+          "user",
+          "list",
+          "settings",
+          "logout",
+          "rows"
+        ])
+    }.merge(user)
```
ovan: Minor style thing but I'd find it more readable style if both user and the other has get local names and are merged together as a separate step at the end. Alternatively define them as one hash that is directly returned if wanting to avoid coming up with good names.

app/controllers/application_controller.rb
```diff
@@ -644,6 +644,51 @@ def onboarding_topbar_props
...
+  end
+
+  helper_method :header_props
+
+  # TODO Can we remove this?
+  def get_full_locale_name(locale)
+    Maybe(Sharetribe::AVAILABLE_LOCALES.find { |l| l[:ident] == locale.to_s })[:name].or_else(locale).to_s
+  end
```
ovan: What's required to answer the question in TODO comment? Would be cleaner to remove this at the same time (or replace the comment if it can't be removed)

app/helpers/application_helper.rb
```diff
@@ -20,6 +19,19 @@ def self.icon_specified?(icon_name)
...
+
+  def icon_class_tag(icon_class, additional_classes = [])
+    classes_string = [icon_class].concat(additional_classes).join(" ")
+    "<i class=\"#{classes_string}\"></i>".html_safe
+  end
+
+  def icon_map_tag(icon_map, icon_name, additional_classes = [])
+    icon_class_tag(icon_map[icon_name], additional_classes)
+  end
+
```
ovan: Can all the icon stuff be pulled to a separate module that is mixed in (included) to application helper?

ovan:
What do you think, does the icon stuff fall into same category as translations and routes so that it would be good that views could generate icon names without needing them to be passed in as properties? Isn't that just a static map with values for proprietary and open source icon class names? If so, then one implementation would be to have a json file that defines the mapping and can be both read in by server code and packages into JS build.

PR #2013, opened 2016-05-04
Link to onboarding guide from onboarding top bar

mporkola:


client/app/components/OnboardingTopBar/OnboardingTopBar.scss
```diff
@@ -29,6 +29,9 @@ $gradientEnd: rgb(89,165,179);
   font-size: 16px;
   flex-shrink: 0;
   margin-right: 0.5em;
+  a {
```
Gnito: I don't like this. Could you just use class? 
My idea is that we have one globals.css that is the only place where we touch tags, and other files touch only classes.

PR #2022, opened 2016-05-06
Onboarding final final fixes

Gnito:
Juho wanted to change his earlier copys + found couple of bugs

### DIFF
--- a/app/controllers/admin/getting_started_guide_controller.rb
+++ b/app/controllers/admin/getting_started_guide_controller.rb
@@ -19,6 +19,9 @@ class Admin::GettingStartedGuideController < ApplicationController
     has_sub_path = (path_parts.count == 2 && path_parts[1] != "/")
     sub_path = has_sub_path ? path_parts[1] : ""
 
+    listing_shape = ListingShape.where(community_id: @current_community.id).first
+    alternative_cta = edit_admin_listing_shape_path(listing_shape[:name])
+
     onboarding_status = Admin::OnboardingWizard.new(@current_community.id).setup_status
     links = {
       slogan_and_description: {
@@ -39,7 +42,7 @@ class Admin::GettingStartedGuideController < ApplicationController
       paypal: {
         sub_path: 'paypal',
         cta: admin_paypal_preferences_path,
-        alternative_cta: admin_listing_shapes_path,
+        alternative_cta: alternative_cta,
         info_image: view_context.image_path('onboarding/step5_screenshot_paypal@2x.png')
       },
       listing: {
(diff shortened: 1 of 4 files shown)


app/controllers/admin/getting_started_guide_controller.rb
```diff
@@ -19,6 +19,9 @@ def data
     has_sub_path = (path_parts.count == 2 && path_parts[1] != "/")
     sub_path = has_sub_path ? path_parts[1] : ""
 
+    listing_shape = ListingShape.where(community_id: @current_community.id).first
```
kpuputti: The other place that used the same expression `ListingShape.where(...).first` wrapped it within a Maybe, should this do that as well? Does this assume a shape is always found for a community? I don't know if anything else is possible, that's why I'm asking.

app/controllers/admin/getting_started_guide_controller.rb
```diff
@@ -19,6 +19,10 @@ def data
     has_sub_path = (path_parts.count == 2 && path_parts[1] != "/")
     sub_path = has_sub_path ? path_parts[1] : ""
 
+    alternative_cta = Maybe(ListingShape.where(community_id: @current_community.id).first)
```
rap1ds: `ListingShape` model shouldn't be referenced directly from a controller. You can use the `ListingService::API::Api.shapes.get(community_id: @current_community.id)` method.

app/controllers/styleguide/pages_controller.rb
```diff
@@ -29,6 +29,10 @@ def data
       invitation: true
     }
 
+    alternative_cta = Maybe(ListingShape.where(community_id: @current_community.id).first)
```
rap1ds: Same here.

app/controllers/admin/getting_started_guide_controller.rb
rap1ds: Btw. how do you know the `first` order type is the one that needs to be edited?

Gnito: Ah, true.
(btw. part: that's how Juho wanted this to work.)

config/locales/onboarding/en.yml
```diff
@@ -53,7 +53,7 @@ en:
           title_done: "All done!"
           description_p1: "To get your marketplace up and running, there are a few essential steps you need to take, as listed below."
           description_p2: "Once finished, your marketplace will be ready to receive its first visitors!"
-          congratulation_p1: "You did it! Your marketplace is now up and running. However, this is just the beginning. You've only seen a fraction of the customization options that Sharetribe offers. To read more about how to customize your marketplace further, check out <a href='http://support.sharetribe.com/knowledgebase' alt='knowledge base' target='_blank'>our knowledge base</a>."
+          congratulation_p1: "You did it! Your marketplace is now up and running. However, this is just the beginning. You've only seen a fraction of the customization options that Sharetribe offers. To read more about how to customize your marketplace further, check out <a href='http://support.sharetribe.com/knowledgebase/articles/892140-what-to-do-after-the-basic-setup-of-your-marketpla' alt='knowledge base' target='_blank'>our knowledge base</a>."
```
rap1ds: HTML in the translations :(

PR #2024, opened 2016-05-09
Add plan features for marketplace plans

sktoiva:
A first move towards getting rid of plan levels.


app/services/plan_service/store/plan.rb
```diff
@@ -2,20 +2,26 @@ module PlanService::Store::Plan
 
   class TrialModel < ActiveRecord::Base
     self.table_name = :marketplace_trials
+    serialize :features, Array
```
rap1ds: We may not need to store the `features` to the `marketplace_trials` table, because we know that these are trials and trials have certain set of features. To me this is a bit analogous to `plan_level`: We are not storing `plan_level` to `marketplace_trials` table, because we know that the value is always 0. Does that make sense?

sktoiva: Yes it does, and is fair point.

app/services/plan_service/data_types.rb
```diff
@@ -9,6 +9,8 @@ module PlanService::DataTypes
     [:id, :fixnum, :optional], # For OS, the plan is not actually in DB. Thus, optional.
     [:community_id, :fixnum, :mandatory],
     [:plan_level, :fixnum, :mandatory],
+    [:features, :array, :optional],
```
rap1ds: I'm thinking, is array the right data structure for features or should it be hash?

What if we have a feature which allows users to take backups from their marketplace in admin site. That would be a operation that takes a lot of time and thus we may want to limit the number of backups they can take per day. For starter plans, the value could be 1 backup / day and for pro plans it could be 6 backups / day. In that case, the array is not enough but hash would do the trick:

```
# trial
features: {
  whitelabel: false,
  backups_per_day: 0
}

# starter
features: {
  whitelabel: false,
  backups_per_day: 1
}

# pro
features: {
  whitelabel: true,
  backups_per_day: 6
}

Comment (sktoiva) [inline app/services/plan_service/data_types.rb]: Again a good point.

Comment (lyyder) [inline app/services/plan_service/data_types.rb]: This way maybe the member limit could also be included in the features.

File: app/services/plan_service/store/plan.rb (hunk 1 above)
Comment (lyyder) [inline app/services/plan_service/store/plan.rb]: @rap1ds one of the ideas behind replacing `plan_level` with `features` is to simplify filtering the correct features for a given plan. With `features` field, the same logic can be applied to all plans regardless of the plan type.

Comment (lyyder) [inline app/services/plan_service/store/plan.rb]: @rap1ds sorry, misinterpreted your intention. Yes it does make sense not to store that information.

**Comment by lyyder** [comment]
Applied @rap1ds improvements.

File: app/services/plan_service/data_types.rb [hunk 3]
```diff
@@ -18,6 +20,8 @@ module PlanService::DataTypes
     [:marketplace_plan_id, :fixnum, :mandatory],
     [:marketplace_id, :fixnum, :mandatory],
     [:plan_level, :fixnum, :mandatory],
+    [:features, :hash, :mandatory],
+    [:member_limit, :fixnum, :optional],
```
rap1ds: This `ExternalPlan` entity is admittedly badly named. It is used to pass initial trial data to Charm in the login link JSON Web Token payload. `features` and `member_limit` (and actually also `plan_level`) are unnecessary information and could (should?) be removed from the payload.

app/services/plan_service/data_types.rb
```diff
@@ -9,6 +9,8 @@ module PlanService::DataTypes
     [:id, :fixnum, :optional], # For OS, the plan is not actually in DB. Thus, optional.
     [:community_id, :fixnum, :mandatory],
     [:plan_level, :fixnum, :mandatory],
+    [:features, :hash, :mandatory],
```
rap1ds: This could be also an Entity

db/migrate/20160513124246_remove_features_from_marketplace_trials.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveFeaturesFromMarketplaceTrials < ActiveRecord::Migration
+  def change
+    remove_column :marketplace_trials, :features, :text, after: :community_id
```
rap1ds: The "Add features" and "Remove features" migrations could be removed. (and also the member limit)

app/services/plan_service/data_types.rb
rap1ds: Btw. if you decide to remove the `plan_level` from this entity, then the external plan service Entity type for this must be also changed.

app/services/plan_service/api/no_plans.rb
```diff
@@ -25,6 +25,7 @@ def get_current(community_id:)
         Plan.call(
         community_id: community_id,
         plan_level: PlanService::Levels::OS,
+        features: { all: true },
```
rap1ds: This is now not according to the Entity spec

PR #2025, opened 2016-05-10
Use I18n-js to export translation bundle

rap1ds:
Purpose: This PR tries to solve the problem of passing translations to front-end code (JavaScript) and using interpolation, pluralization, date formatting e.g. in JS.

# Client-side translations

_This documentation describes how to use translations in the client-side JavaScript code. The document applied only to the new React based JavaScript code._

## Usage

In the React component, you need to import `t` function from the `i18n` module. After that you can use all the translation keys from the `en.js.*` scope:

```js
// MyReactComponent.js

import { t } from '../../utils/i18n';

class MyReactComponent extends Component {
  render() {
    return span({className: "hello-world"}, t("js.hello_world")
  }
}
```

The example above assumed that there's a `js.hello_world` key in the translation YML file, e.g.:

```yml

# en.yml

en:
  js:
    hello_world: "Hello JavaScript world!"
```

### I18n-js

The client-side translations are powered by [i18n-js](https://github.com/fnando/i18n-js/) gem. The gem provides three important utilities:

* `rake i18n:js:export` task to export the translations to `.js` bundle
* `i18n-js` npm package, which helpers for translations, pluralizations, etc.
* `I18n::JS::Middleware` which compiles the `.js` bundle every request in development mode

## Client-side rendering

The translation bundle for all languages is big, so it makes sense to split it per language for client-side rendering purposes. In addition, the translation bundle can be cached because it doesn't change between deploys.

In production mode, you should see two `<script>` tags, one for the language bundle and one for the `application.js` with the fingerprint in the filename:

```html
<script src="https://your_cdn.com/assets/i18n/en-9b9ce41ada0d1b7ad028dda2c64c23d8.js"></script>
<script src="https://your_cdn.com/assets/application-6e9fbeebcaa14c12939b47fab1e53769.js"></script>
```

## Server-side rendering

When doing server-side rendering, the `utils/i18n.js` file takes care of loading the translation bundle `all.js`.

### Deployment to production

`rake i18n:js:export` task is configured so that it's always called before `rake assets:precompile`. So no extra steps are needed during the deployment.

Todo:

- [X] Create bundles per language
- [X] In development environment, create bundles on fly
- [X] Run `rake i18n:js:export` every time before `assets:precompile`
- [X] Include global `I18n` library
- [X] Change Onboarding topbar to use `I18n`
- [x] Server rendering
- [x] Documentation
- [ ] Remove `utils/i18nUtils.js`
- [ ] Change the guide to use `I18n`
- [x] Consider changing the `i18n.js` utility so that we can also export only the `t` function.
- [x] App configuration (middleware)

### DIFF
--- a/app/views/layouts/application.haml
+++ b/app/views/layouts/application.haml
@@ -61,6 +61,12 @@
   -# load Facebook SDK for the whole page
   = render :partial => "layouts/facebook_sdk"
 
+  -# Initialize global I18n variable and load translations there
+  :javascript
+    window.I18n = {};
+
+  = javascript_include_tag "i18n/#{I18n.locale}"
+
   = javascript_include_tag 'application'
 
   :javascript
--- a/client/app/components/OnboardingTopBar/OnboardingTopBar.js
+++ b/client/app/components/OnboardingTopBar/OnboardingTopBar.js
@@ -1,25 +1,54 @@
 import { Component, PropTypes } from 'react';
-import { translate } from '../../utils/i18nUtils';
 import { div, span, a } from 'r-dom';
+import { t } from '../../utils/i18n';
 
 import css from './OnboardingTopBar.scss';
 
-const next = function next(nextStep, guideRoot, t) {
-  return t(nextStep) ? {
-    title: t(nextStep),
-    link: `${guideRoot}/${nextStep}`,
-  } : null;
+const next = function next(nextStep, guideRoot) {
+  switch (nextStep) {
+    case 'slogan_and_description':
+      return {
+        title: t('admin.onboarding.topbar.slogan_and_description'),
+        link: `${guideRoot}/${nextStep}`,
+      };
+    case 'cover_photo':
+      return {
+        title: t('admin.onboarding.topbar.cover_photo'),
+        link: `${guideRoot}/${nextStep}`,
+      };
+    case 'filter':
+      return {
+        title: t('admin.onboarding.topbar.filter'),
+        link: `${guideRoot}/${nextStep}`,
+      };
+    case 'paypal':
+      return {
+        title: t('admin.onboarding.topbar.paypal'),
+        link: `${guideRoot}/${nextStep}`,
+      };
+    case 'listing':
+      return {
+        title: t('admin.onboarding.topbar.listing'),
+        link: `${guideRoot}/${nextStep}`,
+      };
+    case 'invitation':
+      return {
+        title: t('admin.onboarding.topbar.invitation'),
+        link: `${guideRoot}/${nextStep}`,
+      };
+    default:
+      return null;
+  }
 };
 
 class OnboardingTopBar extends Component {
 
   nextElement() {
-    const t = translate(this.props.translations);
-    const nextStep = next(this.props.next_step, this.props.guide_root, t);
+    const nextStep = next(this.props.next_step, this.props.guide_root);
     if (nextStep) {
       return (
         div({ className: css.nextContainer }, [
-          div({ className: css.nextLabel }, t('next_step')),
+          div({ className: css.nextLabel }, t('admin.onboarding.topbar.next_step')),
           a({ href: nextStep.link, className: css.nextButton }, [
             span(nextStep.title),
           ]),
@@ -30,12 +59,11 @@ class OnboardingTopBar extends Component {
   }
 
   render() {
-    const t = translate(this.props.translations);
     const currentProgress = this.props.progress;
     return div({ className: css.topbarContainer }, [
       div({ className: css.topbar }, [
         a({ className: css.progressLabel, href: this.props.guide_root }, [
-          t('progress_label'),
+          t('admin.onboarding.topbar.progress_label'),
           span({ className: css.progressLabelPercentage },
                `${Math.floor(currentProgress)} %`),
         ]),
@@ -49,7 +77,6 @@ class OnboardingTopBar extends Component {
 }
 
 OnboardingTopBar.propTypes = {
-  translations: PropTypes.objectOf(PropTypes.string).isRequired,
   guide_root: PropTypes.string.isRequired,
   progress: PropTypes.number.isRequired,
   next_step: PropTypes.string.isRequired,
--- a/client/app/utils/i18n.js
+++ b/client/app/utils/i18n.js
@@ -1,0 +1,53 @@
+// This file has three tasks:
+//
+// 1. Initialize global.I18n if we are in server environment
+// 1. Load the language bundle if we are in server environment
+// 2. Require the i18n-js library and export it
+//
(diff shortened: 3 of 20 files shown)


client/app/components/OnboardingTopBar/OnboardingTopBar.js
```diff
@@ -4,22 +4,51 @@ import { div, span, a } from 'r-dom';
...
+      link: `${guideRoot}/${nextStep}`
+    };
+  case "invitation":
+    return {
+      title: I18n.t("admin.onboarding.topbar.invitation"),
+      link: `${guideRoot}/${nextStep}`
+    };
+  default:
+    return null;
+  }
```
kpuputti: Why is the generic function "opened up" to this switch case that has hard-coded nextStep values?

rap1ds: For greppability. The current RoR has tons of old unused translations in the translation files, because the `t` function is called with a variable. So I'd prefer hard-coding the translations keys so that they can be grep'd and removed in the future.

Gnito:
Related to server side rendering problem. Is the language file just a hash?
This might be the correct way to pass Rails-side context-data to React components:
https://github.com/sharetribe/sharetribe/blob/master/config/initializers/react_on_rails.rb#L9

rap1ds:
@Gnito To my understanding, `railsContext` is also (the same way as `props`) rendered by server to some hidden element which will the later read by JS. Is that right?

The README says this, so that's why I concluded that the `railsContext` must be rendered to the HTML:

> This information (props and railsContext) should be the same regardless of either client or server side rendering.

I'd like to avoid that. The language bundle will grow bigger and the information rarely changes (only between deployes) so would like to avoid growing the initial page load payload.

client/app/components/OnboardingTopBar/OnboardingTopBar.js
Gnito: We are now tightly coupling frontend and backend code. :(
All client devs (web apps, mobile apps etc.) need to understand Rails-side translation context and they need to write their code in a way that Rails-side changes can be grepped from their code-base.

Related:
I also think that we should think about hard-coding all the needed locales to PropTypes of Page-component, 
```
Page.propTypes = {
  translations: PropTypes.objectOf(PropTypes.shape({
    admin: PropTypes.objectOf(PropTypes.shape({
      onboarding: PropTypes.objectOf(PropTypes.shape({
        topbar: PropTypes.objectOf(PropTypes.shape({
          slogan_and_description: PropTypes.string.isRequired,
        }).isRequired,
      }).isRequired,
    }).isRequired,
  }).isRequired,
};
```
Looks a bit heavy, but that way we get an alert if something is missing while rendering.

That method doesn't say that we should pass translations as props from haml/erb - in startup / registration step we could figure out what translations we actually need and pass them as props from there to actual Page component.

client/app/utils/i18n.js
```diff
@@ -0,0 +1,39 @@
...
+  try {
+    require('../../../app/assets/javascripts/i18n/all.js'); // eslint-disable-line no-undef
+  } catch (e) {
+    console.warn("Can't load language bundle all.js"); // eslint-disable-line no-console
+  }
+}
+
+// It seems that we need to use `require` here instead of `import`.
+// With `import` the translations seem not to be populated.
+const I18n = require('i18n-js'); // eslint-disable-line no-undef
```
Gnito: 'Require' loadings are synchronous, maintaining the order. However, I don't understand why this is the order needed - I thought it was supposed to be the other way around. Btw. have you tried loading the page with throttling connection?

app/views/layouts/application.haml
```diff
@@ -61,6 +61,7 @@
   -# load Facebook SDK for the whole page
   = render :partial => "layouts/facebook_sdk"
 
+  = javascript_include_tag "i18n/#{I18n.locale}"
```
Gnito: Add this also to the views/layouts/styleguide.html.erb

client/app/utils/i18n.js
rap1ds: Ok! That explains the difference between `import` and `require`. Thanks!

Good questions about the order. I made some changes to the code and try to simplify this a bit.

Actually, the order doesn't matter anymore. But what *does* matter is that the global `I18n` needs to be initialized before loading the `i18n-js` library. See this line in the `i18n-js` code: https://github.com/fnando/i18n-js/blob/2ca6d31365bb41db21e373d126cac00d38d15144/app/assets/javascripts/i18n.js#L31

If the global `I18n` is not initialized before loading the library, the library will create a new object and the translations get loaded to another object.

I haven't tried loading the page with throttled connection. Should I?

app/views/layouts/application.haml
rap1ds: Will do

config/i18n-js.yml
```diff
@@ -0,0 +1,37 @@
...
+
+  # Bundles per language
+  #
+  # Per language bundles are used in frontend because we don't
+  # want to send too big language bundles over the wire
+  #
+  # Use a custom namespace that uses either global (in node) or window (in browser)
+
+  - file: "app/assets/javascripts/i18n/%{locale}.js"
+    only: ["*.js", "*.admin.onboarding"]
```
rap1ds: Maybe `web` is a better scope name

client/webpack.client.rails.build.config.js
```diff
@@ -14,7 +14,7 @@ config.module.loaders.push(
   {
     test: /\.js$/,
     loader: 'babel-loader',
-    exclude: /node_modules/,
+    exclude: [/node_modules/, /i18n\/all.js/],
```
rap1ds: Is this needed when we have the ignore plugin?

config/i18n-js.yml
```diff
@@ -0,0 +1,37 @@
...
+
+  # Bundle for all languages
+  #
+  # This bundle is used for server rendering. We can bundle
+  # all languages in on file because we don't need to care
+  # about the filesize
+  #
+  # Use a custom namespace that uses either global (in node) or window (in browser)
+
+  - file: "app/assets/javascripts/i18n/all.js"
```
rap1ds: `all.js` shouldn't be under `app/assets`

lib/tasks/assets.rake
```diff
@@ -23,4 +23,4 @@ namespace :assets do
   task :clobber do
     rm_r Dir.glob(Rails.root.join("app/assets/webpack/*"))
```
rap1ds: `clobber` should also delete `app/assets/i18n`

rap1ds: Does this really extend the task?

ovan:
Add to documentation that this doesn't work for per marketplace translations.

ovan:
Looks good to me now!

PR #2026, opened 2016-05-10
Update nodejs and npm packages

kpuputti:
Update Node.js and NPM packages.

When using NVM:

```bash
nvm install 6.1
nvm alias default 6.1
cd client
npm install
npm rebuild # not sure if this is needed, but doesn't do any harm
```


mporkola:
Looks good. This apparently needs some manual environment changes, so UPGRADE.MD should be updated as well.

kpuputti:
Summary from IRL conversation: the "Unreleased' sections in UPGRADE.md and CHANGELOG.md already have the React on Rails integration and this can be part of that, not a separate change.

PR #2034, opened 2016-05-12
Add PostCSS plugin: Rails image-path helper fix

Gnito:


kpuputti:
Quite a few linting errors, check the CI output.

client/plugins/postcss-image-path.js
```diff
@@ -0,0 +1,38 @@
...
+module.exports = postcss.plugin('postcss-image-path', function (opts) {
+  opts = opts || {};
+  const imagePathHelper = opts.helper || 'image-path';
+  const assetPath = opts.assetPath || process.cwd();
+
+  return function (css, result) {
+    css.walkDecls(function (decl) {
+      // Return if css declaration value part doesn't contain Rails helper
+      // e.g. image-path('image.png')
+      if (!decl.value || decl.value.indexOf( imagePathHelper + '(') === -1) {
```
kpuputti: Can the `decl.value` se something else than a String? If it is truthy, but not a String, the `.indexOf` method might not be found.

kpuputti: Our version of Node.js has the `String.prototype.startsWith` method, so that can be used here as well.

client/plugins/postcss-image-path.js
```diff
@@ -0,0 +1,38 @@
...
+      // Figure out relative path to assets within webpack
+      // Images should be found from ./client/app/assets/images/
+      const relativePath = path.relative(
+        path.dirname(filePath),
+        process.cwd() + assetPath
+      );
+
+      // replace: background-image: url(image-path("img.png"))
+      // => background-image: url("../../assets/images/img.png")
+      const re = /(image-path\(["'](.*?)["']\))/
```
kpuputti: You might need the `g` flag for the RegExp.

client/plugins/postcss-image-path.js
Gnito: 1) As far as I know, it can't.  (and it's tested quite well already: this goes through every line of our current css without errors.)
2) the value doesn't start with 'image-path(' but with something else like "url(image-path("

client/plugins/postcss-image-path.js
Gnito: You are absolutely right - I'll change!

PR #2036, opened 2016-05-13
Steps to array

rap1ds:
The data type for onboarding steps is currently a Hash, and that creates some issues:

- In Hashes (or Object, in JS language) the order of the Hash items is not clearly defined. In some languages, the order of the Hash can not be relied. Steps have a clear order, so that's why Array is more suitable data structure.
- Hashes are more difficult to handle in code for this particular case. There are some cumbersome code that takes the Object, iterates the keys, takes the value from the object by the key and then desides what to do with the value. In array, this is a simple call to `find` function.

In addition, this PR removes one duplicated validation, where the container component (that only passes the data forward) does data validation.


client/app/components/onboardingGuide/GuideStatusPage.js
```diff
@@ -89,7 +90,7 @@ GuideStatusPage.propTypes = {
     }),
     PropTypes.bool.isRequired,
   ]).isRequired,
-  onboarding_data: PropTypes.objectOf(PropTypes.shape({
+  onboarding_data: PropTypes.arrayOf(PropTypes.shape({
     info_image: PropTypes.string,
     cta: PropTypes.string,
     complete: PropTypes.bool.isRequired,
```
kpuputti: You should add the new key `step` to the validation as well. This should be caught in the linting by eslint-plugin-react.

kpuputti: On another thought, eslint-plugin-react can most likely only complain about missing PropTypes for root-level props, not the data within those.

client/app/components/onboardingGuide/OnboardingGuide.js
```diff
@@ -196,12 +198,23 @@ OnboardingGuide.propTypes = {
...
+          'listing',
+          'invitation',
+          'all_done',
+        ]).isRequired,
+        info_image: string,
+        cta: string.isRequired,
+        alternative_cta: string,
+        complete: bool.isRequired,
+      }).isRequired
+    ).isRequired,
```
kpuputti: This is really nice 👍

PR #2044, opened 2016-05-16
Remove HTML from translations

rap1ds:
- Guide uses new t function and interpolation
- Change the onboarding guide to use the new `t` function
- Remove all markup from translations
- Use forked version of i18n-js, with interpolationMode
- Move all the JS translations under `web.*` scope
- Remove `i18nUtils.js` as it's not needed anymore
- ESLint: Add global `process` and `commonjs` environment

Might be also good idea to review the forked i18n-js: https://github.com/fnando/i18n-js/compare/master...sharetribe:interpolation-mode

### DIFF
--- a/client/.eslintrc.js
+++ b/client/.eslintrc.js
@@ -23,6 +23,11 @@ module.exports = {
   env: {
     browser: true,
     es6: true,
+    commonjs: true
+  },
+  globals: {
+    // Webpack plugin adds `process` global
+    process: false
   },
   rules: {
     // See http://eslint.org/docs/rules/ for documentation for
--- a/client/.eslintrc.js
+++ b/client/.eslintrc.js
@@ -26,7 +26,8 @@ module.exports = {
     commonjs: true
   },
   globals: {
-    // Webpack plugin adds `process` global
+    // `process` global is added by Webpack plugin.
+    // Tell the existence of `process` to ESLint
     process: false
   },
   rules: {
--- a/client/.eslintrc.js
+++ b/client/.eslintrc.js
@@ -23,7 +23,6 @@ module.exports = {
   env: {
     browser: true,
     es6: true,
-    commonjs: true
   },
   globals: {
     // `process` global is added by Webpack plugin.
--- a/client/app/components/onboardingGuide/GuideCoverPhotoPage.js
+++ b/client/app/components/onboardingGuide/GuideCoverPhotoPage.js
@@ -1,5 +1,5 @@
 import { PropTypes } from 'react';
-import r, { div, h2, p, img, a, span, br } from 'r-dom';
+import r, { div, h2, p, img, a, br } from 'r-dom';
 import css from './styles.scss';
 import { t } from '../../utils/i18n';
 
@@ -31,17 +31,14 @@ const GuideCoverPhotoPage = (props) => {
       div({
         className: css.infoTextContent,
       }, [
-        span(
-          t('web.admin.onboarding.guide.cover_photo.advice.content1',
-            { link: a({
-              href: 'http://support.sharetribe.com/knowledgebase/articles/744438',
-              target: '_blank',
-              alt: t('web.admin.onboarding.guide.cover_photo.advice.alt'),
-            }, t('web.admin.onboarding.guide.cover_photo.advice.link')) })),
+        t('web.admin.onboarding.guide.cover_photo.advice.content1',
+          { link: a({
+            href: 'http://support.sharetribe.com/knowledgebase/articles/744438',
+            target: '_blank',
+            alt: t('web.admin.onboarding.guide.cover_photo.advice.alt'),
+          }, t('web.admin.onboarding.guide.cover_photo.advice.link')) }),
         br(),
-        span(
-          t('web.admin.onboarding.guide.cover_photo.advice.content2', { width: 1920, height: 450 })
-        ),
+        t('web.admin.onboarding.guide.cover_photo.advice.content2', { width: 1920, height: 450 }),
       ]),
     ]),
 
--- a/client/app/components/onboardingGuide/GuideCoverPhotoPage.js
+++ b/client/app/components/onboardingGuide/GuideCoverPhotoPage.js
@@ -35,6 +35,7 @@ const GuideCoverPhotoPage = (props) => {
           { link: a({
             href: 'http://support.sharetribe.com/knowledgebase/articles/744438',
             target: '_blank',
+            rel: 'noreferrer',
             alt: t('web.admin.onboarding.guide.cover_photo.advice.alt'),
           }, t('web.admin.onboarding.guide.cover_photo.advice.link')) }),
         br(),
--- a/client/app/components/onboardingGuide/GuideCoverPhotoPage.js
+++ b/client/app/components/onboardingGuide/GuideCoverPhotoPage.js
@@ -5,6 +5,9 @@ import { t } from '../../utils/i18n';
 
 import GuideBackToTodoLink from './GuideBackToTodoLink';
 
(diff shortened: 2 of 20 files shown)


client/.eslintrc.js
```diff
@@ -23,6 +23,11 @@ module.exports = {
   env: {
     browser: true,
     es6: true,
+    commonjs: true
```
kpuputti: We might not want to allow using this env everywhere. Was there too many files where you would've had to add an explicit `/* eslint-env commonjs */` in the top?

client/.eslintrc.js
```diff
@@ -23,6 +23,11 @@ module.exports = {
...
+  },
+  globals: {
+    // Webpack plugin adds `process` global
+    process: false
```
kpuputti: This probably doesn't do what expected.

http://eslint.org/docs/user-guide/configuring#specifying-globals

According to the documentation, specifying `process: false` in the globals actually defines a global called `process`! the false value just disallows overwriting it.

client/app/components/onboardingGuide/GuideCoverPhotoPage.js
```diff
@@ -29,18 +30,25 @@ const GuideCoverPhotoPage = (props) => {
       }),
       div({
         className: css.infoTextContent,
-        dangerouslySetInnerHTML: { __html: t('advice') }, // eslint-disable-line react/no-danger
-      }),
+      }, [
+        t('web.admin.onboarding.guide.cover_photo.advice.content1',
+          { link: a({
+            href: 'http://support.sharetribe.com/knowledgebase/articles/744438',
+            target: '_blank',
```
kpuputti: Maybe not relevant in this case (although I'd still change it to be sure), but I'll remind about this every time I see a link with `target="_blank"`:

https://mathiasbynens.github.io/rel-noopener/

client/app/components/onboardingGuide/GuideCoverPhotoPage.js
```diff
@@ -29,18 +30,25 @@ const GuideCoverPhotoPage = (props) => {
...
+            alt: t('web.admin.onboarding.guide.cover_photo.advice.alt'),
+          }, t('web.admin.onboarding.guide.cover_photo.advice.link')) }),
+        br(),
+        t('web.admin.onboarding.guide.cover_photo.advice.content2', { width: 1920, height: 450 }),
```
kpuputti: What are these magic dimensions?

client/app/components/onboardingGuide/GuideInvitationPage.js
```diff
@@ -1,26 +1,29 @@
...
-      className: css.description,
-      dangerouslySetInnerHTML: { __html: t('description') }, // eslint-disable-line react/no-danger }
-    }),
+    r(GuideBackToTodoLink, { changePage, initialPath }),
+    h2({ className: css.title }, t('web.admin.onboarding.guide.invitation.title')),
+    p({ className: css.description }, t('web.admin.onboarding.guide.invitation.description.content', {
+      preview_link: a({
+        href: '/?big_cover_photo=true',
+        alt: t('web.admin.onboarding.guide.invitation.description.preview_link_alt'),
+        target: '_blank',
```
kpuputti: Why do we open an internal link to a new tab?

client/app/components/onboardingGuide/GuideStatusPage.js
```diff
@@ -85,13 +108,10 @@ GuideStatusPage.propTypes = {
...
-    shape({
-      title: string.isRequired,
-      link: string.isRequired,
-    }),
-    bool.isRequired,
-  ]).isRequired,
+  nextStep: shape({
+    title: string.isRequired,
+    link: string.isRequired,
+  }),
```
kpuputti: The `nextStep` shape itself should probably have `.isRequired` as well.

kpuputti:
Nice work, great to get rid of all the `dangerouslySetInnerHTML` usage!

I see some LISP influence in the paren positioning, and long translation keys might become quite painful to write, but I trust your judgement on those. :)

client/app/components/onboardingGuide/GuideStatusPage.js
rap1ds: Previously the function that returns the `nextStep` returned either a shape of `false`. So the `oneOf [ shape or bool ]` was set to be required. Now I change the function so that it returns either shape or `null`.

client/app/components/onboardingGuide/GuideInvitationPage.js
rap1ds: Hmh... no idea. That's how it was. I guess the idea was to leave the user where she was and open the preview in another tab?

client/app/components/onboardingGuide/GuideCoverPhotoPage.js
rap1ds: Those are ideal cover photo sized. They will be interpolated to the translation:

`"Note: the photo will be resized to %{width}x%{height} pixels. The top and bottom sections of images that are taller will be cropped."` becames `"Note: the photo will be resized to 1920x450 pixels. The top and bottom sections of images that are taller will be cropped."`

I think the dimensions should not be in the translation itself. If we for example decide to change the dimensions and they are in hard-coded in the translations, that would mean changing all the translations.

kpuputti: Maybe I was a bit too implicit here. I'm not after what those dimensions mean here in the review, but they should be clear from the code, e.g. `const COVER_PHOTO_DIMENSIONS = { width: 1920, height: 450 };`.

client/app/components/onboardingGuide/GuideInvitationPage.js
kpuputti: Ok for me if it already was designed like that. Just couldn't see it in the diff since the old link was somewhere else in the translations.

client/app/components/onboardingGuide/GuideCoverPhotoPage.js
rap1ds: Good point. I'll change those.

client/.eslintrc.js
rap1ds: No, not too many, only one :)

In ESLint docs, it says this about the `commonjs` environment:

> commonjs - CommonJS global variables and CommonJS scoping (use this for browser-only code that uses Browserify/WebPack).

So since we are actually using WebPack, is there any reason why we shouldn't use this environment?

client/.eslintrc.js
rap1ds: I understood the "define" as in "tell ESLint about the existence of this variable", not as in "define a new global variable called `process`. And yeah, setting the value `false` to my understanding means that all attempts to change the `process`, e.g. `process.env.NODE_ENV = "development"` are forbidden.

client/app/components/onboardingGuide/GuideCoverPhotoPage.js
rap1ds: Good point. I'll change that.

client/.eslintrc.js
kpuputti: I'm not sure what the scoping means and how it relates to us using ES modules, but I'd just want to avoid exposing e.g. the `require` function when an import should be used. Not a big deal, though.

client/.eslintrc.js
kpuputti: Badly worded comment from me, I meant exactly that it defines the existence of a global for ESLint to know.

Based on the comment in the code, it looked like the process global was attempted to be removed from the global scope. But apparently you knew exactly what you were doing and I made an incorrect interpretation. :)

PR #2045, opened 2016-05-16
Remove 'window' reference

rap1ds:


client/app/utils/i18n.js
```diff
@@ -13,35 +13,33 @@
...
-    require('../i18n/all.js'); // eslint-disable-line no-undef
-  } catch (e) {
-    console.warn("Can't load language bundle all.js"); // eslint-disable-line no-console
-  }
-}
+// Initialize global.I18n
+// This might be initialized already, because in browser
+// we do initialize this before the translations get loaded
+// in a separate script tag
+global.I18n = global.I18n || {}; // eslint-disable-line no-undef
```
kpuputti: Now that this is in the top level module scope, what does `global` mean in the browser's context?

rap1ds: It means `Window` in browser. I tried it :)

kpuputti: screen shot 2016-05-16 at 11 40 18

Are you sure? I get the above error.

rap1ds: I get the same error when I try `global` in browser console.

But when I add this to the `i18n.js` file:

screen shot 2016-05-16 at 11 47 41

I see this in the console:

screen shot 2016-05-16 at 11 47 30

rap1ds:
Seems that this PR is not solving the issue with `window ReferenceError`. Closing.

Issue #2043, opened 2016-05-16
Error: Cannot find module "./app/startup/serverRegistration" (ExecJS::ProgramError)


wolfieorama:
##### Sharetribe Version:
86eed984d42b88c62eb791b55e6c5d03740d2c8e

##### Environment:

Linux ubuntu 14.04 64 bit
##### Server mode:

development 
##### Description:

when running the procfiles i am getting the following error:

 `webpackMissingModule ((execjs):79:43): Error: Cannot find module "./app/startup/serverRegistration" (ExecJS::ProgramError)`

##### Steps To Reproduce:

_List the actions that will reproduce the issue._

1. step 13 and 14 of installation 

##### Additional information

my log below:

john@john-Lenovo-G40-30:~/sharetribe$ foreman start -f Procfile.static
07:59:20 rails.1               | started with pid 8649
07:59:20 rails-client-assets.1 | started with pid 8650
07:59:20 rails-server-assets.1 | started with pid 8652
07:59:22 rails-client-assets.1 | 
07:59:22 rails-client-assets.1 | > @ build:dev:client /home/john/sharetribe/client
07:59:22 rails-client-assets.1 | > webpack -w --config webpack.client.rails.build.config.js
07:59:22 rails-client-assets.1 | 
07:59:22 rails-server-assets.1 | 
07:59:22 rails-server-assets.1 | > @ build:dev:server /home/john/sharetribe/client
07:59:22 rails-server-assets.1 | > webpack -w --config webpack.server.rails.build.config.js
07:59:22 rails-server-assets.1 | 
07:59:26 rails-client-assets.1 | Webpack dev build for Rails
07:59:32 rails.1               | => Booting WEBrick
07:59:32 rails.1               | => Rails 4.2.5.2 application starting in development on http://0.0.0.0:3000
07:59:32 rails.1               | => Run `rails server -h` for more startup options
07:59:32 rails.1               | => Ctrl-C to shutdown server
07:59:32 rails.1               | ERROR when compiling base_js_code! See file tmp/base_js_code.js to correlate line numbers of error. Error is
07:59:32 rails.1               | 
07:59:32 rails.1               | Error: Cannot find module "./app/startup/serverRegistration"
07:59:32 rails.1               | 
07:59:32 rails.1               | webpackMissingModule ((execjs):79:43)
07:59:32 rails.1               | Object.<anonymous> ((execjs):79:114)
[...]

Issue #2037, opened 2016-05-13
can't start server in dev environment on mac OS X


copremesis:
##### Sharetribe Version:

_Share the git commit hash of the version of Sharetribe that you are using. You can find it by running `git rev-parse HEAD` in your Sharetribe directory._

```
6ca0d2ee4d527a81f616b0968b71bf35a16e33b5
```
##### Environment:

_What's your environment?_  **Mac OS X**
```Darwin ethereal.local 13.4.0 Darwin Kernel Version 13.4.0: Sun Aug 17 19:50:11 PDT 2014; root:xnu-2422.115.4~1/RELEASE_X86_64 x86_64```

```Homebrew 0.9.5 (git revision 3c896; last commit 2016-03-01)```
```bash-3.2$ node -v
v0.10.33```
```bash-3.2$ npm -v
2.1.14```

##### Server mode:

development

##### Description:

when attempting to run a job or server I'm getting this error message

```bash
ERROR when compiling base_js_code! See file tmp/base_js_code.js to correlate line numbers of error. Error is

Error: Cannot find module "./app/startup/serverRegistration"

webpackMissingModule ((execjs):79:43)
Object.<anonymous> ((execjs):79:114)
__webpack_require__ ((execjs):51:30)
```

##### Steps To Reproduce:

attempt any of the following:

```
rake jobs:work
```
```
foreman start -f Procfile.static
```

Any help would be greatly appreciated

mporkola:
Thanks for the report,

You seem to be running an old Node version, we currently require Node 6.1. This post on our forum might help you: https://www.sharetribe.com/community/t/help-webpack-and-react/151/2

copremesis:
now running 
```

bash-3.2$ node -v
v6.1.0
bash-3.2$ npm -v
3.9.0
```


still getting 


```
ethereal:sharetribe rob$ rake jobs:work
ERROR when compiling base_js_code! See file tmp/base_js_code.js to correlate line numbers of error. Error is

Error: Cannot find module "./app/startup/serverRegistration"

webpackMissingModule ((execjs):79:43)
Object.<anonymous> ((execjs):79:114)
__webpack_require__ ((execjs):51:30)

```


so it's not the code it's node? ... should I blow out the npm modules and rebuild?

rap1ds:
Why is React on Rails doing something when running normal `rake` command that is totally unrelated to React?

copremesis:
[...]

PR #2046, opened 2016-05-16
Wait for Webpack bundle files before starting Rails server

mporkola:
Our Rails server depends on Webpack-built frontend bundles. When starting Rails server and Webpack builds in parallel we have no guarantees that the Webpack builds are actually completed before Rails gets interested in them, which results in timing dependent startup crashes.

This PR makes Rails server wait for Webpack to build server JS bundle before starting.

Fixes https://github.com/sharetribe/sharetribe/issues/2043 and partially https://github.com/sharetribe/sharetribe/issues/2037.

### DIFF
--- a/Procfile.static
+++ b/Procfile.static
@@ -1,5 +1,5 @@
-# Run Rails without hot reloading (static assets).
-rails: REACT_ON_RAILS_ENV= rails s -b 0.0.0.0
+# Run Rails server after Webpack build is ready.
+rails: script/wait_for_server_js.sh && rails s -b 0.0.0.0
 
 # Build client assets, watching for changes.
 rails-client-assets: cd client && npm run build:dev:client
--- a/script/wait_for_file.sh
+++ b/script/wait_for_file.sh
@@ -1,0 +1,8 @@
+wait_for_file() {
+  local file="$1"; shift
+  local wait_seconds="${1:-10}"; shift # 10 seconds as default timeout
+
+  until [ $((wait_seconds--)) -eq 0 -o -f "$file" ] ; do sleep 1; done
+
+  [ "$wait_seconds" -ne -1 ]
+}
--- a/Procfile.hot
+++ b/Procfile.hot
@@ -1,6 +1,6 @@
 # Procfile for development with hot reloading of JavaScript and CSS
 
-rails: rails s -b 0.0.0.0
+rails: script/wait_for_server_js.sh && rails s -b 0.0.0.0
 
 # Run the React Storybook for client development
 storybook: cd client && npm run styleguide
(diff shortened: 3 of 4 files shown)


rap1ds:
Are you planning to make the same change for all Procfiles? Or only for .static?

Procfile.static
```diff
@@ -1,8 +1,4 @@
...
-
-# Build client assets, watching for changes.
-rails-client-assets: cd client && npm run build:dev:client
-
-# Build server assets, watching for changes. Remove if not server rendering.
-rails-server-assets: cd client && npm run build:dev:server
+# 1. Build client assets, watching for changes.
+# 2. Build server assets, watching for changes. Remove if not server rendering.
+# 3. Run Rails server.
+web: cd client/ && npm run build:server && npm run build:client && cd .. && rails s
```
Gnito: I'm not sure how watches are affecting to this setup.

script/wait_for_file.sh
```diff
@@ -0,0 +1,8 @@
+wait_for_file() {
+  local file="$1"; shift
+  local wait_seconds="${1:-10}"; shift # 10 seconds as default timeout
+
+  until test $((wait_seconds--)) -eq 0 -o -f "$file" ; do sleep 1; done
+
+  ((++wait_seconds))
```
bladealslayer: this is outside of the loop, so timeout won't work

bladealslayer: What is the purpose of this?

bladealslayer: Ah, this is intended as the return value. However, in the other script you use it as `wait_for_file ... || error`. Your return code will always be non-zero, since you increment from 0 or larger integer.

Maybe last statement (return value) should be `[ "$wait_seconds" -ne 0 ]`?

bladealslayer: And I got confused by prefix / postfix `--` and `++`. So if timeout was reached, `wait_seconds` will actually be `-1`. Still, in shell 0 is true, and non-zero is false, so you have to negate the return value: 

```
[ "$wait_seconds" -ne -1]
```

script/wait_for_file.sh
```diff
@@ -0,0 +1,8 @@
+wait_for_file() {
+  local file="$1"; shift
+  local wait_seconds="${1:-10}"; shift # 10 seconds as default timeout
+
+  until test $((wait_seconds--)) -eq 0 -o -f "$file" ; do sleep 1; done
```
bladealslayer: Small race condition here - if file appears on last iteration, you'd miss this, effectively waiting only N-1 seconds for the file to appear. But probably not worth it complicating the loop just to catch that.

script/wait_for_file.sh
bladealslayer: And, obviously, I didn't know what the return value of `(( expression ))` is (From manual: If the value of the  expression  is non-zero, the return status is 0; otherwise the return status is 1.). So your code was correct, but then it even more seems magical. I'd prefer the more explicit test in the end.

PR #2047, opened 2016-05-17
Removed some unused translation keys

thomasmalbaux:
From earlier PayPal connection wizard, no longer used with the new one (deployed few months ago).


mporkola:
👍

PR #2048, opened 2016-05-17
Run Webpack before tests

mporkola:


rap1ds:
The bad thing about this is that it makes running the tests slower... I'm not completely sure whether we should add this or just add a documentation about "if the tests are failing because ... run rake assets:webpack"

ovan:
Yeah, I agree, let's not do this now. Slowing down the tests, especially in the case of rspec, feels bad. Let's see how painful this becomes and if better ideas surface.

PR #2051, opened 2016-05-18
React component cache

rap1ds:
Cache helper for React components

The cache key is generated from these values:

- component name
- props
- locale
- bundle file name (fingerprinted in production)

All the dynamic values (like current locale) that are not passed to the component via props (but via railsContext for example) should be included in the list of values that is used to generate the cache key.

Please note that if you turn on caching in development but do not run assets:precompile, the cache will not be invalidated because the server bundle filename doesn't change.


mporkola:
👍 except the dev mode digests issue

PR #2052, opened 2016-05-18
Use plan features and implement whitelabel branding

sktoiva:
- replaces plan level with features as determining factor for adding custom admin email and community deletion
 - introduces Sharetribe branding for non-whitelabel communities

## Branding

  1. Adds `powered by` info to every email footer:
  screen shot 2016-05-19 at 15 03 13

  2. Adds `powered by` info to all pages except admin pages
screen shot 2016-05-24 at 18 28 26

  3. Adds upgrade information as well as the `powered by` text to admin interface:
  screen shot 2016-05-19 at 15 07 08

## Remaining work

- [x] Add different message to admins that prompts them to buy pro plan for whitelabel removal.
- [x] Translation for admin email view branding message went missing after latest changes, add that


app/assets/stylesheets/main.css.scss
```diff
@@ -1925,6 +1925,27 @@ figure.fluidratio {
 }
 
 /*
+  Branding
+*/
+.branding-container {
+  clear:  both;
+
+  hr {
```
mporkola: I'd prefer a css class here, in this case using tag name seems very unlikely to cause issues but I'd like to stick to the principle of not using them except in resets.

app/assets/stylesheets/reset.scss
```diff
@@ -123,6 +123,10 @@ input, textarea, button, select {
   vertical-align: baseline;
   font-family: inherit; }
 
+select {
```
rap1ds: Is this necessary? Why is this added? How does this affect to all other `select` elements? Can we avoid adding this?

app/controllers/admin/communities_controller.rb
```diff
@@ -404,11 +406,11 @@ def delete_redirect_url(configs)
   end
 
   def can_delete_marketplace?(community_id)
-    PlanService::API::Api.plans.get_current(community_id: community_id).data[:plan_level] == PlanUtils::FREE
+    PlanService::API::Api.plans.get_current(community_id: community_id).data[:features][:deletable]
   end
 
   def can_set_sender_address(plan)
-    PlanUtils.valid_plan_at_least?(plan, PlanUtils::PRO)
+    plan[:features][:admin_email]
```
mporkola: very much 👍  for these changes

app/views/admin/communities/edit_welcome_email.haml
```diff
@@ -13,6 +13,16 @@
...
+  - if show_branding_info || !can_set_sender_address
+    .row
+      .col-12
+
+        .alert-box-warning
+          %p
+            %span.alert-box-icon<>
+              = icon_tag("alert", ["icon-fix"])
+            %span<
+              = t("admin.communities.outgoing_email.white_label_offer", upgrade_pro_plan_link: link_to(t("admin.communities.outgoing_email.upgrade_pro_plan_link"), admin_plan_path)).html_safe
```
mporkola: There appears to be a translation key clash

> Set your own email address as the sender and remove all Sharetribe branding from outgoing email messages by päivittää tilauksesi Pro-maksuvaihtoehtoomme..

mporkola:
Just something I started thinking: in application.haml we call display_branding_info?, which results in +1 DB query per pageview. Is this a performance hit we want to avoid? We could make plan changes touch community (= update updated_at column), and cache by that.

Otherwise 👍

app/controllers/application_controller.rb
```diff
@@ -621,6 +621,11 @@ def setup_logger!(metadata)
     logger.add_metadata(metadata)
   end
 
+  def display_branding_info?
+    !PlanService::API::Api.plans.get_current(community_id: @current_community.id).data[:features][:whitelabel]
```
rap1ds: Maybe you could use `@current_plan` and avoid DB hit

app/assets/stylesheets/reset.scss
sktoiva: Check the first comment here https://github.com/sharetribe/sharetribe/pull/2052 and the explanation about `select` tags. I hope it's not necessary, but I don't know how to circumvent it.

app/controllers/application_controller.rb
sktoiva: Yeah, good catch.

app/assets/stylesheets/main.css.scss
rap1ds: Another possibility is to use just `border-top` instead of `hr` tag.

app/assets/stylesheets/main.css.scss
```diff
@@ -1925,6 +1925,25 @@ figure.fluidratio {
 }
 
 /*
+  Branding
+*/
+.branding-container {
```
rap1ds: One more comment. Should we move these to separate `branding.css.scss`. I feel that the `main` is an unordered trash bin that we shouldn't grow anymore :)

sktoiva: Can do.

config/locales/en.yml
```diff
@@ -815,6 +815,10 @@ en:
       you_have_not_yet_accepted_or_rejected_offer: "You have not yet accepted or rejected the offer %{title} you received %{date}."
       you_have_not_yet_accepted_or_rejected_request: "You have not yet accepted or rejected the request %{title} you received %{date}."
       show_thread: "Show conversation"
+    branding:
```
rap1ds: Do we want to have there in the `en.yml`, which means that translators can translate this?

Gnito: Good question! And if we go to that direction, should the onboarding language folder be renamed as _marketing_ or _branding_ or _sharetribe_translations_

PR #2055, opened 2016-05-18
Stylelint setup

kpuputti:
This PR enables [stylelint](http://stylelint.io/) for linting CSS code within the `client/` directory.

The setup extends the [stylelint-config-standard](https://github.com/stylelint/stylelint-config-standard) configuration with the [suggested additions](https://github.com/stylelint/stylelint-config-standard#suggested-additions).

The `.css` files can be linted by running `npm run stylelint` within the `client/` directory or together with ESLint by running `npm run lint`.

### DIFF
--- a/travis/install.sh
+++ b/travis/install.sh
@@ -13,7 +13,7 @@ case "$SUITE" in
 esac
 
 case "$SUITE" in
-    cucumber|eslint)
+    cucumber|lint)
         echo "Installing and selecting Node.js version with nvm for suite: $SUITE"
         # shellcheck source=/dev/null
         . "$HOME/.nvm/nvm.sh"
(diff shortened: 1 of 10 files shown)


travis/install.sh
```diff
@@ -19,6 +19,7 @@ case "$SUITE" in
         . "$HOME/.nvm/nvm.sh"
         nvm install
         nvm use
+        nvm alias default $(cat .nvmrc)
```
kpuputti: This was an attempt to use the selected Node.js version from `.nvmrc`, but somehow the `npm run stylelint` still fails claiming it's using "v0.10.36" instead of "v6.1" that is defined in the `.nvmrc` file.

mporkola:
Good 👍

`client/README.md` should be updated as well, it still says

> You can run the linting with:
 `npm run lint`

kpuputti:
@mporkola Now updated.

PR #2058, opened 2016-05-19
Monkey-patch the using_i18n_fallbacks_module? method

rap1ds:
This Pull Request fixes an issue with fallbacks in JS translations. The fallbacks are not currently working. You can verify this by using `fr-CA` locale (Canadian French), which should fallback to `fr` but it falls back to `en`.

- [X] Correctly configure `I18n.available_locales`
- [X] Don't include the `Fallbacks` module to the `CommuityBackend`. It implements its own fallback logic in the `lookup` method
- [X] Add a new `Tag` implementation that returns the locale fallbacks without splitting the locale
- [X] Monkey-patch the `FallbackLocales.using_i18n_fallbacks_module?` so that it returns `true` if any of the backend in the backend chain has `Fallbacks` module included

### using_i18n_fallbacks_module?

We are using a chained I18n backend. Because of this, the `using_i18n_fallbacks_module?` was always returning `false`, because the chain backend doesn't have the `Fallbacks` module, instead, one of the backend in the backed chain may have the `Fallbacks` module included. This monkey-patch changes the method so that instead of asking from the Chain backend if the `Fallbacks` module is included, it iterates the whole chain and returns `true` if any of the backends has `Fallbacks` included.

See more: https://github.com/fnando/i18n-js/issues/406

### I18n::Locale::Tag::DefinedFallbacksOnly

`I18n::Locale::Tag::DefinedFallbacksOnly` is a `Tag` implementation which returns the locale fallbacks without splitting the locale. Here's the difference between this and the default `Simple` implementation:

```ruby
# Simple
tag_impl = I18n::Locale::Tag::Simple
expect(tag_impl.tag("fi-FI-helsinki").self_and_parents.map(&:to_s)).to eq(["fi-FI-helsinki", "fi-FI", "fi"])

# DefinedFallbacksOnly
tag_impl = I18n::Locale::Tag::DefinedFallbacksOnly
expect(tag_impl.tag("fi-FI-helsinki").self_and_parents.map(&:to_s)).to eq(["fi-FI-helsinki"])
```

The problem with the default implementation is that we have locales such as `da-DK`, but we don't have a locale `da`. Thus, the default implementation returns fallback locales that we don't have. Rails doesn't seem to care about this, but the I18nJs gem does.

### DIFF
--- a/config/initializers/i18n-js.rb
+++ b/config/initializers/i18n-js.rb
@@ -7,23 +7,23 @@
 module I18n
   module JS
 
-    def self.translations
-      all_translations = {}
-      selected_backends = []
+    def self.selected_backends
       current_backend = ::I18n.backend
 
       case current_backend
       when I18n::Backend::Chain
-        current_backend.backends.each do |backend_in_chain|
-          if backend_in_chain.respond_to?(:translations, true)
-            selected_backends << backend_in_chain
-          end
+        current_backend.backends.select do |backend_in_chain|
+          backend_in_chain.respond_to?(:translations, true)
         end
       else
         if current_backend.respond_to?(:translations, true)
-          selected_backends = [current_backend]
+          [current_backend]
         end
       end
+    end
+
+    def self.translations
+      all_translations = {}
 
       selected_backends.each do |selected_backend|
         selected_backend.instance_eval do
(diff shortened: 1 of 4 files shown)


config/initializers/i18n-js.rb
```diff
@@ -38,5 +38,16 @@ def self.translations
       all_translations
     end
 
+    class FallbackLocales
```
ovan: Where is this class used? Or are you opening up an existing class and adding methods to it?

rap1ds: Opening up existing class. Here's the original https://github.com/fnando/i18n-js/blob/master/lib/i18n/js/fallback_locales.rb

ovan:
Yay, this looks good now.

PR #2061, opened 2016-05-20
Routes for JavaScript

rap1ds:
#Todo:

- [X] Use js-routes
- [X] Integrate route exporting to Procfiles
- [X] Middleware
- [X] Docs
- [X] All routes from props replaced with Routes helper
- [X] Testing locally
- [X] Change routes so that they don't require community id
- [x] Remove the *all routes and make the routes explicit
- [x] Move `nextStep` logic from the `OnboardingGuide` container to the `GuideStatusPage`.
- [x] Server tells the app what is the current page (via `page` props)
- [x] Use `railsContext.pathname` to obtain the current path
- [x] Small refactoring: Move the `setPushState` out of the component (it doesn't use `this` to anything so we can take it out of the object scope)
- [x] Pass subset of Routes to child components
- [x] Styleguide
- [x] Remove `initializeEnv`
- [x] TopBar, don't use concatenation
- [ ] #2083 merged to master
- [x] Topbar CTA button history is broken in staging
- [x] Testing on staging
- [x] Babel polyfill issue: Is this branch somehow provoking it? How to fix?
- [x] Remove extra params from styleguide
- [x] Back button is not working correctly

# Client-side routes

_This document description how to use routes in the client-side JavaScript code._

## Usage

In the top-level React component, you need to import `subset` function from `utils/routes` module and call that function with the named routes you want to use. After that, you need to pass the newly created route subset object to the children, if they need routes. Note that you can pass default options to the `subset` function.

```js
// App.js
import { subset } from '../utils/routes';

export default (props, railsContext) => {
  const routes = subset([
    'homepage',
  ], { locale: "en" })

  return r(MyComponent, { routes });
};
```

In the component, you then use the routes subset to create the routes you want:

// MyComponent.js

class MyComponent extends React.Component {
  render() {
    return a(href: this.props.root.homepage_path())
  }
}

## Parameters

You can give parameters to the route helper function as you would do in Rails:

```js
Routes.person_path({username: "johndoe", show_closed: true})
```

## Cleaning up

`assets:clobber` deletes all the compiled route bundles.

```bash
rake assets:clobber
```

## Implementation details

The route bundling is powered by [js-routes](https://github.com/railsware/js-routes) gem. The gem provides two important utilities:

* `rake js:routes` task to export the routes to `.js` bundle
* The JavaScript `_path` helpers

In addition to that, we have implemented a middleware `JsRoutes::Middleware` that watches changes in the `config/routes.rb` file and compiles a new `.js` route bundle when ever that file changes. By default, this happens only in development mode.

### Deployment to production

`rake js:routes` task is configured so that it's always called before `rake assets:precompile`. So no extra steps are needed during the deployment.


Procfile.spec
```diff
@@ -3,7 +3,7 @@
 # in rails_helper.rb.
 
 # Build client assets, watching for changes.
-rails-client-assets: cd client && npm run build:dev:client
+rails-client-assets: script/wait_for_routes_js.sh && cd client && npm run build:dev:client
```
rap1ds: Should be `export_routes`

client/app/components/OnboardingGuide/OnboardingGuide.js
```diff
@@ -10,29 +10,28 @@ import GuidePaypalPage from './GuidePaypalPage';
...
 const selectChild = function selectChild(data, nextStep) {
   const { path, onboarding_data } = data;
-  const pageData = (path.length > 0) ?
-    _.find(onboarding_data, (pd) => pd.sub_path === path.substring(1)) :
-    {};
+  const pageData = _.find(onboarding_data, (pd) => pd.sub_path === path) || {};
 
   switch (path) {
-    case '/slogan_and_description':
+    case 'slogan_and_description':
```
rap1ds: Add the `/` back

lib/jsroutes_middleware.rb
```diff
@@ -0,0 +1,119 @@
...
+    #
+    # # Cache file doesn't exist
+    # # Route files and cache size are different (files were removed/added)
+    # # Routes file has been updated
+    #
+    def verify_route_files!
+      valid_cache = []
+      new_cache = {}
+
+      valid_cache.push cache_path.exist?
```
ovan: I'm guessing this is partly copy&paste code but our style has been to always use the parenthesis in method calls.

client/app/components/OnboardingGuide/GuideCoverPhotoPage.js
```diff
@@ -46,17 +47,13 @@ const GuideCoverPhotoPage = (props) => {
       ]),
     ]),
 
-    a({ className: css.nextButton, href: pageData.cta }, t('web.admin.onboarding.guide.cover_photo.add_your_own')),
+    a({ className: css.nextButton, href: Routes.admin_look_and_feel_edit_path() }, t('web.admin.onboarding.guide.cover_photo.add_your_own')),
```
ovan: This is debatable whether **presentation** level components should be aware of Rails routes or not. The tradeoff is the extra work of having to pass down the routes as props from top level vs putting the Rails dependency deep into the component tree. What do you think @Gnito, what should be the guideline here?

ovan: To clarify further, there's never a need for routes to go into redux store. They are static from the js app perspective so there's no need to trigger refresh as a result of routes changing.

client/app/components/OnboardingGuide/OnboardingGuide.js
```diff
@@ -9,87 +9,58 @@ import GuideFilterPage from './GuideFilterPage';
...
+  const canUseDOM = !!(typeof window !== 'undefined' &&
+                        window.document &&
+                        window.document.createElement);
+  const canUsePushState = !!(typeof history !== 'undefined' &&
+                              history.pushState);
+
+  if (canUseDOM && canUsePushState) {
+    window.history.pushState(state, title, path);
+  }
+}
```
ovan: Not in the scope of this PR but history management should be wrapped and this stuff should be handled generically.

rap1ds: Definitely. I kind of started doing something like that but then reverted, because I got a bit stuck + got the feeling that this PR is large enough already.

client/app/utils/initialize-env.js
```diff
@@ -0,0 +1,12 @@
+import { initialize as initializeI18n } from './i18n';
+import { initialize as initializeRoutes } from './routes';
+
+// This function is a single point where application initialization
+// should happen. Initialization can be for example setting the app
+// in correct state based on railsContext or node environment
+function initializeEnvironment(railsContext, nodeEnv) {
+  initializeI18n(railsContext, nodeEnv);
+  initializeRoutes(railsContext);
+}
```
ovan: How about instead of passing in the whole railsContext to i18n/routes initializers destructure here and pass in only what's needed. Feels like unnecessary coupling to railsContext now.

client/app/utils/routes.js
```diff
@@ -0,0 +1,98 @@
...
+};
+
+//
+// Public API
+//
+
+/**
+ * Creates a subset of all routes.
+ *
+ * You can pass also `defaultOpts` object, for example for "locale"
```
ovan: I find the name defaultOpts a bit misleading. If I get this right these are global arguments that are passed to all reverse routing functions so you don't have to pass them separately in every call? If so then I suggest renaming to globalParams.

docs/js-routes.md
```diff
@@ -0,0 +1,60 @@
...
+class MyComponent extends React.Component {
+  render() {
+    return a(href: this.props.root.homepage_path())
+  }
+}
+
+
+## Parameters
+
+You can give parameters to the route helper function as you would do in Rails:
```
ovan: Minor thing about naming, parameters vs arguments. Parameters are the things you give names in function signature and arguments are the values you pass in when you call a function. So this section is about arguments, not about parameters.

ref: http://stackoverflow.com/questions/1788923/parameter-vs-argument

client/app/utils/routes.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+};
+
+const didYouMean = function didYouMean(routeName) {
+  if (_.endsWith(routeName, '_path') && Routes[routeName]) {
+    return `Did you mean '${routeName.replace(/_path$/, '')}'?`;
+  } else {
+    return null;
+  }
+};
```
ovan: nice touch! 👍

PR #2068, opened 2016-05-24
Extend railsContext with marketplaceId

rap1ds:
This Pull Request extends the `railsContext` with `marketplaceId`. It also changes the cache helper so that the extended `railsContext` will invalidate the cache, if needed. In addition, a new module `marketplaceId` is created to encapsulate the id. The module is globally available, so the id is globally available to all components. It's not passed via `props`.

Todo:

- [X] Extend `railsContext` with `marketplaceId`
- [X] Invalidate cache if `railsContext` changes
- [X] Implement `marketplaceId` module.

Review comments wanted:

- `marketplaceId` is now globally available. There's no need to pass it via `props`. Is that ok? It's admittedly less clean: components aren't so "pure" anymore. On the otherhand, it's more convenient: no need to pass it via `props` through the component hierarchy.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -471,6 +471,10 @@ class ApplicationController < ActionController::Base
     @minutes_to_maintenance = NextMaintenance.minutes_to(now)
   end
 
+  def current_community_id
+    Maybe(@current_community).id.or_else(nil)
+  end
+
   private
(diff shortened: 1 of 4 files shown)


rap1ds:
Review comments: Remove the "Make marketplaceId globally available" commit

PR #2069, opened 2016-05-24
Topbar search

kpuputti:
Search component with combined keyword and location search inputs for the new top bar.

## Design

screen shot 2016-05-27 at 16 10 32

## Interactive component in the Styleguide

[Top bar search](https://sharetribe.github.io/sharetribe/topbar-search/?selectedKind=Top%20bar&selectedStory=With%20keyword%20and%20location%20search)

TODO:

 - [x]  new component in Styleguide
 - [x] focus/hover styles and dimensions
 - [x] font styles and spacings
 - [ ] ~~submit action~~ (another PR)
 - [ ] ~~location autocompletion~~ (another PR)
 - [ ] ~~mobile layout~~ (another PR)


client/app/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,76 @@
...
+  constructor(props) {
+    super(props);
+
+    // The values of the inputs are synced to the component state
+    // until the form is submitted.
+    this.state = {
+      keyword: '',
+      location: '',
+    };
+  }
```
Gnito: I guess, this is good solution now, but I think that, in the future, we need to add location and keyword to redux store. (I.e. the future map component probably wants to listen location info too.)

client/app/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,76 @@
...
+    });
+
+    const hasKeywordInput = mode === SEARCH_MODE_KEYWORD || mode === SEARCH_MODE_KEYWORD_AND_LOCATION;
+    const hasLocationInput = mode === SEARCH_MODE_LOCATION || mode === SEARCH_MODE_KEYWORD_AND_LOCATION;
+
+    return form({
+      classSet: {
+        [css.root]: true,
+      },
+      onSubmit: this.handleSubmit.bind(this),
```
Gnito: I've usually kept a habit of writing 'this' bindings in constructor. It's a good place to do bindings since functions can be used in several places - and having bindings all over the code seems a bit noisy/unclear for me.

Gnito:
Have you tested this on different browsers?

client/app/components/SearchBar/SearchBar.js
kpuputti: @Gnito The state keeps track of the real time value, and there's a callback to get a submitted/selected value to the parent component (that can be put to a store etc.). Do you think it makes sense to sync the "incomplete" value to the store?

kpuputti:
@Gnito Chrome, Firefox, and Safari. I need to get an IE somewhere.

client/app/components/SearchBar/SearchBar.js
kpuputti: I just wanted to avoid having to repeat the function name in three places for a single use.

I'd like to avoid using `this` altogether and not having to bind at all. Would inlining the submit function make sense?

PR #2072, opened 2016-05-25
Clean details and look 'n feel routes

rap1ds:
- [X] Clean "edit details" routes
- [X] Redirect old "edit details" routes
- [X] Clean "look and feel" routes
- [X] Redirect old "look and feel" routes

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -146,10 +146,10 @@ Kassi::Application.routes.draw do
       get "getting_started_guide(/*all)" => "getting_started_guide#index", as: :getting_started_guide
 
       # Details and look 'n feel
-      get   "/look_and_feel/edit" => "communities#edit_look_and_feel"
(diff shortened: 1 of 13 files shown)


config/routes.rb
```diff
@@ -142,15 +142,18 @@
...
       patch "/settings" => "communities#update_settings", as: :update_settings
 
+      # Guide
       get "getting_started_guide(/*all)" => "getting_started_guide#index", as: :getting_started_guide
 
+      # Details and look 'n feel
+      get   "/look_and_feel/edit" => "communities#edit_look_and_feel"
+      patch "/look_and_feel"      => "communities#update_look_and_feel"
+      get   "/details/edit"       => "community_customizations#edit_details"
+      patch "/details"            => "community_customizations#update_details"
```
ovan: I think it would be better style to explicitly name the routes with :as to make it clear what the path/url helpers are called.

PR #2076, opened 2016-05-25
Component folders according to atomic design

Gnito:
We discussed about folder structure in Top Gun team and propose something like this:
Components are separated based on [Atomic design](http://bradfrost.com/blog/post/atomic-web-design/): _elements_ (aka atoms) , _composites_ (aka molecules), and _sections_ (aka sections).
- **Elements** are React components which are basic visual elements. For example: avatar image.
- **Composites** are combined elements. E.g. ProfileCard combining Avatar and Name elements could be a composite.
- **Sections** are higher level composites. They are responsible for page sections like html5 tags do. (Think about ```<header>```, ```<footer>```, ```<main>```, ```<aside>```, and ```<section>```)
screen shot 2016-06-07 at 13 53 33

Inspirational blog post: https://www.smashingmagazine.com/2016/05/creating-a-living-style-guide-case-study/#style-guide-architecture


ovan:
How does this relate to the Redux classification of components into presentational and container components http://redux.js.org/docs/basics/UsageWithReact.html? Is this about subdividing the presentational into these 3 groups? I think the react classification is very important because the that way the presentational part (which I assume to be the bulk of components) stays independent of Redux and maybe independent of Rails too. So basically presentational components can be re-used across the multiple "SPAs" we have but containers are likely to be per SPA.

About naming, I maybe don't have the strongest voice in this discussion but I've not a big fan of these borrowed / artifical in the context names. I liked the names they used in the other article but if there's a fear of too much overloading for component and module then how about:
1. **Elements** - Like in that blog post and also in the description you wrote above, these are ~ HTML elements
2. **Composites** - Dictionary says: *1. a thing made up of several parts or elements*
3. **Sections** - You keep explaining the third thing as "think about page sections..." so why not call them that directly?

Gnito:
Elements and sections have already some semantics attached to them within html & layouting craft, but I'm fine with using those.

ovan:
They do but aren't those semantics actually useful in this context helping people to understand what's the scope of a component in question ref: the descriptions given in the original comment using those words to explain the artificial terms?

ovan:
What about by other question, how does this relate to presentational vs container component thinking?

Gnito:
Atoms/elements and molecules/composites can appear in multiple places on the page. So, it would make sense to keep them purely presentational.

Organisms/sections have a bigger responsibility and they most likely appear only once in any page - (in my mind, two "headers" make sense only if their inner structure/layout is different). Page (/SPA) could also swap those on the fly (e.g. Page could change organism/section called GridView to MapView.)

So, in that sense, organisms/sections could be responsible for handling domain data and therefore work as containers.

client/app/routes/routes.js
```diff
@@ -1295,15 +1295,6 @@ Based on Rails routes of Kassi::Application
...
-// styleguide => /styleguide(.:format)
-  // function(options)
-  styleguide_path: Utils.route([], ["format"], [2,[7,"/",false],[2,[6,"styleguide",false],[1,[2,[8,".",false],[3,"format",false]],false]]], {}),
-// styleguide_example_server_side_redux_app => /styleguide/example_server_side_redux_app(.:format)
-  // function(options)
-  styleguide_example_server_side_redux_app_path: Utils.route([], ["format"], [2,[7,"/",false],[2,[6,"styleguide",false],[2,[7,"/",false],[2,[6,"example_server_side_redux_app",false],[1,[2,[8,".",false],[3,"format",false]],false]]]]], {}),
-// styleguide_getting_started_guide => /styleguide/getting_started_guide(/*all)(.:format)
-  // function(options)
-  styleguide_getting_started_guide_path: Utils.route([], ["all","format"], [2,[7,"/",false],[2,[6,"styleguide",false],[2,[7,"/",false],[2,[6,"getting_started_guide",false],[2,[1,[2,[7,"/",false],[5,[3,"*all",false],false]],false],[1,[2,[8,".",false],[3,"format",false]],false]]]]]], {}),
```
kpuputti: Is this an autogenerated file? And should it be committed whenever it changes?

rap1ds: No it shouldn't. Sorry, my bad. I added it to `.gitignore` but forgot to remove the file itself. Here's a PR that fixes the issue: https://github.com/sharetribe/sharetribe/pull/2094

PR #2077, opened 2016-05-26
Assets are copied to public/webpack folder

Gnito:
also in Rails + Webpack dev environment.

- small images are converted to data url (nothing changed)
- hotMode defined: (webpack server handles files - i.e. styleguide - we emit the file as before to webpack output directory.)
- asset_host defined: assets are copied to public/webpack and asset_host is used (lines are changed, behaviour stays the same)
- default (Rails is running and serving assets): assets are copied to public/webpack (previously they were copied to webpack output directory)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ client/node_modules
 !/client/app/i18n/.keep
 /client/app/routes/*
 !/client/app/routes/.keep
+
+# Generated assets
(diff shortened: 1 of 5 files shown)


Gnito:
You might need to refresh your local cache of npm modules. Execute something like:
cd ./client && npm uninstall customfile-loader && npm install && cd ..

PR #2078, opened 2016-05-27
Route for landing page

rap1ds:


rap1ds:
This PR, as it is here, will not be merged. I'm going to leave the branch, it has some commits that are useful in the future.

PR #2079, opened 2016-05-27
Distance filter and sort parameter to Search

mporkola:


app/services/listing_index_service/search/zappy_adapter.rb
```diff
@@ -74,15 +72,18 @@ def format_params(original)
...
        :'page[size]' => original[:per_page],
        :'filter[price_min]' => Maybe(original[:price_cents]).map{ |p| p.min }.or_else(nil),
        :'filter[price_max]' => Maybe(original[:price_cents]).map{ |p| p.max }.or_else(nil),
        :'filter[omit_closed]' => !original[:include_closed],
        :'filter[listing_shape_ids]' => Maybe(original[:listing_shape_ids]).join(",").or_else(nil),
        :'filter[category_ids]' => Maybe(original[:categories]).join(",").or_else(nil),
-       :'search[locale]' => original[:locale]
-      }.merge(search_params).merge(custom_fields).compact
+       :'search[locale]' => original[:locale],
+       :'filter[distance_max]' => original[:distance_max],
```
ovan: I'd say defining this logically belongs together with the location_params above.

app/controllers/homepage_controller.rb
```diff
@@ -171,6 +173,8 @@ def find_listings(params, listings_per_page, filter_params, includes, location_s
       page: Maybe(params)[:page].to_i.map { |n| n > 0 ? n : 1 }.or_else(1),
       price_min: params[:price_min],
       price_max: params[:price_max],
+      distance_max: params[:distance_max],
```
ovan: Where is the params[:distance_max] set now? What about setting the min to 5km?

Gnito:
This doesn't take into account that Google returns spot locations (no viewport) and 
we also have custom fetch for those cases where user just presses enter.
I'll continue with this PR, but getting this ready will take a bit more time.

app/controllers/homepage_controller.rb
```diff
@@ -5,6 +5,7 @@ class HomepageController < ApplicationController
 
   APP_DEFAULT_VIEW_TYPE = "grid"
   VIEW_TYPES = ["grid", "list", "map"]
+  APP_MINIMUM_DISTANCE_MAX = 5
```
Gnito: Was this something that was agreed earlier - or just a placeholder value?

app/assets/javascripts/location_search.js
```diff
@@ -2,26 +2,60 @@ window.ST = window.ST || {};
...
     var boundingboxInput = document.getElementById('boundingbox');
+    var maxDistanceInput = document.getElementById('distance_max');
     var homepageForm = document.getElementById('homepage-filters');
     var autocomplete = new window.google.maps.places.Autocomplete(searchInput, { bounds: { north: -90, east: -180, south: 90, west: 180 } });
     autocomplete.setTypes(['geocode']);
 
     boundingboxInput.value = null;
 
+    function toRadians(degrees) {
+      return degrees * (3.1415/180);
```
kpuputti: You can use `Math.PI` instead of the literal value.

app/assets/javascripts/location_search.js
```diff
@@ -2,26 +2,60 @@ window.ST = window.ST || {};
...
+      var R = 6371; // Earth's radius in km
+
+      var lat1 = a.lat();
+      var lat2 = b.lat();
+      var lng1 = a.lng();
+      var lng2 = b.lng();
+      var φ1 = toRadians(lat1);
+      var φ2 = toRadians(lat2);
+      var Δφ = toRadians(lat2-lat1);
+      var Δλ = toRadians(lng2-lng1);
```
kpuputti: These variable names are kinda cool, but are you sure we want to do this? Using those vars pretty much requires copy pasting them around since at least I have no idea how to input them with a keyboard.

app/assets/javascripts/location_search.js
```diff
@@ -2,26 +2,60 @@ window.ST = window.ST || {};
...
+
+      var a = Math.sin(Δφ/2) * Math.sin(Δφ/2) +
+              Math.cos(φ1) * Math.cos(φ2) *
+              Math.sin(Δλ/2) * Math.sin(Δλ/2);
+      var c = 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1-a));
```
kpuputti: If the above (and below) code is some known algorithm, a link would help digesting this.

app/assets/javascripts/location_search.js
Gnito: @kpuputti yeah, I was thinking the same. (@mporkola I'll change these)

app/assets/javascripts/location_search.js
Gnito: True, I'll change this

app/assets/javascripts/location_search.js
Gnito: Probably, I think I've seen this somewhere related to google maps.

app/controllers/homepage_controller.rb
```diff
@@ -156,30 +155,30 @@ def find_listings(params, listings_per_page, filter_params, includes, location_s
     dropdowns = filter_params[:custom_dropdown_field_options].map { |dropdown_field| dropdown_field.merge(type: :selection_group, operator: :or) }
     numbers = numeric_search_params.map { |numeric| numeric.merge(type: :numeric_range) }
 
-    coordinates = Maybe(params[:lc]).map { search_coordinates(params[:lc]) }.or_else({})
     distance_unit = (location_search_in_use && MarketplaceService::API::Api.configurations.get(community_id: @current_community.id).data[:distance_unit] == :metric) ? :km : :miles
-
-    distance_max = [5, params[:distance_max]].max if params[:distance_max].present?
+    location_search_params = location_search_params(
```
ovan: How would you feel about renaming this locale variable. My worry is that it is easy to confuse the variable name with the method name now.

app/controllers/homepage_controller.rb
ovan: 5km was agreed as a good minimum filter range.

app/assets/javascripts/location_search.js
Gnito: Haversine formula:
http://www.movable-type.co.uk/scripts/latlong.html

app/assets/javascripts/location_search.js
```diff
@@ -2,26 +2,63 @@ window.ST = window.ST || {};
...
+      return d/2;
+    };
+
+    function updateViewportData(viewport) {
+      if (viewport) {
+        var boundingboxRadius = computeScale(viewport.getNorthEast(), viewport.getSouthWest());
+        maxDistanceInput.value = boundingboxRadius;
+        boundingboxInput.value = viewport.toUrlValue();
+      } else {
+        maxDistanceInput.value = minimumDistanceMax;
```
mporkola: Minimum distance of 5 km is now enforced in two places, here and here: https://github.com/sharetribe/sharetribe/pull/2079/files#diff-8b4aa234db6cb84dd4803f9b2c78ce5fR165

I think `HomepageController` is where it belongs to, this should return `null` or `undefined` if there's no viewport.

app/controllers/homepage_controller.rb
```diff
@@ -154,26 +157,30 @@ def find_listings(params, listings_per_page, filter_params, includes, location_s
...
       keywords: filter_params[:search],
-      latitude: coordinates[:latitude],
-      longitude: coordinates[:longitude],
-      distance_unit: distance_unit,
       fields: checkboxes.concat(dropdowns).concat(numbers),
       per_page: listings_per_page,
       page: Maybe(params)[:page].to_i.map { |n| n > 0 ? n : 1 }.or_else(1),
       price_min: params[:price_min],
       price_max: params[:price_max],
+      sort: nil,
```
mporkola: I think it would be more clear if the `sort` param was decided here, instead of in `location_search_params`

PR #2088, opened 2016-06-06
Add Georgian (ka) locale

kpuputti:
https://en.wikipedia.org/wiki/Georgian_language

Also fix the Czech Republic region code.


thomasmalbaux:
Should the region be `KA` instead of `GE` (as done with Hungarian or Czech) (as the ident id `ka` and not `ka-GE`, which would be a more specific locale)?

kpuputti:
@thomasmalbaux The instruction comment in the file said the region format is from [ISO 3166](https://en.wikipedia.org/wiki/ISO_3166-1), and the code for Georgia is GE there, so I'm not sure what should be done here.

rap1ds:
LGTM

PR #2089, opened 2016-06-06
Landing page controller

rap1ds:
Base controller for landing page. It contains the Hero section, some sample data and the data denormalizer.

The controller is unaccessible at the moment, but if you want to test it, go to routes.rb and uncomment the landing page routes.

This PR doesn't contain the assets needed for testing the landing page (`public/landing_page` folder missing). Ping me if you need these.

### DIFF
--- a/app/controllers/landing_page_controller.rb
+++ b/app/controllers/landing_page_controller.rb
@@ -2,9 +2,9 @@ class LandingPageController < ActionController::Metal
 
   class Denormalizer
 
-    def initialize(root: "composition", hooks: {})
+    def initialize(root: "composition", link_resolvers: {})
       @root = root
-      @hooks = hooks
+      @link_resolvers = link_resolvers
     end
 
     def to_tree(normalized_data)
@@ -24,7 +24,7 @@ class LandingPageController < ActionController::Metal
               raise ArgumentError.new("Invalid link: #{v.inspect} has a 'type' key but no 'id'")
             else
               # Is a link
-              hook(type, normalized_data[type][id])
+              resolve_link(type, id, normalized_data)
             end
 
           [k, new_v]
@@ -71,11 +71,11 @@ class LandingPageController < ActionController::Metal
 
     private
 
-    def hook(type, value)
-      if @hooks[type].respond_to? :call
-        @hooks[type].call(value)
+    def resolve_link(type, id, normalized_data)
+      if @link_resolvers[type].respond_to? :call
+        @link_resolvers[type].call(type, id, normalized_data)
       else
-        value
+        normalized_data[type][id]
       end
     end
   end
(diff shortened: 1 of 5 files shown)


app/utils/hash_utils.rb
```diff
@@ -144,4 +144,39 @@ def flatten(h)
...
+  #
+  # Example (stringify keys):
+  #
+  # deep_map(a: 1, b: 2) { |k, v|
+  #   [k.to_s, v]
+  # }
+  #
+  # Unlike Ruby's Hash#map, this method returns a Hash, not an Array.
+  #
+  def deep_map(obj, &block)
```
ovan: I'm again thinking about naming and what's the general case here. This is similar to tree walking except that in a typical tree walk you can process the non-leaf nodes too if you want to. This impl would be something like `prewalk-leaves`, which really is just a special case of `prewalk`. See e.g. https://clojure.github.io/clojure/clojure.walk-api.html

ovan: Actually I read this wrong. This does allow you to process all nodes.

rap1ds: I don't know what is the actual definition of "walking", but to me it sounds more generic than this operation, which is "mapping". To me "walking" sounds like iterating all the nodes, but doesn't specify what the operation is (e.g. replacing all the `.map` calls in this function with `.each`). But again, I'm not sure what's the correct definition. Do you know?

app/controllers/landing_page_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+    end
+
+    def to_tree(normalized_data)
+      root = normalized_data[@root]
+
+      HashUtils.deep_map(root) { |k, v|
+        case v
+        when Hash
+          type = v["type"]
+          id = v["id"]
```
ovan: `type, id = v.values_at("type", "id")`

app/controllers/landing_page_controller.rb
```diff
@@ -0,0 +1,110 @@
+class LandingPageController < ActionController::Metal
+
+  class Denormalizer
+
+    def initialize(root: "composition", hooks: {})
+      @root = root
+      @hooks = hooks
```
ovan: So the hooks are post-processors for linked data?

app/controllers/landing_page_controller.rb
rap1ds: TIL

app/controllers/landing_page_controller.rb
rap1ds: Yeah, that's the only thing they do at the moment. To be honest I didn't think too much about hooks, but just needed some way to add the proper path to the images.

PR #2091, opened 2016-06-07
Removed TRY as currency for new marketplaces in Turkey

thomasmalbaux:
As PayPal is no longer available in Turkey. 
Previously, payments in TRY were only available for in-country payments.

I have not tested this and I'm actually not even sure if more changes are needed elsewhere...


thomasmalbaux:
Based on Mikko's recommendations, minimum_commission has been added back as it doesn't hurt and avoid potential troubles.

rap1ds:
I'll test this in staging now.

PR #2092, opened 2016-06-07
Show branding info in transaction emails

sktoiva:
Some of the transactional emails such as payment receipts are set up differently from the rest. This causes `@show_branding_info` to be nil and inconsistency in the emails.

This pull request sets up the `@show_branding_info` properly for those emails.


rap1ds:
Code looks good. Didn't test it so remember to test.

PR #2095, opened 2016-06-08
Marketplace resolver

rap1ds:
This PR extracts the logic for finding current marketplace in a new module, `MarketplaceResolver`. In the future, if we decide to do dynamic routing, which requires us to fetch the marketplace in the Rack middleware, this module can be used there.\

Needs testing:

- [x] Braintree
- [ ] Mercury editor
- [x] Errors controller


sktoiva:
Looks good, remember to test.

PR #2098, opened 2016-06-09
Write status to plans

sktoiva:


spec/services/plan_service/api/plans_spec.rb
```diff
@@ -118,6 +125,17 @@
             community_id: 123,
             plan: {
               plan_level: PlanService::Levels::PRO,
+              status: :active,
+              member_limit: 1000,
+            }) }.to raise_error(ArgumentError)
+        end
+
+        it "raises error if features is missing"do
```
rap1ds: raises error if status is missing?

sktoiva: Ah yeah. Copypaste.

PR #2100, opened 2016-06-09
Use status

sktoiva:


app/views/admin/_uservoice_widget.html.erb
```diff
@@ -34,7 +34,7 @@ UserVoice.push(['identify', {
    id:           '<%= @current_community.id %>',
    name:         '<%= @current_community.full_url %>',
    created_at:   '<%= @current_community.created_at.to_time.to_i %>',
-   plan:         '<%= @current_plan[:plan_level] %>'
+   plan:         '<%= @current_plan[:status] %>'
```
rap1ds: This is now a bit different information, but I don't think it matters.

sktoiva: Yeah, we discussed this already.

PR #2101, opened 2016-06-09
Remove plan level usage

sktoiva:


app/services/plan_service/api/plans.rb
```diff
@@ -37,7 +37,7 @@ def create(community_id:, plan:)
     # All plans should come from the external plan service and that's
     # why this function is deprecated
     #
-    def create_initial_trial(community_id:, plan:)
+    def create_initial_trial(community_id:, plan: {})
```
rap1ds: Hmh... this seems to be deprecated function. Is this used anywhere? I'm thinking, can we just remove it?

rap1ds: Ahaa, apparently some tests are still using this.

rap1ds:
The code looks good to me, didn't test it though. Remember to test.

app/services/plan_service/api/plans.rb
sktoiva: And int_api/marketplaces_controller also uses. This is deprecated because we should get all trials from charm, but currently we don't.

PR #2102, opened 2016-06-09
Remove plan level from db

sktoiva:


db/migrate/20160609081158_remove_plan_level_from_plans.rb
```diff
@@ -0,0 +1,9 @@
+class RemovePlanLevelFromPlans < ActiveRecord::Migration
+  def up
+    remove_column :marketplace_plans, :plan_level
+  end
+
+  def down
+    add_column :marketplace_plans, :plan_level, :integer, after: :community_id
```
rap1ds: The `up` migration removes now data and the column, but the `down` migration only brings back to column but not data. We could write also down migration that brings back the data (e.g. read the plan-name and based on that populate the plan_level).

But I don't know if that's necessary...

rap1ds: Whoops, sorry, actually my comment was meant to the external plan service repo.

PR #2103, opened 2016-06-09
Change landingpage data structure

rap1ds:
- Move ids from hash keys to hash values

### DIFF
--- a/app/controllers/landing_page_controller.rb
+++ b/app/controllers/landing_page_controller.rb
@@ -69,6 +69,10 @@ class LandingPageController < ActionController::Metal
       end
     end
 
+    def self.find_link(type, id, normalized_data)
+      normalized_data[type].find { |item| item["id"] == id }
+    end
+
     private
 
     def resolve_link(type, id, normalized_data)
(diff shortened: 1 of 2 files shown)


app/controllers/landing_page_controller.rb
```diff
@@ -117,17 +122,24 @@ def data
...
           "background_image" => {"type" => "assets", "id" => "myheroimage"},
           "search_placeholder" => "What kind of turbojopo are you looking for?",
           "search_button" => "Search",
         },
-        "thecategories" => {"type" => "categories", "slogan" => "blaablaa", "category_ids" => [123, 432, 131]},
-      },
+
+        {
+          "id" => "thecategories",
+          "type" => "categories",
```
ovan: s/type/kind ?

app/controllers/landing_page_controller.rb
```diff
@@ -75,9 +75,13 @@ def resolve_link(type, id, normalized_data)
...
         @link_resolvers[type].call(type, id, normalized_data)
       else
-        normalized_data[type][id]
+        self.class.find_link(type, id, normalized_data)
       end
     end
+
+    def self.find_link(type, id, normalized_data)
+      normalized_data[type].find { |item| item["id"] == id }
+    end
```
ovan: This is ok for now but let's see when the structure gets bigger if it needs a by-id index instead of linear search.

PR #2104, opened 2016-06-09
Add datamodel for storing landingpages

ovan:


app/controllers/landing_page_controller.rb
```diff
@@ -93,12 +96,60 @@ def resolve_link(type, id, normalized_data)
...
+      render_landing_page(structure)
+    rescue LandingpageContentNotFound
+      render_not_found()
+    end
   end
 
   private
 
-  def landing_page
+  def parse_int(int_str_or_nil)
```
sktoiva: If I get you correctly, you are aiming to parse string to integer with the exception that if the string is `nil` you want to return nil?

This can be achieved with `str&.to_i`. Now `Integer("")` throws an exception and you are forced to rescue it and return nil.

app/controllers/landing_page_controller.rb
```diff
@@ -93,12 +96,60 @@ def resolve_link(type, id, normalized_data)
...
+    Integer(int_str_or_nil || "")
```
rap1ds:
```
2.3.1 :007 > Integer(nil)
TypeError: can't convert nil into Integer
	from (irb):7:in `Integer'
	from (irb):7
	from /Users/mikko/.rvm/rubies/ruby-2.3.1/bin/irb:11:in `<main>'
```

db/migrate/20160609080700_add_clp_datamodel.rb
```diff
@@ -0,0 +1,19 @@
+class AddClpDatamodel < ActiveRecord::Migration
+  def up
+    create_table :landingpages do |t|
+      t.column :community_id, :integer, null: false
+      t.column :version, :integer, null: false
+      t.column :released, :datetime
+      t.column :content, :text, limit: 16.megabytes - 1, null: false
```
sktoiva: Do we want the default `timestamps` for landingpages?

db/migrate/20160609080700_add_clp_datamodel.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
```
sktoiva: Indices are missing.

db/migrate/20160609080700_add_clp_datamodel.rb
rap1ds: To my understanding, `16.megabytes - 1` is the default, so it's not needed here. If you want to keep it here for the sake of explicity, then 👍

db/migrate/20160609080700_add_clp_datamodel.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    add_column :marketplace_configurations, :clp_enabled, :boolean, after: :distance_unit, null: false, default: false
+    add_column :marketplace_configurations, :clp_version, :integer, after: :clp_enabled
```
rap1ds: Can be null?

rap1ds: Answering to myself: Of course, if the landing page is not in use.

app/controllers/landing_page_controller.rb
sktoiva: Yeah, see my previous comment.

app/controllers/landing_page_controller.rb
rap1ds: `to_i` returns 0, for many cases, e.g. `"not an integer".to_i`. I would suggest regexp here.

rap1ds: ...or start the versioning from version 1.

app/controllers/landing_page_controller.rb
ovan: There is a rescue block and nil is not passed in because of `|| ""`, or what am I missing?.

db/migrate/20160609080700_add_clp_datamodel.rb
ovan: Explicit was the goal.

ovan: I'm not adding default timestamps because this data is never written via AR. It's a good point still that should I add a last_updated field and manually update that too. Hmm...

app/controllers/landing_page_controller.rb
sktoiva: Ah yeah, I checked only the `nil` case.

db/migrate/20160609080700_add_clp_datamodel.rb
```diff
@@ -0,0 +1,28 @@
...
+      t.column :version, :integer, null: false
+      t.column :released, :datetime
+      t.column :content, :text, limit: 16.megabytes - 1, null: false
+      t.timestamps
+    end
+    add_index :landing_page_versions, [:community_id, :version], unique: true
+  end
+
+  def down
+    remove_index :landing_pages, :community_id
```
rap1ds: This is probably unnecessary since the table is dropped anyway

PR #2107, opened 2016-06-10
Fix input styles leaking to global scope

kpuputti:
CSS modules change class selectors to something unique, but element selectors stay intact and in the global scope (if not nested within another selector).

The input styles from PR #2069 leaked to the global scope, breaking search filters. This fixes the problem by scoping the selector in the proper place.

Tested in staging.


mporkola:
looks 👍

PR #2110, opened 2016-06-10
Change automatic emails name to "Newsletters"

Luis-RG:
To make user settings and listing options more consistent with the
admin panel settings.


mporkola:
Otherwise good, but a test (link below) expects to find a text "Send me an update email daily"

https://github.com/sharetribe/sharetribe/blob/master/features/settings/user_changes_notification_settings.feature

PR #2112, opened 2016-06-13
SearchBar location autocompletion

kpuputti:
Add Google Maps location autocompletion to the `SearchBar` React component. The selected location comes as coordinates in the `onSubmit` callback value.

## Example 1: Selecting a location and submitting form

After selecting location:

screen shot 2016-06-13 at 15 18 26

`onSubmit` yields

```javascript
{
  keyword: null,
  location: "Hellevoetsluis, Netherlands",
  coordinates: "51.831863,4.13181"
}
```

## Example 2: Submitting form without selecting location

screen shot 2016-06-13 at 15 15 42

`onSubmit` yields

```javascript
{
  keyword: null,
  location: "hell",
  coordinates: "60.167215,24.965187" // first place predicted from the "hell" query
}
```

## Example 3: keyword only

screen shot 2016-06-13 at 15 20 13

`onSubmit` yields

```javascript
{
  keyword: "bicycle",
  location: "",
  coords: null
}
```

The `onSubmit` is **not** called if all the inputs are empty.

## Styleguide

See the [interactive component](https://sharetribe.github.io/sharetribe/topbar-location-autocomplete/?selectedKind=Top%20bar&selectedStory=With%20location%20search) in the Styleguide.

### DIFF
--- a/client/.storybook/head.html
+++ b/client/.storybook/head.html
@@ -1,1 +1,2 @@
 <script>window.I18n = {};</script>
+<script src="https://maps.googleapis.com/maps/api/js?libraries=places"></script>
--- a/client/app/components/composites/SearchBar/SearchBar.css
+++ b/client/app/components/composites/SearchBar/SearchBar.css
@@ -1,3 +1,29 @@
+/* The Google Maps Places Autocomplete dropdown box is added in the
+   end of the document body, and we must scope the global style
+   overrides within a class generated from the SearchBar component.
+
+   Classes available for customization can be seen at:
+
+   https://developers.google.com/maps/documentation/javascript/places-autocomplete#style_autocomplete
+*/
+.topLevelBody {
+  /* stylelint-disable selector-class-pattern */
+  & :global(.pac-container[style]) {
+    /* Generated autocompletion has inline styles, thus the ugly override.
+       See: https://css-tricks.com/override-inline-styles-with-css/ */
+    width: 20em !important;
+    margin-top: 10px;
+  }
+
+  & :global(.pac-item) {
+    line-height: 40px;
+  }
+  & :global(.pac-icon) {
+    margin-top: 10px;
+  }
+  /* stylelint-enable selector-class-pattern */
+}
+
 .root {
   display: inline-flex;
   width: 28.3em;
--- a/client/app/components/composites/SearchBar/SearchBar.js
+++ b/client/app/components/composites/SearchBar/SearchBar.js
@@ -1,3 +1,5 @@
+/* eslint-disable react/no-set-state */
+
 import { Component, PropTypes } from 'react';
 import { form, input, button, span } from 'r-dom';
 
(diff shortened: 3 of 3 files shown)


client/app/components/SearchBar/SearchBar.js
```diff
@@ -31,14 +156,24 @@ class SearchBar extends Component {
...
+      ref: (c) => {
+        this.keywordInput = c;
+      },
     });
     const locationInput = input({
       type: 'search',
       className: css.locationInput,
       placeholder: locationPlaceholder,
       autoComplete: 'off',
-      onChange: (e) => this.setState({ location: e.target.value }), // eslint-disable-line react/no-set-state
```
kpuputti: The Google Map Autocomplete sets the value of the input, which doesn't trigger `onChange` and results in the state not being in sync with what the user has selected. Therefore we check the input value in the `handleSubmit` method when the value is actually needed.

client/app/components/SearchBar/SearchBar.js
```diff
@@ -13,16 +15,139 @@ const SEARCH_MODES = [
...
+      // Only keyword input, submitting current input value
+      onSubmit({
+        keyword: keywordValueStr,
+        location: null,
+        coordinates: null,
+      });
+      return;
+    }
+
+    const keywordValue = this.keywordInput ? this.keywordInput.value : null;
```
kpuputti: Notice the clear distinction of the empty String vs. `null` value in the `onSubmit` callback. Empty value means that the input is in use but the value was empty, `null` means that the input is not in use. This way the returned object shape is always the same, but the caller can still differentiate which inputs provided some actual values.

client/app/components/SearchBar/SearchBar.js
```diff
@@ -13,16 +15,139 @@ const SEARCH_MODES = [
...
+
+    this.state = { selectedPlace: null };
+  }
+  componentDidMount() {
+    document.body.classList.add(css.topLevelBody);
+    if (!this.locationInput) {
+      return;
+    }
+    const bounds = { north: -90, east: -180, south: 90, west: 180 };
+    const autocomplete = new window.google.maps.places.Autocomplete(this.locationInput, { bounds });
```
kpuputti: I think this will break server side rendering?

client/app/components/SearchBar/SearchBar.js
```diff
@@ -13,16 +15,139 @@ const SEARCH_MODES = [
...
+  if (place && place.geometry) {
+    return place.geometry.location.toUrlValue();
+  }
+  return null;
+};
+
+const getDetails = (placeId) => new Promise((resolve, reject) => {
+  const serviceStatus = window.google.maps.places.PlacesServiceStatus;
+  const div = document.createElement('div');
+  const service = new window.google.maps.places.PlacesService(div);
```
mporkola: what happens here?

client/app/components/SearchBar/SearchBar.js
mporkola: Probably, unless we have initialised Google Maps also on server side. Hmm, this probably needs some research.

client/app/components/SearchBar/SearchBar.js
kpuputti: @mporkola [Weird API](https://developers.google.com/maps/documentation/javascript/reference#PlacesService)

The `div` element isn't used anywhere, but API just requires one.

client/app/components/SearchBar/SearchBar.js
kpuputti: The SearchBar wouldn't do anything useful with the Autocomplete in the server side, I'll change this to skip initialising it when rendering in the backend.

kpuputti: I just verified that the `componentDidMount` isn't called when rendering in the server side, so no changes needed here.

PR #2113, opened 2016-06-13
Babel-polyfill somehow breaks build, use es6-object-assign instead

mporkola:
Needs further investigation, but at least deployment works now


Gnito:
Alternative fix: https://github.com/sharetribe/sharetribe/pull/2114

kpuputti:
Let's rather go with PR #2114 

We probably have to clean up the polyfills at some point, but then we have to know precisely which ones we really need.

PR #2114, opened 2016-06-13
Process 0.11.4 didn't work with our dependency tree

Gnito:
New version of Process dependency didn't work. 
Issue: https://github.com/defunctzombie/node-process/issues/60
Fix: https://github.com/defunctzombie/node-process/pull/61/files

This PR overwrites dependency to use previous version 0.11.3 - i.e. let's wait for fix to be merged.


rap1ds:
The fix is now merged and node-process 0.11.5 is released.

PR #2124, opened 2016-06-15
Landing page improvements

rap1ds:
- [X] Move hero and footer sections to partials
- [X] Include needed JS libraries and application JS code in the main template instead of including them in the partial (e.g. Google Places and other location search related JS code)
- [X] Change all linked data (path e.g.) to object format (e.g. `{id: "signup_path", value: "/signup"}` instead of plain string `/signup`}
- [X] Rename `marketplace_color` to more generic `marketplace_data`


rap1ds:
Woops. The target branch should be master. Closing.

PR #2126, opened 2016-06-15
Apply current marketplace to env in middleware

sktoiva:


ovan:
Looks good. Test carefully before pushing to master. Looks innocent but touches every request. :)

PR #2127, opened 2016-06-15
Feeding marketplace data to CLP rendering

ovan:


app/controllers/landing_page_controller.rb
```diff
@@ -94,23 +83,37 @@ def community_id(request)
...
+    primary_color = Community.where(id: cid)
+                    .pluck(:custom_color1)
+                    .first
+
+    name, slogan, description = CommunityCustomization
+                                .where(community_id: cid, locale: locale)
+                                .pluck(:name, :slogan, :description)
+                                .first
+
+    { "primary_color" => primary_color ? "#" + primary_color : nil,
```
rap1ds: What about using `.present?` there to make sure that empty string is threated as falsy

`primary_color.present? ? "#" + primary_color : nil,`

rap1ds:
Added one comment, otherwise good!

PR #2128, opened 2016-06-15
Increase the size of allowed_emails field

gnomet:
In practice this changes it in MySQL from TEXT to MEDIUMTEXT
Based on quick online research and local testing, this shouldn't
cause any noticeable change in performance.

E.g. [here](http://stackoverflow.com/questions/3516290/any-pitfalls-of-converting-mysql-text-field-to-mediumtext) the performance effect is discussed

I also did testing locally and load times and memory usage didn't show any significant difference (although lot of variation in load times on very page load so not too reliable. Anyway dramatic change probably would have been noticeable.)

Anyway, doesn't hurt to keep an eye on production memory usage after deploy. (I can do that.)

### DIFF
--- a/db/migrate/20160615145518_increase_max_size_of_allowed_emails.rb
+++ b/db/migrate/20160615145518_increase_max_size_of_allowed_emails.rb
@@ -1,5 +1,8 @@
 class IncreaseMaxSizeOfAllowedEmails < ActiveRecord::Migration
-  def change
+  def up
     change_column :communities, :allowed_emails, :text, limit: 16.megabytes - 1
+  end
+  def down
+    change_column :communities, :allowed_emails, :text
   end
 end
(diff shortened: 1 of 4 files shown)


db/migrate/20160615145518_increase_max_size_of_allowed_emails.rb
```diff
@@ -0,0 +1,5 @@
+class IncreaseMaxSizeOfAllowedEmails < ActiveRecord::Migration
+  def change
+    change_column :communities, :allowed_emails, :text, limit: 16.megabytes - 1
+  end
```
ovan: I wonder if just implementing change is enough in this case to support migrations to both directions. How would Activerecord Migrations know what's the previous type of the column based upon this? I believe you need separate up and down migrations here.

gnomet: Indeed, it raises "ActiveRecord::IrreversibleMigration" if tried to rollback. Good catch. I've been happily using change so many times that forgot to think this line doesn't include the earlier state.

PR #2132, opened 2016-06-16
Clp conditional routing

sktoiva:
Enables conditional routing to landing pages if it's enabled.

Might still break root redirection for marketplaces that have landing page enabled, but this can be taken to production and `redirect_to root` checked afterwards.

Changes to functionality:

  - Route / for marketplaces with CLP to custom landing page
  - Route /s for marketplaces with CLP to homepage#index
  - Route / for marketplaces without CLP to homepage#index
  - Redirect /s for marketplaces without CLP to homepage#index with 307, Temporary
  - Redirect to landing page for private marketplaces with CLP from homepage, if not logged in
  - Check protocol in middleware and perform redirect to https already there
  - Conditionally link to either / or /s from homepage#index views depending on if CLP is enabled
  - Replaces redirect roots with either landing page or search
  - Deprecates root helper
  - Replaces root_paths with search_path
  - Replaces root_urls with search_urls
  - Links to landing page instead of homepage in header icon


app/controllers/homepage_controller.rb
```diff
@@ -8,6 +8,10 @@ class HomepageController < ApplicationController
   APP_MINIMUM_DISTANCE_MAX = 5
 
   def index
+    redirect_to root and return if CustomLandingPage::LandingPageStore.enabled?(@current_community.id) &&
```
ovan: This set of conditions could be captured to a private method to give it a name.

lib/rack_middleware/enforce_ssl.rb
```diff
@@ -0,0 +1,21 @@
+class EnforceSsl
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    if ::APP_CONFIG.always_use_ssl.to_s == "true" && env["HTTPS"] == "off" # always_use_ssl can be string if it comes from ENV
+      req = ::Rack::Request.new(env)
+      port_s = req.port ? ":#{req.port}" : ""
+      redirect("https://#{req.host}#{port_s}#{req.fullpath}")
```
ovan: Does req.fullpath include e.g. query string? I'm trying to think through if this can lose some data.

sktoiva: Yes, it includes.

app/helpers/application_helper.rb
```diff
@@ -675,6 +675,23 @@ def sort_link_direction(column)
     params[:sort].eql?(column) && params[:direction].eql?("asc") ? "desc" : "asc"
   end
 
+  def search_path(opts = {})
```
ovan: Maybe this should have a different name so it wouldn't look like an autogenerated path helper fn.

lib/rack_middleware/enforce_ssl.rb
```diff
@@ -0,0 +1,21 @@
+class EnforceSsl
```
rap1ds: Did you consider using the default Rails `force_ssl`? http://guides.rubyonrails.org/configuring.html#rails-general-configuration Why did you end up writing own implementation?

rap1ds: Did you consider using the default Rails `force_ssl`? http://guides.rubyonrails.org/configuring.html#rails-general-configuration Why did you end up writing own implementation?

config/routes.rb
```diff
@@ -49,16 +49,32 @@
...
+  get '/s' => 'homepage#index', as: :search_without_locale, constraints: ->(request) {
+    CustomLandingPage::LandingPageStore.enabled?(request.env[:current_marketplace]&.id)
+  }
+
+  # Default routes for homepage, these are matched if custom landing page is not in use
   # Inside this constraits are the routes that are used when request has subdomain other than www
   get '/:locale/' => 'homepage#index', :constraints => { :locale => locale_matcher }, as: :homepage_with_locale
   get '/' => 'homepage#index', as: :homepage_without_locale
-  root :to => 'homepage#index'
+  get '/:locale/s', to: redirect('/%{locale}', status: 307)
```
rap1ds: Shouldn't we add the `locale_matcher` constraints also here?

app/services/custom_landing_page/marketplace_data_store.rb
```diff
@@ -1,7 +1,7 @@
 module CustomLandingPage
   module MarketplaceDataStore
 
-    DEFAULT_COLOR = "A64C5D";
+    DEFAULT_COLOR = "A64C5D"
```
rap1ds: Uups, my bad. Thanks for fixing!

app/helpers/application_helper.rb
```diff
@@ -675,6 +675,23 @@ def sort_link_direction(column)
...
+    o = opts.dup
+    o.delete("controller")
+    o.delete("action")
+
+    case [CustomLandingPage::LandingPageStore.enabled?(@current_community&.id), params[:locale].present?]
```
rap1ds: Why would the `@current_community` be `nil`?

config/routes.rb
```diff
@@ -49,16 +49,32 @@
...
+  get '/s', to: redirect('/')
```
rap1ds: Is the redirect status left out from here on purpose?

app/helpers/application_helper.rb
rap1ds: There is a bit similar logic in `application_controller#set_homepage_path` method. That method takes also `@current_user` into account. I.e. it doesn't add the locale, even if it's in the parameters, if `@current_user` exists, because the locale is already in the session. Does that make sense? Should we add it here also?

lib/rack_middleware/enforce_ssl.rb
sktoiva: Yes. I considered as stated in the `application.rb`. 

```
 # Manually redirect http to https, if config option always_use_ssl is set to true
 # This needs to be done before routing: conditional routes break if this is done later
 # Enabling HSTS and secure cookies is not a possiblity because of potential reuse of domains without HTTPS

sktoiva: And force_ssl adds HSTS and secure cookies.

ovan: Oops, I asked about this too and we ended up having an offline discussion here. Should have documented it here too.

app/helpers/application_helper.rb
sktoiva: Can't think of better name as it generates the search view path.

PR #2133, opened 2016-06-16
Topbar avatar & profile menu

mporkola:
See working example in [Styleguide](https://sharetribe.github.io/sharetribe/avatar/?selectedKind=Top%20bar&selectedStory=Basic%20state)


client/app/components/composites/Avatar/ArrowDropdown.css
```diff
@@ -0,0 +1,66 @@
...
+  border: 1px solid var(--topbarBorderColor);
+  right: 18px; /* we need a var here, whatever our generic margin is*/
+  display: flex;
+  flex-direction: column;
+  background-color: white;
+  /* Notifications_box: */
+  background: #FFFFFF;
+  box-shadow: 0px 2px 4px 0px rgba(0,0,0,0.10);
+  border-radius: 3px;
+  z-index: 10;
```
kpuputti: We might need to control all our z-indexes somewhere, and it will be a lot easier if they are all coming from variables.

client/app/components/composites/Avatar/ArrowDropdown.css
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.profileActions {
+  display: flex;
+}
+
+.profileAction {
+  width: 7.5rem;
+  height: 7.5rem;
```
kpuputti: Using rems might be a problem when adding the component to the existing application where we don't control the root font size.

client/app/components/composites/Avatar/ArrowDropdown.css
```diff
@@ -0,0 +1,66 @@
...
+  border-right: 1px solid var(--topbarBorderColor);
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  cursor: pointer;
+}
+
+.profileAction:hover {
```
kpuputti: No need to repeat the selector, cssnext supports nesting.

client/app/components/composites/Avatar/ArrowDropdown.css
```diff
@@ -0,0 +1,66 @@
...
+.profileAction:last-of-type {
+  border-right: 0;
+}
+
+.profileActionLabel {
+}
+
+.logoutArea {
+  border-top: 1px solid var(--topbarBorderColor);
+  padding: 0.6rem 1rem;
```
kpuputti: As above, rems might be a problem.

client/app/components/composites/Avatar/ArrowDropdown.css
Gnito: This comes from clash between old search bar below. (Two absolutely positioned containers in parallel). 'z-index: 1' should be enough.

client/app/components/composites/Avatar/ArrowDropdown.js
```diff
@@ -0,0 +1,63 @@
...
+      div({ className: css.profileActionLabel }, this.props.label),
+    ]);
+  }
+}
+
+ProfileActionCard.propTypes = {
+  icon: PropTypes.string,
+  label: PropTypes.string,
+  action: PropTypes.func,
+};
```
kpuputti: Are the prop types missing `.isRequired` on purpose?

Gnito:
@mporkola have you tested this mouseover solution with touch devices? (At least, tablet version is going to use profile dropdown.)

client/app/components/composites/Avatar/ArrowDropdown.js
```diff
@@ -0,0 +1,63 @@
...
+}
+
+ArrowDropdown.propTypes = {
+  actions: PropTypes.shape({
+    inboxAction: PropTypes.func,
+    profileAction: PropTypes.func,
+    settingsAction: PropTypes.func,
+    adminDashboardAction: PropTypes.func,
+    logoutAction: PropTypes.func,
+  }),
```
kpuputti: `.isRequired`?

client/app/components/composites/Avatar/Avatar.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+
+  render() {
+    return div({
+      onMouseOver: this.handleHover,
+      onMouseOut: this.handleHover,
+      className: classNames(this.props.className, css.avatar),
+    }, [
+      img({
```
Gnito: Could this be its own Element component? We are going to use it also elsewhere.

client/app/components/composites/Avatar/Avatar.js
```diff
@@ -0,0 +1,71 @@
...
+      this.setState({ menuOpen: true }); // eslint-disable-line react/no-set-state
+    } else if (event && event.type === 'mouseout') {
+      if (!this.debouncedClose) {
+        this.debouncedClose = _.debounce(() => {
+          this.setState({ menuOpen: false }); // eslint-disable-line react/no-set-state
+        }, this.menuCloseDelay);
+        this.debouncedClose();
+      }
+    }
+  }
```
kpuputti: Using JavaScript to handle hover state is bound to leave the menu open in weird situations, especially on mobile devices.

Maybe not version 1, but hopefully we could spend some time getting these dropdowns open with purely CSS if that's possible.

kpuputti:
👍 

The SVG files had ids in them, you could run them through [svgo](https://github.com/svg/svgo) (default settings are fine).

client/app/components/composites/Avatar/Avatar.js
mporkola: That's probably true. I'd still like to have a thorough hover vs click discussion, but yep, in any case let's fix this later.

PR #2134, opened 2016-06-16
Topbar draft added to the application.haml

Gnito:


app/controllers/application_controller.rb
```diff
@@ -630,6 +630,22 @@ def onboarding_topbar_props
...
+    {
+      logo: {
+        href: '/',
+        text: @current_community.name(I18n.locale),
+        image: @current_community.wide_logo.present? ? @current_community.wide_logo.url(:header) : nil,
+        image_highres: @current_community.wide_logo.present? ? @current_community.wide_logo.url(:header_highres) : nil
+      },
+      search_mode: 'keyword-and-location',
+      search_keyword_placeholder: 'Search...',
+      search_location_placeholder: 'Location'
```
mporkola: This _might_ be out of scope for this PR, but we probably should agree about a format for subcomponent props. We're using two different ones here: `logo` subobject vs separate search params. I see clear benefits in the subobject format: The container component doesn't need to care about the props of it's children, and Proptypes don't need to be written in two places.

Gnito: The other point is that with subobject combined props are UI / layout specific, and that should not be decided in backend (whether it's API or Rails)

Gnito: So, in this case, I'm also fine with grouping like:
```
{
  logo: {//as it is already},
  search: {
    mode: 'keyword-and-location',
    keyword_placeholder: //value from db,
    location_placeholder: //value from db,
  }
}
```

kpuputti: Let's not think too far in the future since we don't know how we'd use and share the props then.

Grouping 👍

app/views/layouts/application.haml
```diff
@@ -16,6 +16,12 @@
         - react_component_cache(name: "onboarding_topbar", props: props) do
           = react_component("OnboardingTopBar", props: props, prerender: true)
 
+  - if feature_enabled?(:topbar_v1)
+    - props = topbar_props
+    = javascript_include_tag "https://maps.googleapis.com/maps/api/js?libraries=places"
```
kpuputti: This is a different URL from the one used in index.haml above. Which one should we use?

client/app/components/elements/Logo/Logo.js
```diff
@@ -5,10 +5,15 @@ import { className } from '../../../utils/PropTypes';
 
 import css from './Logo.css';
 
-const logoContent = function logoContent(image, text) {
+const logoContent = function logoContent(image, imageHighRes, text) {
   let content = null;
   if (image) {
```
kpuputti: This only checks that `image` exists, but expects also `imageHighRes` as well. Maybe add that to the condition?

app/views/layouts/application.haml
Gnito: It seems that google has changed this url between versions v2 & v3

client/app/components/sections/Topbar/Topbar.js
```diff
@@ -10,11 +10,11 @@ class Topbar extends Component {
...
       r(Logo, { ...this.props.logo, classSet: css.topbarLogo }),
-      this.props.search_mode ?
+      this.props.search ?
         r(SearchBar, {
-          mode: this.props.search_mode,
-          keywordPlaceholder: this.props.search_keyword_placeholder,
-          locationPlaceholder: this.props.search_location_placeholder,
+          mode: this.props.search.mode,
+          keywordPlaceholder: this.props.search.keyword_placeholder,
+          locationPlaceholder: this.props.search.location_placeholder,
```
mporkola: This could be taken a still tad further and match the prop names above and below, so that you could just pass `this.props.search` as props for `SearchBar` component. A bit like it's done with `Logo`.

mporkola:
Otherwise, LET'S MERGE

client/app/components/sections/Topbar/Topbar.js
Gnito: I like the fact that we use camelCase in presentational components, but I want to leave further adjustments for Kimmo to be done on his own branch.

PR #2135, opened 2016-06-16
Link CLP assets to configurable remote url

ovan:
* Configurable url for linking to marketplace landing page content
* Remove community id resolving hack and use middleware provided object instead
* Refactor marketplace data fetching to a separate store

### DIFF
--- a/app/controllers/landing_page_controller.rb
+++ b/app/controllers/landing_page_controller.rb
@@ -77,15 +77,11 @@ class LandingPageController < ActionController::Metal
   end
 
   def community_id(request)
-    c = request.env[:current_marketplace]
-    if c
-      c.id
-    end
+    request.env[:current_marketplace]&.id
   end
 
   def render_landing_page(cid, structure)
-    locale = structure["settings"]["locale"]
-    sitename = structure["settings"]["sitename"]
+    locale, sitename = structure["settings"].values_at("locale", "sitename")
 
     render :landing_page,
            locals: { font_path: "/landing_page/fonts",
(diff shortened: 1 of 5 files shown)


app/controllers/landing_page_controller.rb
```diff
@@ -75,68 +77,30 @@ def parse_int(int_str_or_nil)
   end
 
   def community_id(request)
-    # TODO - This will come from request.env where the to-be-implemented middleware will put the data
-    ident = request.host.split(".").first
-    if ident == "aalto"
-      501
-    elsif ident == "oin"
-      11
+    c = request.env[:current_marketplace]
```
rap1ds: How about using fancy new Ruby 2.3 feature:

```
request.env[:current_marketplace]&.id
```

app/controllers/landing_page_controller.rb
```diff
@@ -75,68 +77,30 @@ def parse_int(int_str_or_nil)
...
+    if c
+      c.id
     end
   end
 
   def render_landing_page(cid, structure)
     locale = structure["settings"]["locale"]
+    sitename = structure["settings"]["sitename"]
```
rap1ds: Ok, now we can use `values_at` again :)

Previously, the locale variable was set like this:

```
locale = structure["settings"].values_at("locale")
```

The problem with this was that `values_at` returns an array and because of that, `locale` was an array, e.g. `["en"]`. That's why I changed it to `structure["settings"]["locale"]`. But now that we have two values we can use destructuring:

```
locale, sitename = structure["settings"].values_at("locale", "sitename")
```

PR #2144, opened 2016-06-20
Landing page fixes

rap1ds:
General fixes:

- [X] Flexbox browser prefixes
- [X] Add marketplace default color if the marketplace color is not set
- [X] Marketplace name as page title

Hero fixes:

- [X] Mark normalize.css content to be HTML safe so that it will be printed as is and not escaped. This will fix the issue where the search input had default layout (borders) enabled
- [X] Default search placeholder text
- [X] Hide default borders when search button is clicked
- [X] Cursor pointer when hovering search button
- [X] Fix [flexbox issues](https://github.com/philipwalton/flexbugs#1-minimum-content-sizing-of-flex-items-not-honored) with Safari

Footer fixes:

- [X] Bigger click area for links footer links
- [X] Center footer links if there are no social media icons
- [X] Center social media icons if there are no links

### DIFF
--- a/app/assets/stylesheets/landing_page/sections/hero.scss
+++ b/app/assets/stylesheets/landing_page/sections/hero.scss
@@ -91,6 +91,7 @@
 
 .hero__search-bar {
   @include prefix-val(display, flex);
+  @include prefix-prop(flex-shrink, 1);
   margin: 0 auto;
   max-width: 532px;
 }
@@ -157,7 +158,7 @@
   padding: 0;
   border-radius: 0px 29.5px 29.5px 0px;
 
-  padding: 14px 28px;
+  padding: 14px 32px 14px 28px;
 
   line-height: 32px;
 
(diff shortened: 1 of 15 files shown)


app/views/landing_page/landing_page.erb
```diff
@@ -29,7 +29,7 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
 
-  <title>Custom Landing Page</title>
+  <title><%= settings["marketplace_name"]["value"] %></title>
```
ovan: Define a new top level property that would hold global data and link to marketplace data from there. Other stuff that would come under the top level property would be things like meta tags for SEO, i.e. things that cannot be disabled and don't have a position like sections.

app/assets/stylesheets/landing_page/sections/hero.scss
```diff
@@ -166,6 +167,10 @@
   color: #FFFFFF;
   letter-spacing: 0px;
 
+  outline: none;
```
kpuputti: Focus outline is an important accessibility feature when tabbing through links, buttons, etc. to see which one is focused. You should make sure the button has some visual styles that differentiate when it is in focus.

PR #2145, opened 2016-06-20
Clear derivative hidden fields' values.

Gnito:


kpuputti:
Doesn't this loose important information if the `#homepage-filters` form (that these inputs are part of) is submitted by changing something else than location, e.g. filters?

PR #2146, opened 2016-06-21
Topbar search submit

kpuputti:
Construct the search query from the existing URL params and whatever the SearchBar component returns.

Required params:

 - locale: `locale`
 - view type: `view` (map, list, grid)
 - keyword query: `q`
 - location query: `lq`
 - coordinates: `lc`
 - places service status: `ls`
 - bounding box viewport for the map: `boundingbox`
 - max search distance: `distance_max`

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -72,9 +72,9 @@ class HomepageController < ApplicationController
         @listings = listings # TODO Remove
 
         if @view_type == "grid" then
-          render partial: "grid_item", collection: @listings, as: :listing, locals: { show_distance: location_search_in_use }
-        elsif location_search_in_use
-          render partial: "list_item_with_distance", collection: @listings, as: :listing, locals: { shape_name_map: shape_name_map, testimonials_in_use: @current_community.testimonials_in_use, show_distance: location_search_in_use }
+          render partial: "grid_item", collection: @listings, as: :listing, locals: { show_distance: location_in_use }
+        elsif location_in_use
+          render partial: "list_item_with_distance", collection: @listings, as: :listing, locals: { shape_name_map: shape_name_map, testimonials_in_use: @current_community.testimonials_in_use, show_distance: location_in_use }
         else
           render partial: "list_item", collection: @listings, as: :listing, locals: { shape_name_map: shape_name_map, testimonials_in_use: @current_community.testimonials_in_use }
         end
@@ -93,7 +93,7 @@ class HomepageController < ApplicationController
                  testimonials_in_use: @current_community.testimonials_in_use,
                  listing_shape_menu_enabled: listing_shape_menu_enabled,
                  main_search: main_search,
-                 location_search_in_use: location_search_in_use,
+                 location_search_in_use: location_in_use,
                  viewport: viewport }
       }.on_error { |e|
         flash[:error] = t("homepage.errors.search_engine_not_responding")
@@ -107,7 +107,7 @@ class HomepageController < ApplicationController
                  testimonials_in_use: @current_community.testimonials_in_use,
                  listing_shape_menu_enabled: listing_shape_menu_enabled,
                  main_search: main_search,
-                 location_search_in_use: location_search_in_use,
+                 location_search_in_use: location_in_use,
                  viewport: viewport }
       }
     end
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -151,8 +151,9 @@ class HomepageController < ApplicationController
     numbers = numeric_search_params.map { |numeric| numeric.merge(type: :numeric_range) }
 
     marketplace_configuration = MarketplaceService::API::Api.configurations.get(community_id: @current_community.id).data
+    distance_system = marketplace_configuration ? marketplace_configuration[:distance_unit] : nil
 
-    distance_unit = (location_search_in_use && marketplace_configuration[:distance_unit] == :metric) ? :km : :miles
+    distance_unit = (location_search_in_use && distance_system == :metric) ? :km : :miles
     limit_search_distance = marketplace_configuration ? marketplace_configuration[:limit_search_distance] : true
     location_search_hash = location_search_params(
       params[:lc],
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -55,14 +55,16 @@ class HomepageController < ApplicationController
       end
 
     main_search = FeatureFlagHelper.location_search_available ? MarketplaceService::API::Api.configurations.get(community_id: @current_community.id).data[:main_search] : :keyword
-    location_in_use, keyword_in_use = search_modes_in_use(params[:q], params[:lc], main_search)
+    search_modes = search_modes_in_use(params[:q], params[:lc], main_search)
+    keyword_in_use = search_modes[:keyword]
+    location_in_use = search_modes[:location]
 
     search_result = find_listings(params, per_page, compact_filter_params, includes.to_set, location_in_use, keyword_in_use)
 
     shape_name_map = all_shapes.map { |s| [s[:id], s[:name]]}.to_h
 
     if @view_type == 'map'
-      viewport = viewport_geography(params[:boundingbox], params[:lc], @current_community.location)
+      viewport = viewport_geometry(params[:boundingbox], params[:lc], @current_community.location)
     end
 
     if request.xhr? # checks if AJAX request
@@ -304,13 +306,13 @@ class HomepageController < ApplicationController
   end
 
   def search_modes_in_use(q, lc, main_search)
-    [
-      lc && (main_search == :location || main_search == :keyword_and_location),
-      q && (main_search == :keyword || main_search == :keyword_and_location)
-    ]
+    {
+      keyword: q && (main_search == :keyword || main_search == :keyword_and_location),
+      location: lc && (main_search == :location || main_search == :keyword_and_location),
+    }
   end
 
-  def viewport_geography(boundingbox, lc, community_location)
+  def viewport_geometry(boundingbox, lc, community_location)
     coords = Maybe(boundingbox).split(',').or_else(nil)
     if coords
       sw_lat, sw_lng, ne_lat, ne_lng = coords
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -5,8 +5,6 @@ class HomepageController < ApplicationController
 
   APP_DEFAULT_VIEW_TYPE = "grid"
   VIEW_TYPES = ["grid", "list", "map"]
-  APP_MINIMUM_DISTANCE_MAX = 5
-
 
   def index
     redirect_to landing_page_path and return if no_current_user_in_private_clp_enabled_marketplace?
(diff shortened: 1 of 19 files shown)


app/view_utils/topbar_helper.rb
```diff
@@ -54,10 +57,15 @@ def topbar_props(community:, path_after_locale_change:, user: nil, search_placeh
...
-        location_placeholder: 'Location'
+        mode: main_search.to_s,
+        keyword_placeholder: 'Search...', # TODO: localise
+        location_placeholder: 'Location', # TODO: localise
+
+        # TODO: figure where to get these
+        # keyword_query: params[:q],
+        # location_query: params[:lq]
       },
+      search_path: '/', # TODO: fix
```
kpuputti: This could be passed within the search object.

client/app/components/sections/Topbar/Topbar.js
```diff
@@ -1,5 +1,8 @@
+/* eslint-disable no-console */
```
kpuputti: This is temporary and should be removed when the logging is removed.

client/app/components/sections/Topbar/Topbar.js
```diff
@@ -41,7 +44,7 @@ const avatarDropdownProps = (avatarDropdown, customColor, username, routes) => {
 const LABEL_TYPE_MENU = 'menu';
 const LABEL_TYPE_DROPDOWN = 'dropdown';
 
-const SEARCH_ENABLED = false;
+const SEARCH_ENABLED = true;
```
kpuputti: This  whole flag should be removed when the search works.

client/app/components/sections/Topbar/Topbar.js
```diff
@@ -92,6 +95,55 @@ const DEFAULT_CONTEXT = {
   loggedInUsername: null,
 };
 
+const parseQuery = (searchQuery) => {
```
kpuputti: See if the [URL](https://developer.mozilla.org/en/docs/Web/API/URL) API could be used for parsing.

client/app/startup/TopbarApp.js
```diff
@@ -1,5 +1,6 @@
 import r from 'r-dom';
 import { initialize as initializeI18n } from '../utils/i18n';
+
 import Topbar from '../components/sections/Topbar/Topbar';
 import { subset } from '../utils/routes';
```
kpuputti: TopbarApp should be made to use the new TopbarContainer to get the actions.

client/app/components/sections/Topbar/Topbar.js
```diff
@@ -121,6 +121,41 @@ const DEFAULT_CONTEXT = {
   loggedInUsername: null,
 };
 
+const currentSearchParams = (searchQuery) => {
+  const PARAMS_TO_KEEP = ['view', 'locale'];
```
kpuputti: Are these the only params we want to keep form the current URL?

client/app/components/composites/MenuPriority/MenuPriority.css
```diff
@@ -26,4 +26,7 @@
   }
 }
 
-.hiddenLinks {} /* stylelint-disable-line block-no-empty */
+.hiddenLinks {
+  /* Fixes dropdown being below old filter controls and admin panel selectize inputs. */
+  z-index: 2;
```
Gnito: How about other dropdowns? Language (another Menu) & AvatarDropdown

client/app/components/composites/SearchBar/SearchBar.css
```diff
@@ -390,4 +322,28 @@
...
+  position: absolute;
+  bottom: -1px;
+  right: calc(var(--SearchBar_mobileHeight) / 2 - var(--SearchBar_mobileTriangleSize));
+  border: var(--SearchBar_mobileTriangleSize) solid transparent;
+
+  /* We must adjust the brightness here to match either the input
+     background or the submit button background when the toggle
+     position changes in the breakpoint. This cannot be done in JS by
+     switching the color since the breakpoints are unknown there. */
+  filter: brightness(80%);
```
Gnito: Filter brightness doesn't work well with IE11 & Edge. 
Since this arrow is invisible most of the time, so as an alternative, we could just create 2 arrows (with custom colors from client/app/utils/colors.js) and show them based on media queries?

Gnito: It seems we already calculate this brightness(80%) in component

client/app/components/sections/Topbar/Topbar.js
```diff
@@ -233,18 +263,31 @@ class Topbar extends Component {
...
+          keywordQuery: oldSearchParams.q,
+          locationQuery: oldSearchParams.lq,
+          customColor: marketplace_color1,
+          onSubmit: ({ keywordQuery, locationQuery, place }) => {
+            const query = createQuery({
+              q: keywordQuery,
+              lq: locationQuery,
+              lc: placesUtils.coordinates(place),
+              boundingbox: placesUtils.viewport(place),
+              distance_max: placesUtils.maxDistance(place),
```
Gnito: One additional parameter to consider is service status 
(we used that to figure out Google maps related API problems like quota etc.)
However, it can be added after dl if needed.

client/app/components/composites/SearchBar/SearchBar.css
kpuputti: You are correct. This CSS side is older and the brightness calculation in the component is a more recent addition. I'll set the color in the JS side.

kpuputti: No wait, I cannot really do this. The reason why the filter is in the CSS is that it changes based on the viewport width.

On a narrow screen the arrow is next to the darker submit button:

screen shot 2016-08-09 at 10 14 50

but on a wider screen the arrow is next to the lighter input background:

screen shot 2016-08-09 at 10 18 17

And since JS isn't aware of the breakpoints we cannot change the color dynamically without a resize handler that checks the current breakpoint. And I'd like to avoid doing that.

kpuputti: One possible solution would be to create two separate arrows that have bg color assigned in JS, and the CSS would just swap between those in the breakpoint.

client/app/components/composites/MenuPriority/MenuPriority.css
Gnito: I'll make another Trello todo task from this

client/app/components/sections/Topbar/Topbar.js
kpuputti: Done.

client/app/components/composites/MenuPriority/MenuPriority.css
kpuputti: Let's do this in another PR.

client/app/components/composites/SearchBar/SearchBar.css
kpuputti: Let's postpone this since the colors are a bit open anyways.

app/controllers/homepage_controller.rb
```diff
@@ -286,28 +297,31 @@ def search_coordinates(latlng)
...
       @current_community.private &&
       !@current_user
   end
 
+  def search_modes_in_use(q, lc, main_search)
+    [
+      lc && (main_search == :location || main_search == :keyword_and_location),
+      q && (main_search == :keyword || main_search == :keyword_and_location)
+    ]
+  end
```
lyyder: Does this return an array of booleans which indicates what search modes are in use? Would it be clearer if the search mode order was not swapped (param order vs. return value)?

app/controllers/homepage_controller.rb
```diff
@@ -57,34 +55,24 @@ def index
...
-        sw_lat, sw_lng, ne_lat, ne_lng = coords
-        { boundingbox: { sw: [sw_lat, sw_lng], ne: [ne_lat, ne_lng] } }
-      elsif params[:lc].present?
-        { center: params[:lc].split(',') }
-      else
-        Maybe(@current_community.location)
-          .map { |l| { center: [l.latitude, l.longitude] }}
-          .or_else(nil)
-      end
+      viewport = viewport_geography(params[:boundingbox], params[:lc], @current_community.location)
```
mporkola: I think `viewport_geometry` is a better term here

lyyder:
👍  for the js utility libs and the API docs on those.

client/app/components/sections/Topbar/Topbar.css
```diff
@@ -54,16 +54,29 @@
     }
   }
 
+  & button {
```
mporkola: Since we've ended up using tag selector (I trust it was a well-considered choice), I'd like to make it as specific as possible. So maybe a child selector (`& > button`) instead of a more broader descendant selector?

client/app/components/sections/Topbar/Topbar.css
```diff
@@ -54,16 +54,29 @@
...
+    border-radius: 0;
+  }
+
   & .topbarLogo {
     max-height: var(--Topbar_logoMaxHeightMobile);
-    max-width: var(--Topbar_textLogoMaxWidth);
+    max-width: 70%;
```
mporkola: To me this looks like the `max-width` is 70% of the whole top bar container, that's probably wrong?

client/app/components/sections/Topbar/Topbar.js
```diff
@@ -121,6 +121,36 @@ const DEFAULT_CONTEXT = {
   loggedInUsername: null,
 };
 
+const SEARCH_PARAMS_TO_KEEP = ['view', 'locale'];
+const parseKeepParams = urlUtils.currySearchParams(SEARCH_PARAMS_TO_KEEP);
+const SEARCH_PARAMS = ['q', 'lq'];
+const parseSearchParams = urlUtils.currySearchParams(SEARCH_PARAMS);
+
+const isValid = (value) => typeof value === 'number' && !isNaN(value) || !!value;
```
mporkola: maybe more information about what kind of "valid" we're talking about? e.g. `isValidSearchParam`

client/app/utils/places.js
```diff
@@ -0,0 +1,151 @@
...
+
+const toRadians = (degrees) => degrees * (Math.PI / 180); // eslint-disable-line no-magic-numbers
+
+/**
+ * Calculate the distance between two points.
+ *
+ * @param {google.maps.LatLng} pointA - first point
+ * @param {google.maps.LatLng} pointB - second point
+ *
+ * @return {Number} distance between the given points (in km?)
```
mporkola: Yes, in km :) This comes from the unit used in `EARTH_RADIUS` constant below: if we want rods or furlongs we can just change the radius to desired unit.

client/app/utils/places.js
```diff
@@ -0,0 +1,151 @@
...
+ */
+const computeScale = (pointA, pointB) => {
```
mporkola: The naming of this function was my failure, it should be fixed :) It actually doesn't do what the document promises (I think?), but instead returns `distance / 2`. I suggest this should be changed to `getDistance` and then the `/ 2` be done elsewhere.

app/controllers/homepage_controller.rb
Gnito: :+1:
We thought that destructuring was possible with Hashes, but ruby didn't like it so had to change it to array - and we didn't notice this one

client/app/components/sections/Topbar/Topbar.css
Gnito: :+1: or explicit comment that this is override of R&A styles

app/controllers/homepage_controller.rb
kpuputti: Good point. This was originally a hash that was attempted to destructure, but was changed to an array. I'll change this.

client/app/utils/places.js
Gnito: Howabout getBoundingboxRadius? We don't need distance between 2 points anywhere atm. (ie. splitting can be done later too if necessary)

client/app/components/sections/Topbar/Topbar.css
kpuputti: Those generic selectors within the Topbar.css file can be thought of being our CSS normalisation from the old styles.

That specific rule undoes the rounded corners that we have in all buttons in old styles, so I think it makes sense to keep quite generic and get the same effect to all future buttons in the Topbar.

Does this make sense to you?

client/app/components/sections/Topbar/Topbar.css
kpuputti: That's exactly what it is, and it's on purpose.

The max-width is for the text logo in the mobile breakpoint where we want to fit as much text in the Topbar as we can.

client/app/components/sections/Topbar/Topbar.css
mporkola: Yes, makes sense. Just for clarity, would it be possible to make a `resetStylesLeakingFromRails.css` (naming left as exercise) and include it to root components?

client/app/utils/places.js
mporkola: I think radius isn't a correct term, and in any case "distance" is exactly what we're calculating here. The change is very minor, I can do it myself as well :)

client/app/components/sections/Topbar/Topbar.css
kpuputti: @mporkola Great idea!

PR #2149, opened 2016-06-21
Add hover transition to buttons and links

rap1ds:


app/assets/stylesheets/landing_page/sections/hero.scss
```diff
@@ -239,6 +243,14 @@
 
   padding: 18px 36px;
 
+  // Hover color is used to indicate tab focus.
+  // The hover color is set in the layout
+  outline: none;
```
kpuputti: Are the `:focus` styles coming from somewhere else? An element can be in focus, but not in hover state.

app/assets/stylesheets/landing_page/mixins/_transitions.scss
```diff
@@ -0,0 +1,19 @@
...
+ * }
+ *
+ * To specify which property should be animated, use:
+ *
+ * .my-button:hover {
+ *   @include hover-transition(background-color)
+ * }
+ *
+ */
+@mixin hover-transition($transition-prop: all) {
```
kpuputti: Transitioning all properties can be bad for performance, so keep an eye on this.

If you need to explicitly list multiple transitions without using `all`, this mixin wouldn't work anymore since they have to be comma separated.

rap1ds: Ok! Good to keep this in mind.

app/assets/stylesheets/landing_page/sections/hero.scss
rap1ds: Here: https://github.com/sharetribe/sharetribe/blob/master/app/views/landing_page/_hero.erb#L14

PR #2150, opened 2016-06-21
Add non-breaking space to avoid widows

rap1ds:
Same as this https://css-tricks.com/snippets/jquery/add-non-breaking-space-on-title-to-prevent-widows/ but done with Ruby on the server side. Also, we add the non-breaking space to last three words so that there's always at least three words in the last line.

### DIFF
--- a/app/views/landing_page/_hero.erb
+++ b/app/views/landing_page/_hero.erb
@@ -19,11 +19,18 @@
   }
 <% end %>
 
+<%
+# Add non-breaking space to prevent widows
+subtitle_splitted = s["subtitle"]["value"].split(" ")
+last_three = subtitle_splitted.pop(3)
+subtitle = (subtitle_splitted + [last_three.join("&nbsp;")]).join(" ").html_safe
+%>
+
 <section id="<%= section_id %>" class="hero__section">
   <div class="hero__content">
     <div class="hero__content-container">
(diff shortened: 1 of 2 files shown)


app/views/landing_page/_hero.erb
```diff
@@ -19,11 +19,18 @@
   }
 <% end %>
 
+<%
+# Add non-breaking space to prevent widows
+subtitle_splitted = s["subtitle"]["value"].split(" ")
```
kpuputti: Can the text be split to another line from some other character than space, e.g. hyphen?

rap1ds: Yep, it seems so. I tested with this text:

```
SuperlongtextwithnobreaksSuperlongtextwithnobreaks-Superlongtextwithnobreaks
```

and the browser indeed added a linebreak where the hyphen is. But can we do anything about it? We can't replace hyphens with `&nbsp;` because then we would have space instead of hyphen. On the other hand, we can't add `&nbsp;` where the hyphen is, because then we would have both, hyphen and space.

kpuputti: At least hyphens might be simply replaced with their non-breaking versions: http://stackoverflow.com/a/7691604 but I don't know if there are some other characters still.

PR #2159, opened 2016-06-22
Implement caching for landing page index

bladealslayer:
Page is cached in Rails cache (memcached) and client-side for 15 minutes (configurable).

### DIFF
--- a/app/controllers/landing_page_controller.rb
+++ b/app/controllers/landing_page_controller.rb
@@ -9,6 +9,7 @@ class LandingPageController < ActionController::Metal
   # See Rendering Helpers: http://api.rubyonrails.org/classes/ActionController/Metal.html
   #
   include AbstractController::Rendering
+  include ActionController::ConditionalGet
   include ActionView::Layouts
   append_view_path "#{Rails.root}/app/views"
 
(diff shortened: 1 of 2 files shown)


app/controllers/landing_page_controller.rb
```diff
@@ -20,13 +21,19 @@ def index
...
-      structure = CLP::LandingPageStore.load_structure(cid, version)
+    cache_time = APP_CONFIG[:clp_cache_time] ? APP_CONFIG[:clp_cache_time].to_i : 60
+    cache_time = cache_time.seconds
 
-      render_landing_page(cid, structure)
+    begin
+      lp_html = Rails.cache.fetch("clp/#{cid}/#{version}", expires_in: cache_time) do
+        structure = CLP::LandingPageStore.load_structure(cid, version)
+        html = render_landing_page(cid, structure)
+        html
```
rap1ds: Unnecessary assignment to `html`.

app/controllers/landing_page_controller.rb
```diff
@@ -20,13 +21,19 @@ def index
     cid = community_id(request)
     version = CLP::LandingPageStore.released_version(cid)
 
-    # TODO Ideally we would do the caching based upon just clp_version
-    # and avoid loading and parsing the (potentially) big structure
-    # JSON.
-    begin
-      structure = CLP::LandingPageStore.load_structure(cid, version)
+    cache_time = APP_CONFIG[:clp_cache_time] ? APP_CONFIG[:clp_cache_time].to_i : 60
```
ovan: I think the fallback is unnecessary because defaults already specified the default value so this always exists.

app/controllers/landing_page_controller.rb
bladealslayer: Yep, there used to be something in between :) Sneaky LOC

ovan:
This looks good for merge. One question, the etag middleware handles conditional gets by checking if-none-match but I guess we don't handle the "if-modified-since" case now? I expose my ignorance by saying I'm not sure that's critical when etags are the preferred method and widely supported(?).

bladealslayer:
@ovan tbh, ,I don't know either. I'll check. But at least in my previous tests, Chrome wasn't sending `If-Modified-Since` at all. I guess because it supports ETag.

bladealslayer:
Ok, I found a way to deal with both etags and if-modified-since headers, @ovan. The key is the `fresh_when` function. So I'll rework as we initially planned - only meta first fetched from cache and full content only when needed.

app/controllers/landing_page_controller.rb
```diff
@@ -20,13 +21,32 @@ def index
     cid = community_id(request)
     version = CLP::LandingPageStore.released_version(cid)
 
-    # TODO Ideally we would do the caching based upon just clp_version
-    # and avoid loading and parsing the (potentially) big structure
-    # JSON.
+    cache_time = APP_CONFIG[:clp_cache_time].to_i.seconds
```
ovan: Should be a constant. config properties never change dynamically, not even in rails dev mode.

ovan:
@bladealslayer I rewrote this a bit. I'll pass the review ball back to you now so you can let me know if you're ok with the changes.

ovan:
@bladealslayer On second thought, I decided to just squash and move forward with this. Feel free to post-merge review and react.

PR #2160, opened 2016-06-23
Topbar mobile menu

Gnito:
This needs a review or handover.

### DIFF
--- a/client/app/assets/styles/variables.js
+++ b/client/app/assets/styles/variables.js
@@ -1,4 +1,5 @@
 /* eslint-env node */
+const fontSizeSmaller = '12px';
 const fontSizeSmall = '13px';
 const fontSize = '14px';
 const fontSizeBig = '16px';
@@ -11,11 +12,14 @@ const backgroundLightColorHover = 'rgba(169, 172, 176, 0.07)';
 const customColorFallback = '#a64c5d';
 const customColor2Fallback = '#00a26c';
 
-const paddingTopbarVertical = '0.92857em';
-const paddingTopbarHorizontal = '1.7142857em';
 const topbarBorderColor = 'rgba(169, 172, 176, 0.5)';
 const topbarItemHeight = '44px';
 const bodyPadding = '24px';
+
+const pxToEms = function pxToEms(px, againstFontSize) {
+  const emValue = px / againstFontSize;
+  return `${emValue}em`;
+};
 
 module.exports = {
 
@@ -105,6 +109,7 @@ module.exports = {
    // on an input.
   '--Topbar_inputFontSizeMobile': fontSizeBig,
   '--Topbar_fontSize': fontSize,
+  '--Topbar_fontSizeMobile': fontSizeMobile,
 
   '--Topbar_avatarSize': topbarItemHeight,
   '--Topbar_avatarPadding': '18px 0',
@@ -149,10 +154,14 @@ module.exports = {
 
   '--MenuItem_borderColor': topbarBorderColor,
   '--MenuItem_backgroundColorHover': backgroundLightColorHover,
-  '--MenuItem_paddingTopbarVertical': paddingTopbarVertical,
-  '--MenuItem_paddingTopbarHorizontal': paddingTopbarHorizontal,
+  '--MenuItem_paddingTopbarVertical': pxToEms(13, 14), // eslint-disable-line no-magic-numbers
+  '--MenuItem_paddingTopbarHorizontal': pxToEms(24, 14), // eslint-disable-line no-magic-numbers
+  '--MenuItem_paddingOffScreenVertical': pxToEms(10, 17), // eslint-disable-line no-magic-numbers
+  '--MenuItem_paddingOffScreenHorizontal': pxToEms(24, 17), // eslint-disable-line no-magic-numbers
   '--MenuItem_textColor': textColor,
   '--MenuItem_textColorFocus': textColorFocus,
+  '--MenuItem_textColorDefault': '#a64c5d',
+  '--MenuItem_textColorSelected': '#4a4a4a',
   '--MenuItem_letterSpacing': '0.09px',
 
   '--Menu_fontSize': fontSize,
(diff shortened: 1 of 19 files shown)


client/app/components/composites/Menu/Menu.js
```diff
@@ -37,7 +37,7 @@ class Menu extends Component {
     const menuLabel = ReactDOM.findDOMNode(this.menuLabel);
 
     this.setState({ // eslint-disable-line react/no-did-mount-set-state, react/no-set-state
-      arrowPosition: menuLabel.offsetWidth / 2, // eslint-disable-line no-magic-numbers
+      arrowPosition: menuLabel.offsetWidth > 40 ? menuLabel.offsetWidth / 2 : 20, // eslint-disable-line no-magic-numbers
```
kpuputti: This `componentDidMount` + `setState` + `offsetWidth` concerns me. This makes the component render twice initially, with the `offsetWidth` check causing an additional Layout Trashing in the middle. But I don't have a better solution for this.

What are the magic numbers 40 and 20?

client/app/assets/styles/variables.js
```diff
@@ -139,8 +139,10 @@ module.exports = {
 
   '--MenuItem_borderColor': topbarBorderColor,
   '--MenuItem_backgroundColorHover': backgroundLightColorHover,
-  '--MenuItem_paddingTopbarVertical': paddingTopbarVertical,
-  '--MenuItem_paddingTopbarHorizontal': paddingTopbarHorizontal,
+  '--MenuItem_paddingTopbarVertical': '0.92857em',
+  '--MenuItem_paddingTopbarHorizontal': '1.7142857em',
+  '--MenuItem_paddingOffScreenVertical': '0.58823529411em',
+  '--MenuItem_paddingOffScreenHorizontal': '1.41176470588em',
```
kpuputti: The em value reminds me of this: https://twitter.com/pcwalton/status/718565939432718336

Where do you get these values and do we really need that kind of precision?

kpuputti:
Let's not invent a new naming scheme for the customisation class names. We should just put the component name there, e.g. `.MenuLabelMobile` if the component name is `MenuLabelMobile`.

This also helps us in the development when we can see the component name in the DOM tree.

client/app/components/composites/MenuMobile/MenuMobile.js
```diff
@@ -0,0 +1,99 @@
+import { Component, PropTypes } from 'react';
+import r, { div } from 'r-dom';
+
+import css from './MenuMobile.css';
+import OffScreenMenu from './OffScreenMenu';
+import MenuLabelMobile from './MenuLabelMobile';
+
+// const isTouch = !!(typeof window !== 'undefined' && (('ontouchstart' in window) || window.navigator.msMaxTouchPoints > 0));
```
mporkola: This (and related checks) probably should be deleted?

client/app/components/composites/Menu/Menu.js
Gnito: Yeah, the problem is that arrowPosition must be calculated after the component has connection to DOM. I can add some better wording to the magic numbers (which are just setting the arrow to some minimum distance from dropdown box's corner (rounded corner)

client/app/assets/styles/variables.js
Gnito: Yeah, values come from conversions (some padding / font-size). The problem is that since we can't use rems here, I need to use ems, and font-sizes varies a lot on Janne's design.  I'll change this to pixels.

client/app/components/composites/MenuMobile/MenuMobile.js
Gnito: Yes!

client/app/assets/styles/variables.js
```diff
@@ -7,12 +8,15 @@ const textColor = 'rgb(82, 89, 97)';
...
-const paddingTopbarVertical = '0.92857em';
-const paddingTopbarHorizontal = '1.7142857em';
 
 const topbarBorderColor = 'rgba(169, 172, 176, 0.5)';
 const bodyPadding = '18px';
 
+const pxToEms = function pxToEms(px, againstFontSize) {
+  const emValue = px / againstFontSize;
+  return `${emValue}em)`;
+};
```
kpuputti: This should go e.g. to `client/app/assets/styles/functions.js` and in use with `postcss-functions` in the future so it can be used in all CSS files, but fine here for now.

client/app/components/composites/MenuMobile/MenuMobile.css
```diff
@@ -0,0 +1,123 @@
...
+
+.overlay {
+  display: none;
+  cursor: pointer;
+  position: fixed;
+  top: 0;
+  left: 0;
+  z-index: 9;
+  height: 100vh;
+  width: 100vw;
```
kpuputti: Wouldn't 100% width and height work here just as well, and have better browser support?

client/app/components/elements/MenuItem/MenuItem.js
```diff
@@ -8,10 +8,14 @@ class MenuItem extends Component {
   constructor(props, context) {
     super(props, context);
     this.activeColor = this.props.activeColor || '#a64c5d';
+    this.textColor = this.props.active ? '#4A4A4A ' : this.props.textColor || '#a64c5d';
```
kpuputti: These default colors should probably be in `variables.js` and imported here.

kpuputti:
Customisation class names should be `.ComponentName` for a ComponentName component, or `.ComponentName_modifier` for some inner element that isn't a component itself. This way they are completely separate from our inner class names and communicate the component they affect.

Otherwise, when the class names are changed and the linting errors fixed, this should be good to go.

client/app/assets/styles/variables.js
Gnito: I'm not sure is this good for css files - maybe.

client/app/components/sections/Topbar/Topbar.js
```diff
@@ -25,18 +26,70 @@ const avatarDropdownProps = (avatarDropdown) => {
...
+  return [];
+};
+
+
 class Topbar extends Component {
   render() {
+    const marketplaceContext = this.props.railsContext ?
+      this.props.railsContext :
+      { marketplace_color1: '#a64c5d',
+        marketplace_color2: '#00a26c',
```
mporkola: Use default colors from `variables.js`

client/app/assets/styles/variables.js
```diff
@@ -139,10 +144,14 @@ module.exports = {
...
-  '--MenuItem_paddingTopbarVertical': paddingTopbarVertical,
-  '--MenuItem_paddingTopbarHorizontal': paddingTopbarHorizontal,
+  '--MenuItem_paddingTopbarVertical': pxToEms(13, 14), // eslint-disable-line no-magic-numbers
+  '--MenuItem_paddingTopbarHorizontal': pxToEms(24, 14), // eslint-disable-line no-magic-numbers
+  '--MenuItem_paddingOffScreenVertical': pxToEms(10, 17), // eslint-disable-line no-magic-numbers
+  '--MenuItem_paddingOffScreenHorizontal': pxToEms(24, 17), // eslint-disable-line no-magic-numbers
   '--MenuItem_textColor': textColor,
   '--MenuItem_textColorFocus': textColorFocus,
+  '--MenuItem_textColorDefault': '#a64c5d',
+  '--MenuItem_textColorSelected': '#4a4a4a',
```
mporkola: change to exportable variables

PR #2162, opened 2016-06-23
Fix big cover displayed when landing page is enabled

bladealslayer:


rap1ds:
The code looks good and this can be merged as it is now!

However, if you feel like doing a small refactoring to the existing code, here's an idea that would make this functionality a bit cleaner and faster (saves 1 db query per request):

There are 3 places where `show_big_cover_photo?` is called:

```
app/helpers/application_helper.rb:662:19:    block.call if show_big_cover_photo?
app/helpers/application_helper.rb:667:23:    block.call unless show_big_cover_photo?
app/helpers/application_helper.rb:670:7:  def show_big_cover_photo?
app/views/homepage/index.haml:39:37:- if @current_community.private? && show_big_cover_photo?
```

Instead of having a helper functions `with_big_cover_photo` and `with_small_cover_photo` (first two in the list above), we could just have a instance variable `@show_big_cover_photo`. And we could set that instance variable to `true` in HomepageController#index, because that's the only place where we actually want to show the big cover photo. Otherwise we can leave it to uninitialized (i.e. `nil`). That way we would not need to query the landing page enabled status per each request, but only in HomepageController#index.

So, to recap, my suggestion is to move the logic to HomepageController and set `@show_big_cover_photo` instance variable there and get rid of all helper functions that use the `show_big_cover_photo?` method.

bladealslayer:
I'll merge this and make separate PR for the refactoring

PR #2165, opened 2016-06-23
Pull topbar props out of app ctrl. Fix some bugs.

ovan:
* Independently callable topbar props method
* Fix bug in current_locale handling
* Remove unnecessary helpers from app ctrl that were used in building RailsContext. There's a better way.

### DIFF
--- a/config/initializers/react_on_rails.rb
+++ b/config/initializers/react_on_rails.rb
@@ -2,9 +2,10 @@ module RailsContextExtension
   # Return a Hash that contains custom values from the view context that will get passed to
   # all calls to react_component and redux_store for rendering
   def self.custom_context(view_context)
+    community = view_context.request.env[:current_marketplace]
     {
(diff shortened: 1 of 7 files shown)


config/initializers/react_on_rails.rb
```diff
@@ -2,9 +2,10 @@ module RailsContextExtension
   # Return a Hash that contains custom values from the view context that will get passed to
   # all calls to react_component and redux_store for rendering
   def self.custom_context(view_context)
+    community = view_context.request.env[:current_marketplace]
     {
-      marketplaceId: view_context.controller.current_community_id,
-    }.merge(view_context.controller.current_community_custom_colors)
+      marketplaceId: community&.id,
```
Gnito: How do I add current_user.username here too?
That is needed for routes

ovan: Just pull view_context.controller.current_user for now.

PR #2166, opened 2016-06-23
Valid OAuth - problems

Warond:
I had an Instant Article, everything was ok... and suddenly I have a problem, when I tried to add another page on the my site. After I deleted all details from that page, I still have a problem.

I'm using Facebook Instant Articles Plugin.

I got this msg:

URL Blocked: This redirect failed because the redirect URI is not whitelisted in the app’s Client OAuth Settings. Make sure Client and Web OAuth Login are on and add all your app domains as Valid OAuth Redirect URIs. (when I put https://___________.com/login/callback)

When I put just (https://_________.com), I can't list pages through the Instant Article Plugin.

How to resolve the problem? What to do?

When I tried a new app, I got same problem.

Kind Regards,
Alonzo

https://wordpress.org/plugins/fb-instant-articles/

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -124,6 +124,10 @@ class SessionsController < ApplicationController
         redirect_to sign_up_path and return
       end
 
+      if data.username.blank? && data.family_name.blank? && data.given_name.blank?
+        FacebookSignupFails.create(auth_data: data, community_id: @current_community.id)
+      end
+
       facebook_data = {"email" => data.email,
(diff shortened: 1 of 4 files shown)


rap1ds:
Hi @Warond ! I guess you meant to create an Issue instead of a Pull Request?

I'm closing this PR now.

If this is a bug in Sharetribe code, please open a new Issue with exact steps to reproduce. If you're just asking for help, the right place to ask support is the community forum https://www.sharetribe.com/community

PR #2167, opened 2016-06-23
AddNewListing button to Topbar

kpuputti:


client/app/components/elements/AddNewListingButton/AddNewListingButton.js
```diff
@@ -0,0 +1,48 @@
...
+  ]);
+}
+
+AddNewListingButton.propTypes = {
+  text: PropTypes.string.isRequired,
+  url: PropTypes.string.isRequired,
+
+  // Marketplace color or default color
+  customColor: PropTypes.string,
+};
```
rap1ds: I like to use destruct the `PropTypes`. It makes the type definition a bit easier to read:

```js
const { string } = PropTypes;

AddNewListingButton.propTypes = {
  text: string.isRequired,
  url: string.isRequired,

  // Marketplace color or default color
  customColor: string,
};
```

No need to change this.

app/view_utils/topbar_helper.rb
```diff
@@ -58,6 +58,10 @@ def topbar_props(community:, user: nil, community_locales:, path_after_locale_ch
           imageHeight: '44px',
           image: Maybe(user).image.url(:thumb).or_else(missing_profile_image_path()),
         }
+      },
+      newListingButton: {
+        text: I18n.t("homepage.index.post_new_listing"),
+        customColor: CommonStylesHelper.marketplace_colors(community)[:marketplace_color1]
```
rap1ds: This is camelCase, event though other props are snake_case. What should it be?

kpuputti: Good point. It seems that we have both conventions already before this, and I just copied the case style from the avatarDropdown props above. I'll add a separate task to clean these up outside this PR.

PR #2171, opened 2016-06-23
Fallback custom colors from variables.js

mporkola:


client/app/assets/styles/variables.js
```diff
@@ -158,4 +161,8 @@ module.exports = {
   // topbar can't control base font-size.
   '--Menu_labelPaddingVertical': '1.28571429em',
   '--Menu_labelPaddingHorizontal': '1.714285em',
+
+  // plain values
+  customColorFallback,
+  customColor2Fallback,
```
kpuputti: These fallback colors seem like something that will be used straight in the CSS as base values before being overriden with marketplace colors. So why would they be named differently that other CSS variables, in a way that prevents using them in CSS files?

PR #2172, opened 2016-06-27
Google Maps API key

kpuputti:
This PR adds a column to the communities table for a Google Maps API key. This key is then included in the script tab URLs that load the Maps API.

### DIFF
--- a/app/views/layouts/application.haml
+++ b/app/views/layouts/application.haml
@@ -18,7 +18,9 @@
 
   - if feature_enabled?(:topbar_v1)
     - props = topbar_props
-    = javascript_include_tag "https://maps.googleapis.com/maps/api/js?key=#{@current_community.google_maps_key}&libraries=places"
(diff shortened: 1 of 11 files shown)


app/views/layouts/application.haml
```diff
@@ -18,7 +18,9 @@
 
   - if feature_enabled?(:topbar_v1)
     - props = topbar_props
-    = javascript_include_tag "https://maps.googleapis.com/maps/api/js?libraries=places"
+    - maps_key = @current_community.google_maps_key
+    - key_param = maps_key ? "&key=#{maps_key}" : ""
+    = javascript_include_tag "https://maps.googleapis.com/maps/api/js?libraries=places#{key_param}".html_safe
```
rap1ds: Was the `html_safe` needed or did you just forget here when we were testing it?

kpuputti: Just noticed the same, forgot to add those. I'll do it now.

PR #2173, opened 2016-06-27
Topbar layout

mporkola:
- spacer for the empty part between menu and language selector
- move positioning more to the responsibility of container component, for better component reusability
- give maximum width for text logo

### DIFF
--- a/client/app/components/sections/Topbar/Topbar.css
+++ b/client/app/components/sections/Topbar/Topbar.css
@@ -13,6 +13,7 @@
 
   @media (--Topbar_medium) {
     height: var(--Topbar_height);
+    padding: var(--Topbar_padding);
   }
 
   /* Base styles */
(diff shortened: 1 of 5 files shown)


client/app/components/sections/Topbar/Topbar.css
```diff
@@ -52,10 +54,19 @@
     font-size: var(--Menu_fontSizeSmall);
   }
 
+  & > * {
+    margin-right: 24px;
```
kpuputti: This `24px` is repeated in several places and could be a variable to convey its meaning.

client/app/components/sections/Topbar/Topbar.css
```diff
@@ -52,10 +54,19 @@
...
+  }
+
+}
+
+.topbarSpacer {
+  flex-grow: 100;
```
kpuputti: What is this spacer and what does the `flex-grow` value do?

mporkola: It's to create the spacing between the menu and language dropdown and push the items to left and right sides, like the design for desktop shows. `flex-grow` seems to be a bit ambiguous number, didn't find any definition for that but `30` seemed to be good value

client/app/components/sections/Topbar/Topbar.css
```diff
@@ -52,10 +53,26 @@
...
+    &:last-child {
+      margin-right: 0;
+    }
+  }
+
+}
+
+.topbarSpacer {
+  display: none;
+  @media (--Topbar_medium) {
```
kpuputti: This will require a newline before the media query for Stylelint (might be hard to see with the menu errors atm).

PR #2180, opened 2016-06-28
Integrate topbar with landing page

ovan:
Integrates the current version of the topbar component on landing page behind topbar_v1 feature flag. This version uses per marketplace data for topbar so everything can be cached. Another PR will later add dynamic fetching of per user properties and re-rendering topbar with them outside the caching cycle.


app/assets/stylesheets/landing_page/topbar.scss
```diff
@@ -0,0 +1,13 @@
...
+  height: 50px;
+
+  @media #{$topbar_tablet} {
+    height: 50px;
+  }
+
+  @media #{$topbar_desktop} {
+    height: 80px;
+  }
+}
```
rap1ds: new line missing

app/controllers/landing_page_controller.rb
```diff
@@ -203,20 +220,48 @@ def render_landing_page(community_id:, default_locale:, locale_param:, structure
...
+    TopbarHelper.topbar_props(
+      community: community,
+      path_after_locale_change: path,
+      search_placeholder: community_customization&.search_placeholder,
+      locale_param: locale_param)
+  end
+
+  def fetch_topbar_enabled(community_id)
+    flags_res = FeatureFlagService::API::Api.features.get(community_id: community_id)
+    flags_res.maybe
```
rap1ds: Remember, `result.maybe` was kind of a "deprecated" method, because for errored results it returns `None`.

app/controllers/landing_page_controller.rb
```diff
@@ -203,20 +220,48 @@ def render_landing_page(community_id:, default_locale:, locale_param:, structure
...
   end
 
   def render_not_found(msg = "Not found")
     self.status = 404
     self.response_body = msg
   end
 
+  def topbar_props(community, community_customization, request_path, locale_param, default_locale)
+    # TopbarHelper pulls current lang from I18n
+    I18n.locale = locale_param.present? ? locale_param : default_locale
```
rap1ds: `js_translations` method has this same logic. Would it make sense to move this logic a level up (to `render_landing_page` method) and name the result to `topbar_locale`. And then name the `locale` from structure to `landing_page_locale`. To me those names sound better and easier to understand than figuring out what does it mean if there is `locale_param` or `default_locale` or `locale`.

app/views/landing_page/landing_page.erb
```diff
@@ -33,6 +33,12 @@
 
   <title><%= page["title"]["value"] %></title>
 
+  <script>
+    window.I18n = {};
+    <%= javascripts[:translations] %>
+  </script>
```
rap1ds: I would move this `script` tag to line 141 right before including `application.js`. It doesn't need to be in head and it will slow down the rendering (not much) because the script tag content needs to be read and evaluated.

client/app/startup/clientRegistration.js
```diff
@@ -14,3 +16,8 @@ ReactOnRails.register({
 
 ReactOnRails.registerStore({
 });
+
+if (window) {
```
rap1ds: Does this really work if `window` is not defined? My hunch is that this will throw an error "window is not defined" or something like that.

To my understanding, the right way to check this is:

```
if(typeof window !== "undefined") {
  ...
}
```

app/controllers/landing_page_controller.rb
ovan: What's the correct way then?

PR #2181, opened 2016-06-28
Landing page search path

rap1ds:
- [X] Use `PathHelpers`
- [X] Add `all_categories` path

### DIFF
--- a/app/controllers/landing_page_controller.rb
+++ b/app/controllers/landing_page_controller.rb
@@ -30,7 +30,7 @@ class LandingPageController < ActionController::Metal
 
     begin
       content = nil
-      cache_meta = fetch_cache_meta(cid, locale_param, version)
+      cache_meta = fetch_cache_meta(cid, version, locale_param)
       cache_hit = true
 
       if cache_meta.nil?
--- a/app/controllers/landing_page_controller.rb
+++ b/app/controllers/landing_page_controller.rb
@@ -95,7 +95,7 @@ class LandingPageController < ActionController::Metal
 
   def build_html(community_id, default_locale, version)
     structure = CLP::LandingPageStore.load_structure(community_id, version)
-    render_landing_page(community_id, default_locale, structure, true)
+    render_landing_page(community_id, default_locale, structure)
   end
 
   def build_cache_meta(content)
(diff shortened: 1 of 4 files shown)


app/controllers/landing_page_controller.rb
```diff
@@ -118,18 +118,20 @@ def write_cached_content!(community_id, version, content, digest, cache_time)
...
-    end
-  end
-
-  def build_denormalizer(cid:, locale:, sitename:, lp_enabled:)
+  def build_denormalizer(cid:, default_locale:, locale:, sitename:)
+    search_path = ->(opts = {}) {
+      PathHelpers.search_path(
+        community_id: cid,
+        user: nil,
+        locale_param: locale,
```
ovan: This should be `params[:locale]`

app/view_utils/topbar_helper.rb
```diff
@@ -38,7 +43,12 @@ def topbar_props(community:, user: nil, community_locales:, path_after_locale_ch
 
     {
       logo: {
-        href: PathHelpers.landing_page_path(community, user, locale_param),
+        href: PathHelpers.landing_page_path(
+          community_id: community,
```
ovan: community.id

app/view_utils/path_helpers.rb
```diff
@@ -2,16 +2,17 @@ module PathHelpers
 
   module_function
 
-  def search_path(community, user, locale_param, opts = {})
+  def search_path(community_id:, user:, locale_param:, default_locale:, opts: {})
```
ovan: This gets rid of the community model but user model is still passed as model. Could it be in the scope to change that too.

app/controllers/landing_page_controller.rb
```diff
@@ -51,7 +58,7 @@ def index
         if content.nil?
           # This should not happen since html is cached longer than metadata
           cache_hit = false
-          content = build_html(cid, version)
+          content = build_html(cid, default_locale, version)
```
ovan: This is not called with keyword params

app/controllers/landing_page_controller.rb
```diff
@@ -24,20 +24,27 @@ class LandingPageController < ActionController::Metal
...
+          community_id: cid,
+          default_locale: default_locale,
+          locale_param: locale_param,
+          version: version
+        )
         cache_meta = build_cache_meta(content)
 
         # write metadata first, so that it expires first
-        write_cache_meta!(cid, version, cache_meta, CACHE_TIME)
+        write_cache_meta!(cid, version, locale, cache_meta, CACHE_TIME)
```
ovan: s/locale/locale_param. There's no locale in the scope here.

app/controllers/landing_page_controller.rb
ovan: Also, this is called with default_locale instead of locale_param

PR #2182, opened 2016-06-28
Optimize checking landing page status

bladealslayer:
Big cover photo is only relevant for HomepageController#index.


rap1ds:
Code looks good to me. Remember to test.

PR #2183, opened 2016-06-28
Google Analytics and Tag Manager for landing pages

bladealslayer:


app/controllers/landing_page_controller.rb
```diff
@@ -183,12 +191,8 @@ def parse_int(int_str_or_nil)
...
   end
 
-  def community_id(request)
-    request.env[:current_marketplace]&.id
-  end
-
-  def community_default_locale(request)
-    request.env[:current_marketplace]&.default_locale
+  def current_community(request)
+    @current_community ||= request.env[:current_marketplace]
```
ovan: I guess this change is because layouts/google_analytics_script relies on global `@current_community`? That's a style we haven't been using for a while because it's just a really really bad idea for code structuring. How far off would we be from reworking  the partial to take the community as a local? Or even better take the data it needs from the community object as a hash?

bladealslayer: Great point, I did this to make it work quickly :) I can dig in to see how much data is needed from `@current_community`.

ovan: In any case, it would be a good idea to merge these two in first and then continue with the extracting as a separate PR so we can ensure we have something ready to go.

PR #2184, opened 2016-06-28
Topbar layout for different screen sizes

mporkola:


client/app/components/sections/Topbar/Topbar.css
```diff
@@ -102,10 +123,10 @@
   font-family: var(--Topbar_fontFamily);
   font-size: var(--Topbar_fontSize);
   align-self: flex-start;
-  display: none;
+  display: none !important;
```
mporkola: this needs some specificity-fu, suggestions welcome!

client/app/components/composites/SearchBar/SearchBar.css
```diff
@@ -1,5 +1,5 @@
-@custom-media --SearchBar_small (max-width: 40rem);
-@custom-media --SearchBar_medium (min-width: 40rem);
+@custom-media --SearchBar_small (max-width: 62rem);
+@custom-media --SearchBar_medium (min-width: 62rem);
```
kpuputti: I think there should be generic Topbar break points that all the components within it are using. These media queries were mean't to be temporary (sorry for not documenting that in the code).

As for using rems, the breakpoints will behave differently when the Topbar is used within the main app, custom landing pages, and the Styleguide. However, ems are problematic as well since the font size changes in the breakpoints. Should we just use pixels in the breakpoints?

client/app/components/elements/AddNewListingButton/AddNewListingButton.js
```diff
@@ -1,17 +1,17 @@
...
 import { a, span } from 'r-dom';
+import classNames from 'classnames';
+
+import { className } from '../../../utils/PropTypes';
 
 import * as variables from '../../../assets/styles/variables';
 import css from './AddNewListingButton.css';
 
-export default function AddNewListingButton({ text, url, customColor }) {
+export default function AddNewListingButton({ text, url, customColor, className }) {
```
kpuputti: `className` shadows the imported prop type (linting fails here as well).

client/app/components/sections/Topbar/Topbar.css
```diff
@@ -1,5 +1,5 @@
-@custom-media --Topbar_small (min-width: 30rem);
-@custom-media --Topbar_medium (min-width: 50rem);
+@custom-media --Topbar_small (min-width: 38em);
+@custom-media --Topbar_medium (min-width: 62em);
```
kpuputti: These should be defined in `client/app/assets/styles/media-queries.js` if they are also used in components within Topbar e.g. to switch between mobile and tablet/desktop layouts.

client/app/components/sections/Topbar/Topbar.css
```diff
@@ -71,7 +71,13 @@
   }
 
   & > * {
-    margin-right: var(--Topbar_itemSpacing);
+    @media (--Topbar_small) {
+      margin-right: 18px;
```
kpuputti: This could be a var `--Topbar_mobileItemSpacing`?

client/app/components/sections/Topbar/Topbar.css
kpuputti: Where is the conflicting display value defined?

client/app/assets/styles/media-queries.js
```diff
@@ -10,4 +10,5 @@ module.exports = {
   // so we must add those manually to the output.
   '--medium-viewport': `(min-width: ${medium})`,
   '--large-viewport': `(min-width: ${large})`,
+  '--small-and-medium-viewport': `(max-width: ${large})`,
```
kpuputti: I think both the `--large-viewport` and `--small-and-medium-viewport` queries apply when the viewport is exactly `large`, which is probably not what we want?

PR #2185, opened 2016-06-28
CLP topbar with per user data

ovan:


client/app/startup/clientRegistration.js
```diff
@@ -14,3 +16,8 @@ ReactOnRails.register({
 
 ReactOnRails.registerStore({
 });
+
+if (window) {
```
rap1ds: Does this really work if `window` is not defined? My hunch is that this will throw an error "window is not defined" or something like that.

To my understanding, the right way to check this is:

```
if(typeof window !== "undefined") {
  ...
}
```

app/views/landing_page/landing_page.erb
```diff
@@ -112,20 +114,37 @@
 </head>
 <body>
 
-<%= yield :sections %>
+  <div id="topbar-placeholder" style="height: 80px"> </div>
```
rap1ds: The `style` attribute should be removed?

app/controllers/topbar_api_controller.rb
```diff
@@ -0,0 +1,33 @@
+# coding: utf-8
+class TopbarApiController < ApplicationController
+
+  attr_reader :community_customization
```
rap1ds: Do I understand this right: ApplicationController has instance variable `@community_customization` and now you're adding a getter method to it in `TopbarApiController`? If yes, then why not add the getter to `ApplicationController`? Or maybe you could at least add a comment here that the instance variable is in `ApplicationController`. It's a bit confusing when there's the `attr_reader` but no instance variable in this class, so adding a comment which explains where the instance variable actually is would help understanding this code.

app/controllers/landing_page_controller.rb
```diff
@@ -175,20 +188,30 @@ def render_landing_page(cid, structure, lp_enabled)
...
 
   def render_not_found(msg = "Not found")
     self.status = 404
     self.response_body = msg
   end
 
+  def topbar_props(community, community_customization, request_path)
+    TopbarHelper.topbar_props(
+      community: community,
+      path_after_locale_change: request_path.gsub(/^\/en/, "").gsub(/^\//, ""),
```
rap1ds: Out of scope of this PR, but shouldn't the front-end figure out this itself? In the future, when we add routing to front-end, this won't work anymore.

app/controllers/topbar_api_controller.rb
rap1ds: In addition, this seems to be inconsistent with `community` and `user`.

PR #2188, opened 2016-06-29
Rename railsContext

kpuputti:
We need a marketplace wide context object with data that is shared within all section level components (e.g. TopbarApp), but it's weird if its called railsContext when components are rendered sometimes without any actual Rails context to speak of, or in Styleguide.

This PR renames the context object to `marketplaceContext` and makes the prop validation only require props that are currently used in the components.

Some missing props were also added to the Styleguide stories.


client/app/startup/OnboardingGuideApp.js
```diff
@@ -28,10 +28,10 @@ export default (props, railsContext) => {
     'edit_admin_listing_shape',
     'new_invitation',
     'new_listing',
-  ], { locale: railsContext.i18nLocale });
+  ], { locale: marketplaceContext.i18nLocale });
 
   const combinedReducer = combineReducers(reducers);
-  const combinedProps = Object.assign({}, props, { railsContext, routes });
```
kpuputti: RailsContextReducer was removed (unused), so the store shouldn't get the context object anymore.

rap1ds:
LGTM

PR #2189, opened 2016-06-29
Set Google Maps API key for CLP

bladealslayer:
This builds on top of #2183. Please review that PR first.

### DIFF
--- a/app/controllers/landing_page_controller.rb
+++ b/app/controllers/landing_page_controller.rb
@@ -228,6 +228,8 @@ class LandingPageController < ActionController::Metal
     marketplace_context = marketplace_context(community_id, topbar_locale, request)
     topbar_enabled = fetch_topbar_enabled(community_id)
 
+    google_maps_key = community(request).google_maps_key
+
     denormalizer = build_denormalizer(
       cid: community_id,
       locale_param: locale_param,
@@ -248,6 +250,7 @@ class LandingPageController < ActionController::Metal
                      topbar_props_path: ui_api_topbar_props_path(locale: topbar_locale),
                      marketplace_context: marketplace_context,
                      topbar_enabled: topbar_enabled,
+                     google_maps_key: google_maps_key,
                      community_context: community_context(request)
                    }
   end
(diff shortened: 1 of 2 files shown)


app/controllers/landing_page_controller.rb
```diff
@@ -36,7 +42,7 @@ def index
       if cache_meta.nil?
         cache_hit = false
         content = build_html(
-          community_id: cid,
+          community: community,
```
ovan: We're trying to get rid of community etc. objects in 99,99% of the code. That's why in this controller community is only handled at the top level actions and from there on data is passed down as data. This change goes against that idea so I'd vote for passing api key directly.

app/controllers/landing_page_controller.rb
```diff
@@ -19,12 +19,18 @@ class LandingPageController < ActionController::Metal
   # Adds helper_method
   include ActionController::Helpers
 
+  include FeatureFlagHelper
```
ovan: This is a good change. You should also re-implement the way the topbar flag is checked based upon this so we avoid the extra DB call.

app/controllers/landing_page_controller.rb
bladealslayer: For API key your suggestion is easy to do. However this builds on top of Analytics and the analytics partials assume they have access to `@current_community`. So to make those work it would require mode work then.

app/controllers/landing_page_controller.rb
bladealslayer: I guess that's something that got merged in the meantime? I'll check.

app/views/landing_page/landing_page.erb
```diff
@@ -35,6 +35,10 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
 
+  <% if render_analytics %>
```
ovan: This is quickly approaching nitpicking but I assume the only purpose for the `render_analytics` local is to make preview work at editor side where the partial is not available? This could be accomplished with slighty less noise by having an opposite boolean, `skip_analytics` that doesn't need to be passed in at all for default behavior (== render analytics). You can test for local existing with: `local_assigns.has_key?(:skip_analytics)`

bladealslayer: Correct, it's for the editor. I didn't know how to test existence of the local, thanks!

PR #2190, opened 2016-06-29
Fix HTTPS redirects

bladealslayer:
- [x] fix redirect tests not running
- [x] fix http->https redirects in Rack middleware


bladealslayer:
Hmm, spec failed on travis, but works locally. Checking.

lib/rack_middleware/enforce_ssl.rb
```diff
@@ -4,10 +4,9 @@ def initialize(app)
   end
 
   def call(env)
-    if ::APP_CONFIG.always_use_ssl.to_s == "true" && env["HTTPS"] == "off" # always_use_ssl can be string if it comes from ENV
-      req = ::Rack::Request.new(env)
-      port_s = req.port ? ":#{req.port}" : ""
-      redirect("https://#{req.host}#{port_s}#{req.fullpath}")
+    req = ::Rack::Request.new(env)
+    if ::APP_CONFIG.always_use_ssl.to_s == "true" && !req.ssl? # always_use_ssl can be string if it comes from ENV
+      redirect("https://#{req.host}#{req.fullpath}")
```
ovan: Why is the port handling not something we want to do here?

bladealslayer: HTTP port and HTTPS ports are different. We have no idea which is the correct HTTPS port, if non-standard, so we assume it's the default. The old code resulted in redirects to `https://domain:80`, which is definitely wrong.

ovan: Yes, of course. Makes perfect sense. Go ahead and merge!

PR #2192, opened 2016-06-29
Fix avatar props

mporkola:
Couple of simple fixes to:
- silence invalid `Proptype`-errors
- fix empty Avatar image

### DIFF
--- a/app/view_utils/topbar_helper.rb
+++ b/app/view_utils/topbar_helper.rb
@@ -66,7 +66,7 @@ module TopbarHelper
         customColor: CommonStylesHelper.marketplace_colors(community)[:marketplace_color1],
         avatar: {
           imageHeight: '44px',
-          image: Maybe(user).image.url(:thumb).or_else(missing_profile_image_path()),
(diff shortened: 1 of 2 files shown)


app/view_utils/topbar_helper.rb
```diff
@@ -66,7 +66,7 @@ def topbar_props(community:, path_after_locale_change:, user: nil, search_placeh
         customColor: CommonStylesHelper.marketplace_colors(community)[:marketplace_color1],
         avatar: {
           imageHeight: '44px',
-          image: Maybe(user).image.url(:thumb).or_else(missing_profile_image_path()),
+          image: user&.image.present? ? user.image.url(:thumb) : missing_profile_image_path(),
```
kpuputti: Shouldn't there be another safeguard if the `user` is nil?

```
> (nil)&.image.present?
NoMethodError: undefined method `present?' for nil:NilClass

> (nil)&.image&.present?
 => nil
```

mporkola: `nil` gets `.present?` method from ActiveSupport, so this works in our contexts

PR #2213, opened 2016-07-01
Fix Context handling for landing page topbar

ovan:
* Detect logged in user and pass it correctly to topbar
* Add timeout and fallback in case topbar api doesn't respond (fast enough)

### DIFF
--- a/app/views/landing_page/landing_page.erb
+++ b/app/views/landing_page/landing_page.erb
@@ -171,25 +171,35 @@
   <%= javascript_include_tag 'app-bundle' %>
 
   <script>
-    var renderTopBar = function renderTopBar(props, marketplaceCtx) {
-      componentFn = function (props) {
-        return window.ReactOnRails.getComponent('TopbarApp').component(props, marketplaceCtx)
-      }
-      ReactDOM.render(React.createElement(componentFn, props), document.getElementById('topbar-placeholder'));
-    };
+    (function() {
+      var renderTopBar = function renderTopBar(props, marketplaceCtx) {
+        componentFn = function (props) {
+          return window.ReactOnRails.getComponent('TopbarApp').component(props, marketplaceCtx)
+        }
+        ReactDOM.render(React.createElement(componentFn, props), document.getElementById('topbar-placeholder'));
+      };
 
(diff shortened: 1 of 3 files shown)


app/views/landing_page/landing_page.erb
```diff
@@ -178,14 +178,25 @@
...
-    static_props = <%= topbar_props.to_json.html_safe %>
-    marketplaceContext = <%= marketplace_context.to_json.html_safe %>
-
-    renderTopBar(static_props, marketplaceContext);
-
-    window.fetch("<%= topbar_props_path %>", { credentials: 'same-origin' })
-      .then(function (res) { return res.json(); })
-      .then(function (props) { renderTopBar(Object.assign({}, static_props, props), marketplaceContext); });
+    staticProps = <%= topbar_props.to_json.html_safe %>
+    staticMarketplaceContext = <%= marketplace_context.to_json.html_safe %>
```
kpuputti: Add `var` to both these variables.

app/views/landing_page/landing_page.erb
```diff
@@ -178,14 +178,25 @@
...
+
+    p.then(function (resJson) {
+      props = Object.assign({}, staticProps, resJson.props);
+      mCtx = Object.assign({}, staticMarketplaceContext, resJson.marketplaceContext)
+      renderTopBar(props, mCtx);
+    });
+
+    p.catch(function(error) {
+      console.log(error);
+      renderTopBar(staticProps, staticMarketplaceContext);
```
kpuputti: You could use `console.error` for a more visible error in the console.

These two statements could also be rearranged just to make sure the `renderTopBar` is called if the console call fails (not all browsers have it).

app/views/landing_page/landing_page.erb
```diff
@@ -178,14 +178,25 @@
...
+    });
   </script>
```
kpuputti: You are creating multiple global variables, e.g. `p` that pollute all the JS scope. You could wrap the whole script tag in an IIFE:

```html
<script>
  (function () {
    // ...
  }());
</script>
```

This creates a scope around the code.

PR #2214, opened 2016-07-01
Redirect to marketplace domain stripping prefix

bladealslayer:
Avoid issue when e.g. http://www.IDENT.example.com gets redirected to https://www.IDENT.example.com
and there is wildcard SSL certificate covering *.example.com.

Extend test coverage for redirects.

This should not affect any marketplace using a custom domain.

### DIFF
--- a/spec/requests/https_redirects_spec.rb
+++ b/spec/requests/https_redirects_spec.rb
@@ -1,17 +1,14 @@
 require "spec_helper"
 
 describe "Redirect to HTTPS", type: :request do
-  before(:each) do
-    @domain = "market.custom.org"
-    @http_url = "http://#{@domain}"
-    @https_url = "https://#{@domain}"
-    @community = FactoryGirl.create(:community, :domain => @domain, use_domain: true)
 
-    # Refresh from DB
-    @community.reload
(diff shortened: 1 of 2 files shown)


spec/requests/https_redirects_spec.rb
```diff
@@ -20,32 +24,95 @@
...
+    after(:each) do
+      APP_CONFIG.domain = @orig_domain
+    end
+
+    context "when always_use_ssl is true:" do
+      include_context "common with ssl"
+
+      it "redirects and strips additional labels over HTTP" do
+        expect_redirect(www_http_url, https_url)
+        expect_redirect(sub_http_url, https_url)
```
rap1ds: The use of `let` variables makes this test a bit hard to read and reason about. By looking at the 2 lines here, I can't fully understand what's happening, without checking the value of these variables. And the definition of these variables are a bit far away. In addition, mixing the lazy-loaded and eager-loaded `let`s and relying on that order requires some extra brain power to understand what values will be used.

I think just hard-coding the URLs here would make this test easier to reason about, but of course, it adds some repetition:

```
expect_redirect("http://something.else.market.example.com", "https://market.example.com")
```

PR #2217, opened 2016-07-01
Font name

rap1ds:
Change font name to `Proxima Nova Soft`


mporkola:
can't comment on your grepping accuracy, but I trust you were successful 👍

PR #2227, opened 2016-07-03
Landing page layout fixes

rap1ds:
- [X] Add zebra skin to every second consecutive info section without background
- [X] Remove double margins from info section columns
- [X] Add max-width 75% and margin to section titles and paragraphs
- [X] Add support for line breaks in paragraphs


app/controllers/landing_page_controller.rb
```diff
@@ -852,7 +852,7 @@ def data_str
       "columns": [
         {
           "title": "Our mission",
-          "paragraph": "Paragraph. Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis vestibulum. Vivamus sagittis lacus vel augue laoreet rutrum faucibus dolor auctor. Donec id elit non mi porta gravida at eget metus."
+          "paragraphs": ["Paragraph. Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis vestibulum. Vivamus sagittis lacus vel augue laoreet rutrum faucibus dolor auctor. Donec id elit non mi porta gravida at eget metus.","Paragraph. Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis vestibulum. Vivamus sagittis lacus vel augue laoreet rutrum faucibus dolor auctor. Donec id elit non mi porta gravida at eget metus.","Paragraph. Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis vestibulum. Vivamus sagittis lacus vel augue laoreet rutrum faucibus dolor auctor. Donec id elit non mi porta gravida at eget metus."]
```
ovan: @rap1ds The other option to renaming the key would have been to start supporting both strings and arrays as value types. That might be more friendly for entering the input (I can see it's easy to forget the [] when entering only one paragraph) and would require no migration from old key to new key. Any thoughts?

app/views/landing_page/landing_page.erb
```diff
@@ -18,7 +18,23 @@
...
+            if second_in_row
+              [consecutive_infos.add(sect["id"]), false]
+            else
+              [consecutive_infos, true]
+            end
+          else
+            [consecutive_infos, false]
+          end
+        }.first.include?(s["id"])
+      %>
```
ovan: I don't have a better suggestion but this feels like a lot of logic in erb. :/

PR #2228, opened 2016-07-04
Hide admin dashboard link for non admins

mporkola:


app/view_utils/topbar_helper.rb
```diff
@@ -66,7 +66,8 @@ def topbar_props(community:, path_after_locale_change:, user: nil, search_placeh
         customColor: CommonStylesHelper.marketplace_colors(community)[:marketplace_color1],
         avatar: {
           image: user&.image.present? ? user.image.url(:thumb) : missing_profile_image_path(),
-        }
+        },
+        isAdmin: user&.has_admin_rights? || false
```
mporkola: not sure if this is the Ruby way to convert `nil` to `false`, but didn't feel too evil to me

PR #2235, opened 2016-07-04
iOS7 doesn't support flexbox without prefixes

Gnito:


rap1ds:
@Gnito Might be irrelevant but my iPad that I used to test was iOS 8.1 (http://www.whatsmybrowser.org/b/K2EV5DV)

mporkola:
Functionally 👍, but let's discuss browser support first

Gnito:
If I'll change this to iOS 8++?

PR #2237, opened 2016-07-05
Arrow dropdown arrow

mporkola:
Before 
screenshot 2016-07-05 10 36 21

After
screenshot 2016-07-05 10 36 06

I'll squash this, so the commit names are only to make review easier


rap1ds:
The code looks good to me but maybe someone else from the topbar team could have a look at this also.

PR #2238, opened 2016-07-05
Consistent icon sizing

rap1ds:
- [x] Make icon height always 60px. That way they are always aligned
- [x] Replace the watering can icon with 60px version

Screenshots:

screen shot 2016-07-05 at 10 01 53
screen shot 2016-07-05 at 10 01 03


app/controllers/landing_page_controller.rb
```diff
@@ -672,14 +672,14 @@ def data_str
...
-          "icon": "binocular",
+          "icon": "grape",
           "title": "Our mission",
           "paragraph": ["Paragraph. Curabitur blandit tempus porttitor. Nulla vitae elit libero, a pharetra augue. Vivamus sagittis lacus vel.","Paragraph. Curabitur blandit tempus porttitor. Nulla vitae elit libero, a pharetra augue. Vivamus sagittis lacus vel."],
           "button_title": "Section link",
           "button_path": {"value": "https://google.com"}
         },
         {
-          "icon": "piggy-bank",
+          "icon": "watering-can",
```
mporkola: Are these changes included on purpose? Just asking because neither PR or commit description mentions these

mporkola:
👍 if the answer to my comment is 👍

app/controllers/landing_page_controller.rb
rap1ds: Yep, on purpose, since this is actually test data only.

PR #2240, opened 2016-07-05
Cache access to url_helpers module in view utils

ovan:
Rails.application.routes.url_helpers is ~500 times slower than just
Rails.application.routes. This is because .url_helpers dynamically
builds an involved module on every invocation. This module is intended
to be mixed in to classes in which case it gets cached as part of the
Class definition. When accessed directly in other modules it must always
be cached to avoid the costly reconstruction on every call.

This will still work with Rails autoloading because upon detecting a change Rails will autoload all classes and not just Rails specific classes: http://guides.rubyonrails.org/autoloading_and_reloading_constants.html#constant-reloading.

### DIFF
--- a/app/view_utils/path_helpers.rb
+++ b/app/view_utils/path_helpers.rb
@@ -56,7 +56,7 @@ module PathHelpers
   end
 
   def paths
-    RequestStore.store[:url_helpers] ||= Rails.application.routes.url_helpers
+    @_url_helpers ||= Rails.application.routes.url_helpers
   end
 
 end
--- a/app/services/custom_landing_page/category_store.rb
+++ b/app/services/custom_landing_page/category_store.rb
@@ -25,7 +25,7 @@ module CustomLandingPage
     # private
 
     def paths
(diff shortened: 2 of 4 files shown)


app/view_utils/path_helpers.rb
```diff
@@ -56,7 +56,7 @@ def landing_page_path(community_id:, logged_in:, locale_param:, default_locale:)
   end
 
   def paths
-    RequestStore.store[:url_helpers] ||= Rails.application.routes.url_helpers
+    @_url_helpers ||= Rails.application.routes.url_helpers
```
bladealslayer: `@_url_helpers` in `path_helpers` won't be scoped to a request. Is this as intended?

ovan: Yes, it doesn't have to, there's nothing request specific in it. And the autoloading in dev mode still works as intended because of how Rails autoloads all code when it detects a change e.g. in routes.rb.

PR #2243, opened 2016-07-05
Use marketplace custom color for text logo

mporkola:


client/app/components/Styleguide/withProps.js
```diff
@@ -9,7 +9,7 @@ const defaultRailsContext = {
   pathname: '/',
   marketplaceId: 1,
 
-  marketplace_color1: '#ee4',
+  marketplace_color1: '#64A',
```
mporkola: not related, just wanted a nicer color

PR #2248, opened 2016-07-06
Fix location search derivatives

Gnito:
- Clear derivative hidden fields if search string changes
- Make query prediction on submit, if it has not been done already for the current search string


app/assets/javascripts/location_search.js
```diff
@@ -77,7 +86,8 @@ window.ST = window.ST || {};
     // Ensure default events don't fire without correct info
     form.addEventListener('submit', function(e) {
       // If service status is unset and there are no coordinates, do not make search submit
-      if(statusInput.value === "" && coordinateInput.value === "" && searchInput.value !== "") {
+      //if(statusInput.value === "" && coordinateInput.value === "" && searchInput.value !== "") {
```
mporkola: rm this

PR #2250, opened 2016-07-06
Enable onboarding wizard for everybody and remove the feature flag

mporkola:
```bash
~/p/sharetribe git:enable-onboarding-wizard ❯❯❯ git grep "onboarding_redesign_v1" | wc -l
       0
```

### DIFF
--- a/app/controllers/admin/communities_controller.rb
+++ b/app/controllers/admin/communities_controller.rb
@@ -9,11 +9,7 @@ class Admin::CommunitiesController < ApplicationController
     @selected_left_navi_link = "getting_started"
     @community = @current_community
 
-    if(feature_enabled?(:onboarding_redesign_v1))
-      redirect_to admin_getting_started_guide_path
-    else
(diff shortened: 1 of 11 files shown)


app/controllers/admin/communities_controller.rb
```diff
@@ -9,11 +9,7 @@ def getting_started
     @selected_left_navi_link = "getting_started"
     @community = @current_community
 
-    if(feature_enabled?(:onboarding_redesign_v1))
-      redirect_to admin_getting_started_guide_path
-    else
-      render locals: {paypal_enabled: PaypalHelper.paypal_active?(@current_community.id)}
-    end
+    redirect_to admin_getting_started_guide_path
```
Gnito: This should be done in the left navigation bar

rap1ds: We could now remove this whole action and also the related `getting_started.haml` (and all the related translations)

rap1ds: That can be also done in a separate PR if we just want to quickly enable onboarding for everyone, but anyhow it's good to do it.

PR #2252, opened 2016-07-07
Onboarding cleanup

mporkola:
Cleaning up after https://github.com/sharetribe/sharetribe/pull/2250


config/routes.rb
```diff
@@ -195,6 +194,11 @@
           put :menu_links, to: 'communities#update_menu_links'
           delete :delete_marketplace
 
+          # DEPRECATED (2016-07-07)
+          # These routes are not in use anymore, don't use them
+          # See new "Guide" routes above, outside of communities resource
+          get :getting_started, to: 'getting_started_guide#index'
```
rap1ds: 👍  for not just removing the route

rap1ds: I guess redirect would be more appropriate than routing the old URL to new controller?

rap1ds:
One comment about route redirect, otherwise looks good to me! 👍 

Remember to do some testing in staging. Deleting old code may always break things if there are some pages that is still pointing to the old code/controller/action/whatever

PR #2256, opened 2016-07-07
Add SEO tags to landing pages

bladealslayer:


app/views/landing_page/landing_page.erb
```diff
@@ -57,6 +57,10 @@
 
   <title><%= page["title"]["value"] %></title>
 
+  <% unless local_assigns.has_key?(:skip_meta) && skip_meta %>
```
ovan: Is there any reason to have the option to skip meta? This part depends heavily on the structure json so it would help to have the editor preview render it too to catch problems in that part of the json. Providing fake data to resolver shouldn't be a problem in editor, we do that already, just need to extend the fake provider.

app/controllers/landing_page_controller.rb
```diff
@@ -242,7 +245,7 @@ def render_landing_page(default_locale:, locale_param:, structure:)
                      marketplace_context: marketplace_context,
                      topbar_enabled: topbar_enabled,
                      google_maps_key: google_maps_key,
-                     community_context: community_context(request)
+                     community_context: community_context(request, locale)
```
ovan: We have some confusion with the different locales here. The `render_landing_page` has 3 local locale variables in scope: `default_locale`, `locale_param` and `landing_page_locale`. There is no `locale`. The reason this works is some Rails stuff that a magic locale method enters the scope via controller inheritance and of course it's an awesome idea that method calls don't need () so the language doesn't get in the way of a programmer mistaking a locale variable with super class method call </sarcasm>. What this returns? I have no idea. But it's very likely you should be using one of the three instead of this. There's another place in code where `locale` is called on line 212 `initialize_i18n!(c&.id, locale)`. Perhaps @rap1ds you have an idea which locale should really be used there?

PR #2262, opened 2016-07-07
Add writer methods to LandingPageStore and add integration tests for CLP

bladealslayer:
- [x] Add methods for writing data to LandingPageStore
- [x] Write integration tests

### DIFF
--- a/spec/requests/landing_pages_spec.rb
+++ b/spec/requests/landing_pages_spec.rb
@@ -1,0 +1,230 @@
+require "spec_helper"
+
+def create_version(community_id, structure, version_number)
+  s = structure.deep_dup
+  s["page"]["title"] = {value: "Title version #{version_number}"}
+  CustomLandingPage::LandingPageStore.create_version!(community_id, version_number, JSON.generate(s))
+end
+
+def release_version(community_id, version_number)
+  CustomLandingPage::LandingPageStore.release_version!(community_id, version_number)
+end
+
+# Populate category ids in all category sections
+def populate_categories(structure, category_ids)
+  struct = structure.deep_dup
+  struct["sections"].each do |s|
+    next unless s["kind"] == "categories"
+
+    s["categories"].each_with_index do |c, i|
+      c["category"]["id"] = category_ids[i]
+    end
+  end
+  struct
+end
+
+# Populate listing ids in all listing sections
+def populate_listings(structure, listing_ids)
+  struct = structure.deep_dup
+  struct["sections"].each do |s|
+    next unless s["kind"] == "listings"
+
+    s["listings"].each_with_index do |l, i|
+      l["listing"]["id"] = listing_ids[i]
+    end
+  end
+  struct
+end
+
+def expect_string(url, string)
+  get url
+
+  expect(response.status).to eq(200)
+  expect(response.body).to match(/#{string}/)
+end
+
+def expect_controller(url, controller_name, action_name)
+  get url
+
+  expect(controller.controller_name).to eq(controller_name)
+  expect(controller.action_name).to eq(action_name)
+end
+
+describe "Landing page", type: :request do
+  before(:all) do
+    # For performance reasons, we set up data in before(:all) so we need to clean it
+    # up ourselves. before(:all) runs outside of transaction, by default.
+    DatabaseCleaner.start
+
+    @domain = "market.custom.org"
+    @community = FactoryGirl.create(:community, :domain => @domain, use_domain: true)
+    @community.reload
+
+    10.times do
+      FactoryGirl.create(:category, community_id: @community.id)
+    end
+
+    10.times do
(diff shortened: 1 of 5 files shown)


app/models/landing_page_version.rb
```diff
@@ -0,0 +1,19 @@
...
+#  created_at   :datetime
+#  updated_at   :datetime
+#
+# Indexes
+#
+#  index_landing_page_versions_on_community_id_and_version  (community_id,version) UNIQUE
+#
+
+class LandingPageVersion < ActiveRecord::Base
+end
```
ovan: Both of these models are already defined in CustomLandingPage::LandingPageStore module. I'm fine with doing them with own model files (although it does feel redundant) but they should only be defined in one place.

spec/requests/landing_pages_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+      :landing_page,
+      community_id: @community.id,
+      enabled: 0,
+      released_version: nil
+    )
+  end
+
+  context "when not released" do
+    it "landing page is disabled" do
+      expect(::CustomLandingPage::LandingPageStore.enabled?(@community.id)).to eq false
```
rap1ds: Hhmh... Somehow I feel that this test could (and maybe should) be removed. It's testing the internals of the application and the HTTP interface. What we really want to test is that the routing is correct and that's in the next two tests. What if we in the future change the mechanism how we define if landing page is in use or not? The routing still works but this test may break even though the app is working ok.

spec/requests/landing_pages_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+    end
+  end
+
+  context "when released" do
+    before do
+      release_version(@community.id, 1)
+    end
+
+    it "landing page is enabled" do
+      expect(::CustomLandingPage::LandingPageStore.enabled?(@community.id)).to eq true
```
rap1ds: Same here

spec/requests/landing_pages_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+
+        release_version(@community.id, 2)
+
+        expect_string("http://#{domain}", "<title>Title version 2</title>")
+      end
+    end
+
+    describe "caching" do
+      before do
+        APP_CONFIG.clp_cache_time = 60
```
rap1ds: Shouldn't you save the original `clp_cache_time` and restore that in `after` block instead of hard-coding the 900 seconds, which may change?

spec/requests/landing_pages_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+      community_id: @community.id,
+      enabled: 0,
+      released_version: nil
+    )
+  end
+
+  context "when not released" do
+    it "landing page is disabled" do
+      expect(::CustomLandingPage::LandingPageStore.enabled?(@community.id)).to eq false
+    end
```
ovan: I would remove this test. It's an implementation detail on this level. Other tests will expose any problems in this functionality already so it just adds fragility to test suite.

spec/requests/landing_pages_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+  end
+
+  context "when released" do
+    before do
+      release_version(@community.id, 1)
+    end
+
+    it "landing page is enabled" do
+      expect(::CustomLandingPage::LandingPageStore.enabled?(@community.id)).to eq true
+    end
```
ovan: Same comment here, unnecessary test, imho.

spec/requests/landing_pages_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+        before do
+          APP_CONFIG.clp_cache_time = 5
+          Rails.cache.clear
+
+          # Stub CACHE_TIME constant, because controller has already been loaded
+          stub_const("LandingPageController::CACHE_TIME", 5.seconds)
+        end
+
+        after do
+          APP_CONFIG.clp_cache_time = 900
```
rap1ds: Same here

spec/requests/landing_pages_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+
+def create_version(community_id, structure, version_number)
+  s = structure.deep_dup
+  s["page"]["title"] = {value: "Title version #{version_number}"}
+  FactoryGirl.create(
+    :landing_page_version,
+    community_id: community_id,
+    version: version_number,
+    content: JSON.generate(s)
+  )
```
ovan: I'm not 100% factory girl is necessary / a good idea here. The other option would be to extend CustomLandingPage::LandingPageStore to include methods for writing data. It would then fully encapsulate the data access and make ActiveRecord an implementation detail. That's why I didn't originally add the model objects under models.

PR #2264, opened 2016-07-08
Safari 1px video gap

rap1ds:
There seems to be a bit annoying issue in Safari: When the window size is changed, every now and then a 1px gap appears in the top or bottom of the video.

I'm not sure what's the reason for this is. It might be a subpixel rendering bug or something like that. It seems to only appear in Safari.

The hacky solution for this is to make the overlay bigger, so that it covers the missing gap. That's why -10px is added to the position.

Before:

screen shot 2016-07-08 at 10 15 52

After:

screen shot 2016-07-08 at 10 16 25

### DIFF
--- a/app/assets/stylesheets/landing_page/sections/video.scss
+++ b/app/assets/stylesheets/landing_page/sections/video.scss
@@ -65,11 +65,22 @@
 }
 
 .video__youtube-overlay {
+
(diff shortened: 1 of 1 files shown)


mporkola:
hack, but well documented and it works 👍

PR #2266, opened 2016-07-08
Refactor FeatureFlagHelper

ovan:
* Move everything out of ApplicationController
* Don't require global @current_community, use the request.env set by middleware instead
* Comment the bad practice of relying on @current_user


rap1ds:
The code looks good but the tests need to be fixed.

PR #2267, opened 2016-07-08
Add configuration option for distance limit

mporkola:
This allows admins to choose whether the search results are limited by maximum distance or not.

### DIFF
--- a/app/controllers/admin/communities_controller.rb
+++ b/app/controllers/admin/communities_controller.rb
@@ -205,7 +205,8 @@ class Admin::CommunitiesController < ApplicationController
         main_search: marketplace_configurations[:main_search],
         main_search_select_options: main_search_select_options,
         distance_unit: marketplace_configurations[:distance_unit],
-        distance_unit_select_options: distance_unit_select_options
+        distance_unit_select_options: distance_unit_select_options,
+        limit_distance: marketplace_configurations[:limit_search_distance]
       }
     else
       render :settings, locals: {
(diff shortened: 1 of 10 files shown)


app/controllers/admin/communities_controller.rb
```diff
@@ -308,7 +309,8 @@ def update_settings
        MarketplaceService::API::Api.configurations.update({
         community_id: @current_community.id,
         main_search: params[:main_search],
-        distance_unit: params[:distance_unit]
+        distance_unit: params[:distance_unit],
+        limit_search_distance: params[:limit_distance].present?
```
mporkola: A checkbox parameter is either `"true"` (string) or `nil`, so `.present?` maps that nicely to boolean

app/controllers/homepage_controller.rb
```diff
@@ -160,12 +160,15 @@ def find_listings(params, listings_per_page, filter_params, includes, location_s
     dropdowns = filter_params[:custom_dropdown_field_options].map { |dropdown_field| dropdown_field.merge(type: :selection_group, operator: :or) }
     numbers = numeric_search_params.map { |numeric| numeric.merge(type: :numeric_range) }
 
-    distance_unit = (location_search_in_use && MarketplaceService::API::Api.configurations.get(community_id: @current_community.id).data[:distance_unit] == :metric) ? :km : :miles
+    marketplace_configuration = MarketplaceService::API::Api.configurations.get(community_id: @current_community.id).data
+
+    distance_unit = (location_search_in_use && marketplace_configuration[:distance_unit] == :metric) ? :km : :miles
```
mporkola: just extracted `marketplace_configuration` to variable for later use

config/locales/en.yml
```diff
@@ -195,6 +195,7 @@ en:
         select_distance_unit: "Show distance in %{distance_units_selector}"
         km: km
         miles: miles
+        show_only_nearby: "Location search shows only nearby listings in the list and grid view"
```
ovan: I'm thinking how clear the behavior will be for admins from this description. I guess we can explain this further in knowledgebase article (have you coordinated an update to that?) if we link to one somewhere here?

PR #2272, opened 2016-07-08
Brightness util (IE fix)

Gnito:


client/app/components/elements/AddNewListingButton/AddNewListingButton.css
```diff
@@ -28,6 +28,7 @@
   left: 0;
   border-radius: calc(var(--AddNewListingButton_height) / 2);
   transition: filter 100ms ease-in;
+  transition: background-color 100ms ease-in;
```
rap1ds: To my understanding, this will now override the filter transition. Multiple transitions need to be comma-separated, i.e. `transition: filter 100ms ease-in, background-color 100ms ease-in`. Or is this different in CSS Next?

rap1ds:
We are now using quite a many different code paths to calculate the darkened color. Have you tested that all those methods return the same value? Based on my testing (there might be errors in my tests) that doesn't seem to be the case:

1. SCSS `darken(#80E619, 20)` returns `#4d8a0f`, tested in http://sass.js.org/ (this is the method used in R&A)
2. `ColorUtils.darken("80E619", 20)` returns `"4D8A0F"`, RSpec test [here](https://github.com/sharetribe/sharetribe/blob/da29ee98f837472a656e1d54daaf1b1048e941e1/spec/utils/color_utils_spec.rb#L7) (this is the method used in landing page)
3. `filter: brightness(80%)` seems to return `rgb(84, 176, 0`, which is `#54B000`, tested [here](https://jsfiddle.net/9102079c/1/) (this is the method used in Topbar if filter is available)
4. `ColorLuminance("#80E619", -0.2)` returns `#66b814`, tested [here](https://jsfiddle.net/9102079c/1/) (this is used in Topbar if filter is not available)

So quite a lot of variation there, assuming my tests were right. I think we should keep these colors unified.

client/app/utils/colors.js
```diff
@@ -0,0 +1,28 @@
...
+  for (let i = 0; i < 3; i++) {
+    const colorComponent = parseInt(hex.substr(i*2,2), 16);
+    const alteredCC = Math.round(Math.min(Math.max(0, colorComponent * lum), 255)).toString(16);
+    rgb += ('00'+alteredCC).substr(alteredCC.length);
+  }
+
+  return rgb;
+};
+
+const brightness = function brightness(hex, brightness) {
```
rap1ds: Since this is a pure function that makes a non-trivial amount of computation, this could be also wrapped in `_.memoize()`

rap1ds:
Using `filter` seems a bit cumbersome because an extra div is needed. Would it make sense to use the same background color calculation and change method to all browsers?

client/app/components/elements/AddNewListingButton/AddNewListingButton.css
Gnito: Ah, true.

client/app/utils/colors.js
Gnito: Good catch, I added memoize to one place, but not to this one.

Gnito:
Color discussion (ref from elsewhere): webkit-filter is closer to intended color in design. 
CSS filter is better solution, since hover styles work even without JS. I added a comment about why colorHover is used (IE11). If these solutions had been created in different order, it might not be worth the trouble to use css filters, but since we have them already - we should definitely keep it. 

I created this brightness util (not just because of the bug in hand) but also because we will need it later (due to custom color solution and IE support).

PR #2279, opened 2016-07-11
FeatureFlagHelper refactor

bladealslayer:


app/helpers/feature_flag_helper.rb
```diff
@@ -1,89 +1,61 @@
...
 
-      is_admin = called_with_admin_user?
-      temp_flags = fetch_temp_flags(is_admin, request.params, request.session)
+  def with_feature(feature_name, &block)
+    block.call if feature_enabled?(feature_name)
+  end
 
-      request.session[:feature_flags] = temp_flags
+  def feature_flags
+    RequestStore.store[:feature_flags]
```
ovan: This should maybe raise if not initialized? The two step flow this now needs is more prone to user errors so providing a bit of help by clearly telling when the poor developer made a mistake would be nice.

app/controllers/landing_page_controller.rb
```diff
@@ -218,7 +213,9 @@ def render_landing_page(default_locale:, locale_param:, structure:)
                          locale_param,
                          topbar_locale)
     marketplace_context = marketplace_context(c, topbar_locale, request)
-    topbar_enabled = feature_enabled?(FEATURE_FLAG)
+
+    FeatureFlagHelper.init(request, false)
+    topbar_enabled = FeatureFlagHelper.feature_enabled?(FEATURE_FLAG)
```
ovan: Not related to this change but now that I see this I think the FEATURE_FLAG constant was a bad call and makes this less clear than just having the symbol directly here.

PR #2281, opened 2016-07-12
Unify hover colors

rap1ds:
We are using three different methods to calculate the hover color based on marketplace color:

* `filter: brightness(80%)` is used in browsers that support it
* `ColorLuminance` JavaScript code is used in browsers that don't support filter
* `ColorUtils.darken` is used in landing page

**The problem** is that brightness filter and ColorLuminance script were resulting the same color, where as `ColorUtils.darken` was returning a different color.

The first two make the brightening operation each component in RBG. The ColorUtils.darken made the operation by following how SCSS does the brightening, that is, covert RBG to HSL and substracting from L component.

**The solution** is that new method `ColorUtils.brightness` is added in this PR. This method uses follows the CSS filter spec and ColorLuminance implementation, which multiplies each RGB component, resulting the same result as brightness filter.

[This JS Fiddle](https://jsfiddle.net/9102079c/4/) can be used to test the result of brightness filter and ColorLuminance

### DIFF
--- a/app/services/custom_landing_page/marketplace_data_store.rb
+++ b/app/services/custom_landing_page/marketplace_data_store.rb
@@ -38,7 +38,7 @@ module CustomLandingPage
         end
 
       color = primary_color.present? ? primary_color : DEFAULT_COLOR
-      color_darken = ColorUtils.darken(color, 15)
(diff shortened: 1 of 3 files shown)


rap1ds:
Reviewed by @Gnito

PR #2284, opened 2016-07-12
Add fallbacks to make pluralisation work

mporkola:
I18n gem selects pluraliser based on locale, but it doesn't understand two-part locale names. Added one-part fallbacks to two-part names to fix it.


mporkola:
this approach didn't work, https://github.com/sharetribe/sharetribe/pull/2292 instead

PR #2286, opened 2016-07-12
Remove landing page topbar feature flag

rap1ds:


app/controllers/landing_page_controller.rb
```diff
@@ -213,9 +213,6 @@ def render_landing_page(default_locale:, locale_param:, structure:)
                          topbar_locale)
     marketplace_context = marketplace_context(c, topbar_locale, request)
 
-    FeatureFlagHelper.init(request, false)
```
bladealslayer: We still need this, because we pass `feature_flags` as local (change came in recent merge, so you need to rebase to see).

PR #2287, opened 2016-07-12
Refactor analytics partials to use locals instead of globals

bladealslayer:


app/views/landing_page/landing_page.erb
```diff
@@ -52,7 +52,10 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
 
   <% unless local_assigns.has_key?(:skip_analytics) && skip_analytics %>
-    <%= render partial: "layouts/google_analytics_script", locals: {feature_flags: FeatureFlagHelper.feature_flags} %>
+    <%= render partial: "layouts/google_analytics_script",
+               locals: { feature_flags: FeatureFlagHelper.feature_flags,
```
ovan: Even here I would prefer passing the feature flags down from the controller as a local. Now the difference is that there's an implicit dependency from controller having initialized the FeatureFlagHelper properly that using locals would make an explicit part of the API.

PR #2288, opened 2016-07-12
Adding an s to the name of the controller

Luis-RG:
Adding an s to the name of the controller to match the YML file name.


ovan:
The description of the PR confused me totally but the change itself looks good. :) Your adding an s to translation key, not to controller name.

PR #2289, opened 2016-07-12
Admin maps settings

bladealslayer:


config/routes.rb
```diff
@@ -188,8 +188,10 @@
           post :payment_gateways, to: 'communities#create_payment_gateway'
           get :social_media
           get :analytics
+          get :maps
```
rap1ds: To my understanding this will create a route with path like /admin/communities/253462/maps. We have decided that we don't want to expose the community id in the URL. However, some old URLs still do it.

So I'd suggest that you change this. Have a look how the route for look and feel and details have been configured: https://github.com/sharetribe/sharetribe/pull/2289/files#diff-21497849d8f00507c9c8dcaf6288b136R172

features/admin/maps/admin_configures_maps.feature
```diff
@@ -0,0 +1,21 @@
...
+      | kassi_testperson2 | jane       | doe         | test1@example.com   | 2012-03-01 00:00:00 +0200 |
+    And I am logged in as "manager"
+    And "manager" has admin rights in community "test"
+    And I am on the maps admin page
+
+  Scenario: Admin adds a google maps api key
+  When I add "fookey" to the Google Maps API key field
+  And I press submit
+  And I refresh the page
+  Then I should see "fookey" in the Google Maps API key field
```
rap1ds: I have a personal problem with Cucumber tests. 😟  They are slow and quite verbose. We haven't agreed anything, but personally I would even suggest that we don't add any new Cucumber tests.

Would it make sense to test this with RSpec instead? I don't think the browser interaction brings any additional value to this test, so maybe we could test this without browser, that is, with RSpec? Or, of course, one option is not to add test at all if the feature is simple and test doesn't bring any value...

What do you think?

app/views/admin/communities/maps.haml
```diff
@@ -0,0 +1,24 @@
...
+    = "-"
+    = t("admin.communities.maps.maps")
+
+= render :partial => "admin/left_hand_navigation", :locals => { :links => admin_links_for(@current_community) }
+
+.left-navi-section
+  %h2= t(".maps", :community_name => @community.name(I18n.locale))
+  = form_for @community, :url => maps_admin_community_path(@community), :method => :put do |form|
+    .row
+      %h3= t(".google_maps_key")
```
rap1ds: We have agreed to use full paths for translations keys `admin.communities.maps.google_maps_key`. The problem with the shorter version is that they are tightly coupled to the filename and directory, so renaming the file or moving it breaks the translations.

features/admin/maps/admin_configures_maps.feature
bladealslayer: I added the test, since here everything was similar to the `analytics` page. Perhaps we can refactor them afterwards in one go?

PR #2292, opened 2016-07-13
Pluralization monkey patch for tr-TR

mporkola:
Pluralizer is selected by language code, but I18n gem doesn't understand two-part locales we use for many locales. Default fallback is "one-other" pluralisation, but Turkish needs an "other" type pluralizer.

This should be generalized to drop the region code from two-part locales.


rap1ds:
LGTM

PR #2295, opened 2016-07-13
Prepare Landing page OS release

rap1ds:
This PR adds an ability to use "static" landing page, e.g. landing page where the content comes from `example_data.rb` file instead of database. In addition, this PR adds documentation how to setup, enable, configure and modify landing page.

This PR is built on top of #2301


app/controllers/landing_page_controller.rb
```diff
@@ -24,7 +24,7 @@ class LandingPageController < ActionController::Metal
   CACHE_TIME = APP_CONFIG[:clp_cache_time].to_i.seconds
   CACHE_HEADER = "X-CLP-Cache"
 
-  FONT_PATH = APP_CONFIG[:font_proximanovasoft_url].present? ? APP_CONFIG[:font_proximanovasoft_url] : "/landing_page/fonts"
```
ovan: I wanted to use more generic parameter name because the font won't only be used on landing page but everywhere later.

config/config.defaults.yml
```diff
@@ -301,14 +301,52 @@ default: &default_settings
...
+  #
+  # Must include trailing slash.
+  #
+  # Example values:
+  #
+  # - empty: don't use Proximan Nova Soft font
+  # - landing_page/
+  # - https://s3.amazonaws.com/my-bucket/fonts/
+  #
+  clp_font_proximanovasoft_url:
```
ovan: Commented already in other file but I'd rather keep the previous config option name and not scope this to landing pages.

docs/landing-page.md
```diff
@@ -0,0 +1,70 @@
...
+
+#### Hosting the assets in /public (Recommended)
+
+#### Hosting the assets in S3 ()
+
+### Fonts
+
+Landing page is designed to be used with Proxima Nova Soft font, which is not freely available. If you want to use Proxima Nova Soft font in the landing page, you need to buy the font.
+
+`clp_font_proximanovasoft_url` configuration controls the path where the font is loaded from.
```
ovan: Here's a reference to the font variable name which I commented about, so a reminder to change this one too.

app/services/custom_landing_page/landing_page_store_db.rb
```diff
@@ -0,0 +1,102 @@
+module CustomLandingPage
+  module LandingPageStoreDB
+
+    class LandingPage < ActiveRecord::Base; end
+    class LandingPageVersion < ActiveRecord::Base; end
+
+    module_function
+
+    #
+    # Common methods
```
bladealslayer: Perhaps clarify that this is "common interface with the static CLP store".

docs/landing-page-structure.md
```diff
@@ -0,0 +1,276 @@
...
+
+## Settings
+
+The following values are set in `settings`:
+
+* `marketplace_id` is the ID of the marketplace (community) in the database
+* `locale` is the locale for the landing page
+* `sitename` is the name of the landing page as chosen when created
+
+Typically, you would only need to change the `locale` if the marketplace language is changed. Change the `marketplace_id` only if you made a mistake and specified incorrect ID when creating the landing page. Do not change the `sitename`.
```
bladealslayer: Probably the comment about `marketplace_id` is not needed in OS docs.

docs/landing-page-structure.md
```diff
@@ -0,0 +1,276 @@
...
+  * `search`: links to search page (the old homepage)
+  * `signup`: signup page
+  * `login`: login page
+  * `post_a_new_listing`: post a new listing page
+  * `all_categories`: search page with all categories selected
+  * `how_to_use`
+  * `terms`
+  * `privacy`
+
+  If customers are hoping to link to some other internal page that is not available, let the development team know and they are happy to add new paths.
```
bladealslayer: Re-word perhaps without "customers" or remove the entire sentence?

docs/landing-page-structure.md
```diff
@@ -0,0 +1,276 @@
...
+* `social` is an array of social media links. Edit each `url` to be the correct profile URL for the marketplace.
+
+  Available services are:
+
+  * `facebook`
+  * `instagram`
+  * `twitter`
+  * `youtube`
+
+  If customers are hoping links to other social media services, let Janne and the development team know.
```
bladealslayer: Same as above.

docs/landing-page.md
```diff
@@ -0,0 +1,110 @@
+# Landing page
+
+_This feature is an experimental feature. The setup and configuration of the landing page needs manual code editing. There's no user interface in Admin panel to configure landing page._
```
bladealslayer: -> "This is an experimental feature."

docs/landing-page.md
```diff
@@ -0,0 +1,110 @@
...
+
+To enable the landing page, change the `clp_static_enabled` to `true`
+
+```yaml
+# config.yml
+
+clp_static_enabled: true
+```
+
+Restart the server and go to the marketplace root URL. You should now see the landing page with the example content.
```
bladealslayer: This won't work out of the box. Example content links to category and listing IDs that won't be there and that was causing an error at least earlier.

docs/landing-page.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Advanced configuration
+
+### Caching
+
+The landing page is cache heavily in order to make it lightning fast. There are two configurations that control caching:
```
bladealslayer: "is heavily cacheD", "There are two configuration options" perhaps?

docs/landing-page.md
```diff
@@ -0,0 +1,110 @@
...
+
+- `clp_cache_time`: The time in seconds how often the cache is invalidated. Default: 900 seconds, i.e. 15 minutes
+- `clp_static_released_version`: The current landing page version. The version number acts as a cache buster. The cache is in any case invalidated after `clp_cache_time`, but if you want to invalidate the cache sooner, increase the version number.
```
bladealslayer: Mention that the landing page is also cached browser-side (we send `Cache-Control: max-age=#{clp_cache_time}`), so some users would still see older version, even if the app-side cache is invalidated.

docs/landing-page.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+#### Hosting the images in S3
+
+Change the configuration to match the image location in your S3 bucket:
+
+```yaml
+# config.yml
+
+font_proximanovasoft_url: `https://yourbucketnamehere.s3.amazonaws.com/landing_page/assets/`
```
bladealslayer: Shouldn't this be `clp_asset_url` here?

docs/landing-page.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+Even if you use S3 to host the images, it's recommended to use CDN in front of the S3 bucket.
+
+```yaml
+# config.yml
+
+font_proximanovasoft_url: `https://your-cdn-service.com/landing_page/assets/`
```
bladealslayer: same as above

docs/landing-page.md
rap1ds: Actually it does work. The example content has listings and categories section defined, but the sections are not referred from the `composition` section which constructs the page.

PR #2301, opened 2016-07-14
Landing page configuration changes

rap1ds:
This PR contains some landing page configuration changes:

Fonts:

- [X] Font partial is not rendered, if font path is not present
- [X] In the configuration comments, trailing slash is recommended for `font_path`
- [X] `_fonts.erb` partial tolerates and normalized multiple slashes

Assets:

- [X] `clp_asset_host` renamed to `clp_asset_url`
- [X] In the configuration comments, trailing slash is recommended for `clp_asset_url`
- [X] `clp_asset_url` configuration can now contain optional `${sitename}` placeholder, which will be replaced with the landing page site name

### DIFF
--- a/app/services/custom_landing_page/link_resolver.rb
+++ b/app/services/custom_landing_page/link_resolver.rb
@@ -59,7 +59,7 @@ module CustomLandingPage
 
       def append_asset_path(asset)
         host = @_asset_url || ""
-        src = URLUtils.join(@_asset_url, asset["src"]).sub("${sitename}", @_sitename)
+        src = URLUtils.join(@_asset_url, asset["src"]).sub("%{sitename}", @_sitename)
 
         asset.merge("src" => src)
       end
--- a/app/services/custom_landing_page/link_resolver.rb
+++ b/app/services/custom_landing_page/link_resolver.rb
@@ -58,12 +58,10 @@ module CustomLandingPage
       private
 
       def append_asset_path(asset)
-        if @_asset_host.present?
-          asset.merge("src" => [@_asset_host, @_sitename, asset["src"]].join("/"))
-        else
-          # If asset_host is not configured serve assets locally
-          asset.merge("src" => ["landing_page", asset["src"]].join("/"))
-        end
+        host = @_asset_host || ""
+        src = URLUtils.join(@_asset_host, asset["src"]).sub("${sitename}", @_sitename)
+
+        asset.merge("src" => src)
       end
     end
 
--- a/app/services/custom_landing_page/link_resolver.rb
+++ b/app/services/custom_landing_page/link_resolver.rb
@@ -35,12 +35,12 @@ module CustomLandingPage
     end
 
     class AssetResolver
-      def initialize(asset_host, sitename)
+      def initialize(asset_url, sitename)
         unless sitename.present?
           raise CustomLandingPage::LandingPageConfigurationError.new("Missing sitename.")
         end
 
-        @_asset_host = asset_host
+        @_asset_url = asset_url
         @_sitename = sitename
       end
 
@@ -58,8 +58,8 @@ module CustomLandingPage
       private
 
       def append_asset_path(asset)
-        host = @_asset_host || ""
-        src = URLUtils.join(@_asset_host, asset["src"]).sub("${sitename}", @_sitename)
+        host = @_asset_url || ""
+        src = URLUtils.join(@_asset_url, asset["src"]).sub("${sitename}", @_sitename)
 
         asset.merge("src" => src)
       end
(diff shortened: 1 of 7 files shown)


app/utils/url_utils.rb
```diff
@@ -43,4 +43,13 @@ def extract_locale_from_url(url)
   def strip_port_from_host(host)
     host.split(":").first
   end
+
+  # Naive join method, which can be used to normalize multiple slashes
+  #
+  # Usage: URLUtils.join("foo", "bar/", "baz") => "foo/bar/baz"
+  def join(*parts)
```
bladealslayer: I think it would be better if this handles also calls like:

```
join("foo", "/bar")
```

config/config.defaults.yml
```diff
@@ -301,13 +301,37 @@ default: &default_settings
...
+  #
+  # Because a directory is expected, it's recommended to include trailing slash.
+  #
+  # You can add %{sitename} in the URL and it will be replaced with the sitename,
+  # that is saved in the landing page structure JSON.
+  #
+  # Example values:
+  #
+  # - landing_page/
+  # - https://s3.amazonaws.com/my-bucket/sites/${sitename}/
```
bladealslayer: Example above is with `%{sitename}`, and here it's `${sitename}`. Which is the correct one?

app/services/custom_landing_page/link_resolver.rb
```diff
@@ -58,12 +58,10 @@ def call(type, id, normalized_data)
       private
 
       def append_asset_path(asset)
-        if @_asset_host.present?
-          asset.merge("src" => [@_asset_host, @_sitename, asset["src"]].join("/"))
-        else
-          # If asset_host is not configured serve assets locally
-          asset.merge("src" => ["landing_page", asset["src"]].join("/"))
-        end
+        host = @_asset_host || ""
```
bladealslayer: Maybe rename to `@_asset_url` here as well?

app/utils/url_utils.rb
```diff
@@ -43,4 +43,13 @@ def extract_locale_from_url(url)
...
+    parts.select(&:present?).reduce { |a, b|
+      "#{a.sub(/\/+\z/, "")}/#{b}"
```
bladealslayer: How about using `File.join`?

rap1ds: @bladealslayer Yeah, I thought about that as well, but the problem with `File.join` is that the result depends on the OS. It's `foo/bar` for Unix and `foo\bar` for Windows.

app/utils/url_utils.rb
rap1ds: What would be the expected result for `join("foo", "/bar")`? `foo/bar` I guess?

Usually (at least in URL context), the prepending slash `/bar` means path from the URL root. E.g. if the current URL for web page is `http://example.com/foo/` and there's a link pointing to `/bar`, the result is `http://example.com/bar`.

So that's why it's not obvious what the result would be for joining path parts with prepending slashes and that's why I intentionally left it out.

What do you think, how should I proceed?

bladealslayer: I'd expect that `join("foo", "/bar") == "foo/bar"`. At least the method name suggests it's all about joining components and also Ruby's own `File.join` works similarly. I'd think of it like this: User wants a path relative to the arbitrary root of their location (the `/bar` part) and `foo` is the system wide prefix for that.

rap1ds: Tested how `URI.join` works:

```
2.3.1 :002 > URI.join("http://www.example.com", "foo", "/bar")
 => #<URI::HTTP http://www.example.com/bar>
2.3.1 :003 > URI.join("http://www.example.com", "foo", "/bar", "baz", "/quaax")
 => #<URI::HTTP http://www.example.com/quaax>
```

rap1ds: `URI.join` would be great for the purpose, but the problem with it is that it always expect that the result is absolute URL. Relative URLs throw error:

```
2.3.1 :005 > URI.join("foo", "/bar", "baz", "/quaax")
URI::BadURIError: both URI are relative
```

app/utils/url_utils.rb
bladealslayer: Yeah, I thought about the same, but found this: https://www.ruby-forum.com/topic/50137. To make sure, perhaps the `File.join` call can be wrapped in setting and restoring the `File::SEPARATOR`?

app/utils/url_utils.rb
bladealslayer: Hm, alright. Since this is URLHelper, then I guess it's ok. Somehow feels strange to me that a later part can override a (mandatory?) prefix.

If you keep the behaviour of `URI.join`, perhaps add a test case to document that?

rap1ds: Yeah, as you said, Ruby seems to use `/` for all OSes. So I changed the code to just use `File.join`.

Issue #2204, opened 2016-06-30
Add sitemap support


mooreds:
This may be more of a feature request, so feel free to close, but I'd love sitemap support, even if it was just listings.  This gem looks helpful: https://github.com/kjvarga/sitemap_generator

mooreds:
Whoops, I see that this is already a feature request: http://support.sharetribe.com/forums/240322-sharetribe-development-ideas/suggestions/7103653-add-a-sitemap-xml-file-to-marketplace

mooreds:
Would y'all be amenable to a PR adding this feature for listings only?

@rap1ds ?

rap1ds:
I'm not very familiar with sitemaps (and SEO stuff in general). @thomasmalbaux what do you think? Should we have a sitemap support?

thomasmalbaux:
Hey there! Thanks for the suggestion. Adding a sitemap would definitely make sense and would be a great addition!

This probably requires a bit of thinking though, as such sitemap should be updated as soon as listings are opened or closed.
Same thing with new users (when they create/delete their profiles) and categories (that can be updated). Even though I understand that it wouldn't be part of a first version, which might make sense).

Adding static pages ("About/Terms/..." pages) would probably be quite easy and would make sense, on top of listings.
The `<priority>` and `<changefreq>` tags could also be discussed to be relevant enough (for example for static pages).
Finally there might be some challenges with pictures and assets, but they maybe could be left out of a first sitemap version.

Again overall, this would definitely be a great improvement! As you mentioned @mooreds it has been mentioned a few times by Sharetribe users and sitemaps are key for SEO to help content discovery.

Happy to continue this discussion a bit (even though I couldn't help with technical details) if you are interested to consider this feature.

mooreds:
A bigger question for me is how to do this for the open source solution and at the same time make it worth with the different communities.  Maybe the default sitemap url is pulled from the community?

As far as pages, I hear you about the complexities.  For a first pass, I'd just want to get the open listings, and maybe try to do the static pages if it made sense.  

If a listing is closed, does it need to be removed from the sitemap immediately, or could it wait for a day?  I'd definitely suggest a daily task to rebuild the sitemaps, rather than rebuilding them on demand.

mooreds:
Hi folks,

I did this for my local version of sharetribe (based off of 5.6, though I plan to keep it up to date).

It involved a new gem, a new controller, a new config file, and a couple of rake tasks.

[...]

PR #2303, opened 2016-07-14
Add sitemap to sharetribe

mooreds:
This is for issue https://github.com/sharetribe/sharetribe/issues/2204

@rap1ds I didn't implement the CDN stuff yet, wanted to get your feedback on how things looked first (ask early, ask often).


mooreds:
Looks like this failed some formatting checks.  Please let me know if there are other changes to make (beyond correcting those) and I'll do so.

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+            default_host = "https://"+APP_CONFIG.domain
+          end
+
+          community_id = @current_community.id
+
+          SitemapGenerator::Sitemap.create(
+                :default_host => default_host,
+                :verbose => false,
+                :adapter => adapter) do
+                  Listing.where(deleted: false, open: true, community_id: community_id).find_each do |listing|
```
rap1ds: This needs to be tuned a bit. It's now over-fetching listings because it doesn't take into account the `valid_until` date. There's a `currently_open` method in `Listing` model that is useful for this purpose.

I deployed this branch to our testing server and tested the performance with one of our test marketplaces which has tens of thousands listings.

The `GET /sitemap.xml.gz` was completed in ~5000ms (with empty cache), which is way too much. Something like ~300ms would be sufficient.

The reason for this was the over-fetching. `Listing.where(deleted: false, open: true, community_id: community_id)` returned ~9000 listings.

Then I deployed a modified version (using `Listing.currently_open.where(community_id: community_id)` instead). Now ~900 listings were returned and the rendering took something like ~500ms. Not exactly a fast request, but good enough in my opinion, because the request is cached heavily.

In addition, [pluck](http://apidock.com/rails/ActiveRecord/Calculations/pluck) might make the data fetching a bit faster. Because you only need `id` and `updated_at`, `.pluck(:id, :updated_at)` can accomplish the same job.

mporkola: 👍 for `pluck`: it skips instantiating ActiveRecord objects, so the speedup might be quite big

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  def generate
+    render_site_map
+  end
+
+  private
+
+    def render_site_map
+      respond_to do |format|
+
+        data = Rails.cache.fetch(@current_community, expires_in: 24.hours) do
```
rap1ds: To my understanding, if you use `@current_community` as a cache key, the cache will be invalidated if the marketplace admin does any change to the marketplace that changes the `@current_community.updated_at`, and that's probably unwanted behaviour. So I'd recommend to use the marketplace ID instead of the community object.

In addition to the markeplace ID, the cache key needs to have some prefix to prevent it from clashing. So something like this:

```ruby
Rails.cache.fetch("sitemap-#{@current_community.id}", expires_in: 24.hours) do ...
```

Does it make sense?

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+          if @current_community.use_domain 
+            default_host = "https://"+@current_community.domain
+          else 
+            default_host = "https://"+APP_CONFIG.domain
+          end
+
+          community_id = @current_community.id
+
+          SitemapGenerator::Sitemap.create(
+                :default_host => default_host,
```
rap1ds: Would `request.base_url` be enough? I guess it would...

```ruby
:default_host => request.base_url,
```

app/controllers/sitemap_controller.rb
rap1ds: Couple more ideas:

We've been doing some decoupling of controller code and ActiveModels. You've probably seen those "Service" layer implementations here and there in the code. In this case, could you add a new method to controller, say, `find_open_listings(community_id)` which would return an array of listing IDs and update timestamps? Then the controller code would be at least a bit more decoupled from ActiveModel, because the controller code doesn't care anymore whether the data is coming from database, cache, file, over the network e.g. as long as it returns an array of IDs and timestamps.

* Should we add some upper limit to the listings? Like, 5000 newest open listings? If there's a marketplace which has 10000000 open listings and we would fetch them all and render the sitemap, that would be terrible for performance.

config/routes.rb
```diff
@@ -10,6 +10,8 @@
   # first created -> highest priority.
 
   get "/robots.txt" => RobotsGenerator
+  get "/sitemap.xml.gz" => "sitemap#sitemap", format: :xml
```
rap1ds: I'm not sure how search engines work in this case, but I'm thinking do we need to add also `/sitemap.xml` route?

spec/controllers/sitemap_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  describe "GET #sitemap_generate" do
+    it "creates sitemap" do
+      get :generate, {}, valid_session
+      expect(response.status).to eq(200) 
+    end
+  end
+
+end
```
rap1ds: Nice test! 👍

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+                :adapter => adapter) do
+                  Listing.where(deleted: false, open: true, community_id: community_id).find_each do |listing|
+                    add listing_path(id: listing.id), :lastmod => listing.updated_at
+            end
+          end
+
+          #SitemapGenerator::Interpreter.run(:default_host => "http://www.bar.com", :adapter => adapter)
+          data = adapter.get_data
+        end
+        format.xml_gz { send_data  ActiveSupport::Gzip.compress(data) }
```
rap1ds: I wonder if we need this one. This might be redundant since we already use `Rack::Deflater` https://github.com/sharetribe/sharetribe/blob/master/config.ru#L4 . But I'm not sure...

rap1ds:
You need to also check if the marketplace is private, e.g. `@current_community.private?` and if it is a private marketplace, you shouldn't add the links to the listings. In private marketplaces, listings can't be accessed without being logged in, so they shouldn't be indexed in search engines.

rap1ds:
@mooreds This is awesome! 🎉  Thanks for the Pull Request!

I hope I don't scare you off with too many comments 😆 The direction in this PR is definitely correct and we'd love to merge this to master after some changes and polishing!

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,47 @@
+class SitemapController < ApplicationController
```
rap1ds: Are you planning to use the Metal controller? If you do, you can use `request` object to access the current community like this: https://github.com/sharetribe/sharetribe/blob/master/app/controllers/landing_page_controller.rb#L188

app/controllers/sitemap_controller.rb
mooreds: A couple of things.

I realized that we use the id rather than the object, which means that the sitemap points to `/listings/42` as opposed to `/listings/42-cool-listing-name`.  The latter is obviously superior, so I'll see if I can fix that.  (Not sure what that means for the `pluck` suggestion.)

I can definitely add a wrapper method in the controller to abstract the activemodel code.  

Re: large number of links, I think we should limit to at most 50k (see https://github.com/kjvarga/sitemap_generator/blob/master/lib/sitemap_generator.rb#L26 ).  Otherwise we get into a sitemap situation where we have to have multiple sitemap files, and I haven't tested that.  As for whether it should be less, that's really a question for you.  I'd say that if I were running a marketplace, I'd want all the listings to be searched, and would rather deal with the performance issues.  

Re: performance--it sounds like using `currently_open` is OK for now?  I'll implement that.

app/controllers/sitemap_controller.rb
mooreds: Absolutely.  Will make the change.

app/controllers/sitemap_controller.rb
mooreds: Duly noted.

config/routes.rb
mooreds: I don't think you have to.  http://www.sitemaps.org/protocol.html states: "If you would like, you may compress your Sitemap text file using gzip to reduce your bandwidth requirement."  I think you'd just let your users know that they needed to submit the `.gz` url to search engines, and that `sitemap.xml` will 404.

Speaking of that, I also need to add an entry to the robots.txt file with the sitemap name.  Will do that.

mooreds:
No worries about the comments, thanks for all the direction.  

I originally implemented with the Metal controller but had trouble figuring out how to get the current community.  Thanks for the pointer--will use that.

config/routes.rb
mooreds: > Speaking of that, I also need to add an entry to the robots.txt file with the sitemap name. Will do that.

Actually, looks like the robots.txt file isn't generated dynamically, so we can't point to the sitemap.xml.gz file there (and the link in there needs to be absolute, can't be relative: https://moz.com/community/q/robots-txt-sitemap-with-relative-path )

app/controllers/sitemap_controller.rb
mooreds: I am unsure.  Do you have any suggestions on how to find out?

mooreds:
@rap1ds made most of the changes you suggested and added some additional tests.

Please take another look.  The only open issues:

* limits on how many listings to display (I just set it to 50k)
* no use of pluck so we can get nice urls
* question about gzipping vs rack deflate.

Thanks.

mooreds:
@rap1ds any feedback?

app/controllers/sitemap_controller.rb
bladealslayer: You can test by serving this as uncompressed XML from controller and check in browser what response you get if your request specifies gzip as supported (sends e.g. header `Accept-Encoding: gzip, deflate`). If you get response with e.g. `Content-Encoding: gzip`, then it works.

bladealslayer:
@mooreds I'll take a look at this as soon as I can.

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        head :not_found
+        return
+      end
+      if community.private?
+        head :forbidden
+        return
+      end
+      respond_to do |format|
+
+        data = Rails.cache.fetch(community.id, expires_in: 24.hours) do
```
bladealslayer: It would be better to have some prefix to the cache key. Something like `"sitemaps/#{community.id}"` would be great!

bladealslayer:
About pluck and nice URLs for listings. Looking at https://github.com/sharetribe/sharetribe/blob/master/app/models/listing.rb#L143 it looks like you can form the correct pretty URL by plucking the listing `id` and `title` (and `updated_at` for the `:lastmod`. Pluck is really going to help with performance when there's large number of listings.

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+          #SitemapGenerator::Interpreter.run(:default_host => "http://www.bar.com", :adapter => adapter)
+          data = adapter.get_data
+        end
+        format.xml_gz { send_data  ActiveSupport::Gzip.compress(data) }
+        format.html { send_data  ActiveSupport::Gzip.compress(data) }
+      end
+    end
+
+    def self.find_open_listings(community_id)
+      listings = Listing.currently_open.where(community_id: community_id).limit(SitemapGenerator::MAX_SITEMAP_LINKS)
```
bladealslayer: So that's 50K? Would it make sense for this to be a configuration option, defined in `config/config.defaults.yml` perhaps?

app/controllers/sitemap_controller.rb
bladealslayer: It seems to me that it would be better to compress the data before it's written to cache. For large marketplace the sitemap can be of considerable size and we have to keep in mind the case where there are thousands of marketplaces around. That would actually help per request performance too, since you'll do the compression once.

bladealslayer:
In #2204 there was discussion about redirecting from the public sitemap URL to a CDN and using another route for the actual controller action that renders the sitemap. Did you find out some reason why that wouldn't work well @mooreds? CDN feel like a better place to cache the sitemaps instead of in-memory cache.

app/controllers/sitemap_controller.rb
mooreds: Easily done.  Will do that.

mooreds:
re: https://github.com/sharetribe/sharetribe/pull/2303#issuecomment-234026068 

That seems like a good alternative.  My worry is that somewhere down the line the format for a nice looking URL will change, and that we'll forget to update this controller.  Would rather defer to the model, so that logic is all in once place.  That said, I guess I could write a test which would fail if the logic ever changed.

app/controllers/sitemap_controller.rb
mooreds: It's easy enough to go down from that number and make that a configuration parameter.  If we go above that number, we need to split apart the sitemaps (which isn't supported with the current code).  How many marketplaces do you have that have more than 50k listings?

That said, if you're creating a sitemap for SEO, I can't imagine wanting to have fewer than every listing available for google, bing, etc to index.  What am I missing?

mooreds:
re: https://github.com/sharetribe/sharetribe/pull/2303#issuecomment-234032167 I guess I felt like the CDN was additional work that I didn't see a lot of value in, especially since we're caching for 24 hours and not generating any files.  That said, you know your operating environment better than I do, so if a CDN option is needed to merge, I can look into what that entails.  I guess I just took the easier route :)

app/controllers/sitemap_controller.rb
mooreds: Just added code to compress once and cache that.

bladealslayer:
@mooreds about CDN - I'm not completely sure yet, I'll need to test the implementation a bit more. If rendering ends up being fast enough, CDN would be better than wasting cache memory.

Do you mind if I pull your commits in a working branch so that we can more easily test / deploy your code? We can keep polishing if needed within this PR. Alternatively, you can close this PR and open a new one merging to `sharetribe:sitemaps` branch, instead of `sharetribe:master`, so that we have time to test / change implementation, before merging to master.

app/controllers/sitemap_controller.rb
bladealslayer: Let's leave it as is for now. The only reason why we might want to limit the number downward is if rendering is really slow.

mooreds:
re: https://github.com/sharetribe/sharetribe/pull/2303#issuecomment-234168148 yes, please feel free to do whatever you need to do to get this code to the point where you can vote thumbs up or thumbs down :).  

I implemented pluck instead of pulling the whole listing.  I realized I'd already written a test checking for the nice URL ( https://github.com/sharetribe/sharetribe/pull/2303/files#diff-b1be3b5b892346cbc2ee3ac2dd65e569R40 ) so we're covered there.

Please let me know if you have further feedback or need something more from me.  I'd rather you handle the git branching stuff, but if you don't have cycles or need me to, let me know and I'll do my best.

Also, just so you know, on Saturday I'll be on vacation for a week, so won't be responding to anything on this PR thread.

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+    def self.find_open_listings(community_id)
+      listing_data_array = Listing.currently_open.where(community_id: community_id).limit(SitemapGenerator::MAX_SITEMAP_LINKS).pluck(:id, :title, :updated_at);
+      listings_hashes = []
+      listing_data_array.each do |l|
+        hash = {}
+        # this is from Listing.to_param: "#{id}-#{title.to_url}"
+        hash[:id] = "#{l[0]}-#{l[1].to_url}"
+        hash[:lastmod] = l[2]
+        listings_hashes.push(hash)
+      end
```
bladealslayer: You could use `map` here to process the array. It would make the code more readable. Something like this:

```ruby
Listing
  .currently_open
  .where(community_id: community_id)
  .limit(SitemapGenerator::MAX_SITEMAP_LINKS)
  .pluck(:id, :title, :updated_at)
  .map { |l|
    # this is from Listing.to_param: "#{id}-#{title.to_url}"
    {id: "#{l[0]}-#{l[1].to_url}", lastmod: l[2]}
  }
```

bladealslayer:
@mooreds this looks great! Few small things:

* see my inline comment about using `map`
* please, check out what `rubocop` is complaining about and fix the issues it's reporting (you can see the build details by clicking the `Details` link in the "CI build failed" line at the bottom, or you can run `bundle exec rubocop` locally in your development environment
* squash all commits so far into a single one

While you're away next week, I'll try to test this and will let you know how to proceed from there.

Thanks again for your contribution!

mooreds:
@bladealslayer 

I updated to use map and fixed the rubocop issues.  Squashed all the commits into one, and pushed it up again.

Please let me know if there's more to do.

Thanks

mooreds:
@bladealslayer ping.

bladealslayer:
@mooreds sorry, busy days. I haven't had a chance to test this yet.

mooreds:
@bladealslayer Understood.  should I just ping you every few days, or what's the best path forward here?

bladealslayer:
@mooreds no need for pings, unless you don't hear anything for more than a week :) BTW, your branch no longer can merge cleanly, so I'd suggest you rebase on latest master and resolve conflicts.

mooreds:
@bladealslayer ok.  the rebase is done, I believe.

mooreds:
Sorry about that!

bladealslayer:
Actually, it a merge of your correct rebase with the old version of the branch, so pretty weird thing :) But no worries.

mooreds:
ping

rap1ds:
@mooreds pong! I just came back from my holiday and will look at this soon!

rap1ds:
(soon, meaning in the upcoming days)

mooreds:
ping

rap1ds:
pong, still pending me :(

rap1ds:
@mooreds Finally I had some time to work with this! I cherry-picked your commits and created a new clean branch: #2492

mooreds:
@rap1ds Thanks for letting me know.  Seems like you did a fair bit of rewriting--hope the code was useful to you in some way.  Excited to see this finally released!  

Closing since this PR will never be merged :).

rap1ds:
@mooreds Thanks for you help! Your code was a very helpful! I needed to do some more work with CDN and caching.

The branch is now merged to master and deployed to sharetribe.com. We're still a bit concerned about the performance, so we'll monitor the performance for a while before officially announce this feature (i.e. before adding it to the CHANGELOG)

PR #2307, opened 2016-07-15
Add configuration option to set default Google Maps API key

bladealslayer:


app/view_utils/marketplace_helper.rb
```diff
@@ -0,0 +1,9 @@
+module MarketplaceHelper
+  module_function
+
+  def google_maps_key(cid)
+    community_key = Community.where(id: cid).pluck(:google_maps_key).first
```
lyyder: To find by id, also `find` could be used: `Community.find(cid)`

bladealslayer: We need `where` here, as it returns an `ActiveRecord::Relation`, on which we can call `pluck` to avoid building a full ActiveRecord object for the community.

PR #2308, opened 2016-07-17
Fixes to docker-compose setup - work in progress

timabell:
The docker-compose (formerly "fig") setup has got quite out of date it seems and stopped working.

The included changes are enough to bring up the docker-compose web container plus dependencies and get the web one talking to the database.


timabell:
Okay I got this to fire up the "set up your marketplace" page with all the styles missing. Realised that the readme now says use foreman. I've pushed some more commits that bring that in, and a couple of fixes, but it doesn't currently run.

bladealslayer:
Hi @timabell, thanks for contributing! Indeed, the docker-based dev setup for Sharetribe has been left alone for a while. How far did you get this working?

I noticed the TODO note about `sphinxsearch`. On Debian 8 (which is what the `ruby:2.3.1` image is based on) you can install sphinx by first adding the `jessie-backports` repository. Something like this should work:

```
RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/apt/sources.list.d/backports.list
RUN apt-get update && apt-get install -y sphinxsearch
```

docker-compose.yml
```diff
@@ -0,0 +1,70 @@
+version: '2'
+services:
+  db:
+    image: mysql
+    ports:
+      - "3306:3306"
```
bladealslayer: I'd run this mapped to a non-standard port.

bladealslayer:
The error you get might be due to incompatible `node` version. The version in Jessie is ancient :) A good approach would be to install `nvm` and install `node` v.6.1.0 with it.

bladealslayer:
Closing, as this has been inactive for a while.

PR #2313, opened 2016-07-19
User-specific feature flags

lyyder:
Currently our feature flags can only be set for a marketplace. The plan is to extend this feature by adding user-specific feature flags that allow enabling a feature for a single user.

The greater plan here is to allow marketplace admins to enable features in their marketplace. Currently this could be achieved with the marketplace-specific feature flags but with the user-specific feature flags an admin could enable the feature only for themselves and try it out before enabling it for the whole marketplace.

### DIFF
--- a/app/helpers/feature_flag_helper.rb
+++ b/app/helpers/feature_flag_helper.rb
@@ -31,13 +31,17 @@ module FeatureFlagHelper
   end
 
   def fetch_feature_flags(request, is_admin)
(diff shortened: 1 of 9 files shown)


lyyder:
Changing the current feature flags to community specific in [2326](https://github.com/sharetribe/sharetribe/pull/2326). Continuing with this when it's merged.

Gnito:
Check Travis

PR #2320, opened 2016-07-22
Fix CLP top-bar assets not linked to asset_host when configured

bladealslayer:
This happens, because we use `ActionController::Metal` and the
setting of `config.action_controller.asset_host` doesn't take effect.

Alternatively, we could load a bunch of extra modules and hooks, but it
seems an overkill, because there are lots of hooked hooks, it seems:

``` ruby
include AbstractController::AssetPaths
include ActionController::ParamsWrapper
ActiveSupport.run_load_hooks(:action_controller, self)
```

### DIFF
--- a/app/views/landing_page/landing_page.erb
+++ b/app/views/landing_page/landing_page.erb
@@ -129,9 +129,10 @@
         <%= javascripts[:translations] %>
     </script>
 
-    <%= stylesheet_link_tag 'app-bundle' %>
-    <%= javascript_include_tag 'vendor-bundle' %>
-    <%= javascript_include_tag 'app-bundle' %>
+    <% asset_host = local_assigns.key?(:asset_host) ? asset_host : nil %>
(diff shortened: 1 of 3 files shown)


app/views/landing_page/landing_page.erb
```diff
@@ -129,9 +129,10 @@
         <%= javascripts[:translations] %>
     </script>
 
-    <%= stylesheet_link_tag 'app-bundle' %>
-    <%= javascript_include_tag 'vendor-bundle' %>
-    <%= javascript_include_tag 'app-bundle' %>
+    <% asset_host = local_assigns.key?(:asset_host) ? asset_host : nil %>
```
lyyder: What happens if `asset_host` can not be found? Assets are served from rails?

bladealslayer: Yes, the link to the asset will be `/assets/something`, so served from the same domain as the app.

lyyder: Cool!

PR #2321, opened 2016-07-22
Topbar priority menu

Gnito:
Still needs IE check


client/app/components/composites/MenuPriority/MenuPriority.js
```diff
@@ -0,0 +1,166 @@
...
+    const withWidths = linksFromRenderedDiv.map((l) => {
+      const linkData = _.find(links, (link) => link.content === l.textContent);
+      const breakPoint = l.offsetLeft + l.offsetWidth + LINK_SPACING;
+      return Object.assign({}, linkData, { breakPoint });
+    });
+    return withWidths;
+  }
+
+  updateNav(links) {
+    const piorityMenu = document.querySelectorAll(`.${css.menuPriority}`)[0];
```
lyyder: Would this be same as `document.querySelector(`.${css.menuPriority}`)`

client/app/components/composites/MenuPriority/MenuPriority.js
```diff
@@ -0,0 +1,166 @@
...
+    const menuButton = document.querySelectorAll(`.${css.hiddenLinks}`)[0];
```
lyyder: Same here.

client/app/components/composites/MenuPriority/MenuPriority.js
```diff
@@ -0,0 +1,166 @@
...
+import r, { div, a } from 'r-dom';
+import _ from 'lodash';
+import classNames from 'classnames';
+import { canUseDOM } from '../../../utils/featureDetection';
+
+import Menu from '../Menu/Menu';
+import css from './MenuPriority.css';
+
+const EXTRA_SPACING_RIGHT = 24;
+const LINK_SPACING = 18;
```
kpuputti: If these values depend on something in CSS, they could be imported from variables.js, therefore using the same values as the CSS does.

client/app/components/composites/MenuPriority/MenuPriority.js
```diff
@@ -0,0 +1,166 @@
...
+    this.state = {
+      priorityWrapperWidth: '0px',
+      priorityLinks: this.links.filter((l) => l.priority >= 0),
+      hiddenLinks: this.links,
+    };
+  }
+
+  componentDidMount() {
+    if (canUseDOM) {
+      const that = this;
```
kpuputti: `this` within an arrow function refers to the same object as the `this` in the outer scope, so this is unnecessary as `that` is only used within arrow functions.

client/app/components/composites/MenuPriority/MenuPriority.js
```diff
@@ -0,0 +1,166 @@
...
+      // ComponentDidMount is called after React component is passed to DOM,
+      // but painting is not necessarily ready yet at that point
+      window.requestAnimationFrame(() => {
+        if (linksWithBreakPoints == null && menuIsVisible(priorityMenu)) {
+          linksWithBreakPoints = that.updateWidths(that.links);
+          that.updateNav(linksWithBreakPoints);
+        }
+      });
+
+      window.addEventListener('resize', () => {
```
kpuputti: This handler will potentially create tons of layout trashes since the `updateWidths` method queries the offset dimensions for each link, forcing a rerender for each iteration.

Let's keep an eye on the performance of this, and add some throttling etc. if things become unbearable.

kpuputti: There is no equivalent `removeEventHandler` for the resize event, that should definitely be added in the unmount lifecycle event.

client/app/components/sections/Topbar/Topbar.css
```diff
@@ -140,9 +140,9 @@
   }
 }
 
-.topbarSpacer {
+.topbarMenuSpacer {
   display: none;
-  margin: 0;
+  margin: 0 48px 0 0;
```
kpuputti: Magical mystery number 48px

config/locales/en.yml
```diff
@@ -2134,6 +2134,7 @@ en:
   web:
     topbar:
       menu: Menu
+      more: More
```
kpuputti: Is this translation needed still?

client/app/components/composites/MenuPriority/MenuPriority.js
Gnito: I think the reason here was that event listener didn't like it at all.

client/app/components/composites/MenuPriority/MenuPriority.js
Gnito: this.updateWidths is called only once per page load.
Unmount is good point!

client/app/components/sections/Topbar/Topbar.css
Gnito: It actually is, but I'll add that to variables.

config/locales/en.yml
Gnito: Yes.

client/app/components/sections/Topbar/Topbar.css
Gnito: Actually, I removed it altogether (I handle this with extraspacing)

PR #2325, opened 2016-07-25
Get rid of per-marketplace CSS compilation

bladealslayer:
Per marketplace CSS compilation is slow. This PR adds support for on-the-fly CSS overrides for all relevant styles.

## The approach

Due to use of SASS `extend` and `include`, it's hard to track which CSS selectors are ultimately affected by custom colors and images. This PR solves this by introducing a `rake` task, which is meant to be executed once by developers when changes are done to styles. The task generates a HAML template containing the correct style overrides, which is added to the repository and included on each page.

The algorithm is as follows:

1. Compile default application styles
1. Compile the stylesheet again with custom colors and images set to known values
1. Compute the actual CSS selectors affected by the customizations
1. Compute CSS color values for all color variations (darken, lighten) that are used in our .scss styles
1. Lookup all color values and replace colors and image URLs with HAML-style template strings
1. Generate a HAML template file, intended to be rendered on each page

At runtime:

1. Compute color values for used CSS color variations (lighten, darken)
2. Populate CSS overrides template with the required color values and image URLs
3. Render the template on each page after including the default stylesheet

**TODO:**
- [x] rake task for generating overrides template
- [x] render template
- [x] stop calling style rendering service when updates are done to `Community`
- [x] update tests
- [x] optimize generated template for size and speed
- [x] fix / workaround for loosing old attached image url
- [ ] ~~add migration to delete custom stylesheet-related columns~~ - in later PR


app/view_utils/marketplace_helper.rb
```diff
@@ -6,4 +6,22 @@ def google_maps_key(cid)
     community_key ? community_key : APP_CONFIG.google_maps_key
   end
 
+  def style_customizations_map(community)
+    color1 = community.custom_color1&.downcase || "a64c5d"
+    color2 = community.custom_color2&.downcase || community.custom_color1&.downcase || "00a26c"
+
```
lyyder: Where do these colors come from? Could they be read from config?

app/view_utils/marketplace_helper.rb
```diff
@@ -6,4 +6,22 @@ def google_maps_key(cid)
     community_key ? community_key : APP_CONFIG.google_maps_key
   end
 
+  def style_customizations_map(community)
+    color1 = community.custom_color1&.downcase || "a64c5d"
+    color2 = community.custom_color2&.downcase || community.custom_color1&.downcase || "00a26c"
```
lyyder: Where do these colors come from? Could they be read from config?

lyyder:
Brilliant stuff!! Looks good IMHO, @Gnito would you care to take a look also?

app/view_utils/marketplace_helper.rb
bladealslayer: Good point, they are currently in multiple places (I think). Have to check if it's easy to put them all in one place.

Gnito:
Yeah, looks good to me too!

PR #2326, opened 2016-07-26
Community-specific feature flags

lyyder:
This PR makes current feature flags community specific which paves the way for making feature flags user specific.

### DIFF
--- a/app/services/feature_flag_service/api/features.rb
+++ b/app/services/feature_flag_service/api/features.rb
@@ -6,29 +6,29 @@ module FeatureFlagService::API
       @feature_flag_store = feature_flag_store
     end
 
-    def enable(community_id:, features:)
+    def enable(entity_id:, features:)
       if features.blank?
         return Result::Error.new("You must specify one or more flags in #{@feature_flag_store.known_flags} to enable.")
       end
 
-      Result::Success.new(@feature_flag_store.enable(community_id, features))
+      Result::Success.new(@feature_flag_store.enable(entity_id, features))
     end
 
-    def disable(community_id:, features:)
+    def disable(entity_id:, features:)
       if features.blank?
         return Result::Error.new("You must specify one or more flags in #{@feature_flag_store.known_flags} to disable.")
       end
 
-      Result::Success.new(@feature_flag_store.disable(community_id, features))
+      Result::Success.new(@feature_flag_store.disable(entity_id, features))
     end
 
-    def get(community_id:)
-      Result::Success.new(@feature_flag_store.get(community_id))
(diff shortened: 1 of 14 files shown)


db/migrate/20160722115656_set_type_for_feature_flags.rb
```diff
@@ -0,0 +1,16 @@
+class SetTypeForFeatureFlags < ActiveRecord::Migration
+  def up
+    # Set the type of every feature flag to CommunityFeatureFlag
+    execute("
+      UPDATE feature_flags
+      SET type = 'CommunityFeatureFlag'
```
Gnito: I'm not sure we have used CamelCase in Rails db

app/services/feature_flag_service/feature_flags_api.md
```diff
@@ -0,0 +1,223 @@
...
+# feature_flags/v1/user
+
+## GET /:user_id
+
+Request body: empty
+
+Response body:
+
+```ruby
+{ user_id: 123
```
Gnito: person_id ?

app/services/feature_flag_service/api/features.rb
```diff
@@ -6,29 +6,29 @@ def initialize(feature_flag_store)
       @feature_flag_store = feature_flag_store
     end
 
-    def enable(community_id:, features:)
+    def enable(entity_id:, features:)
```
Gnito: As you said earlier, it's also possible to make this so that enable takes hash with keys: community_id, person_id, and features. So, it's possible to make some kind of per flag-service-type check which accepts only parameters community_id + features or person_id + features combos (depending on service type).

app/services/feature_flag_service/feature_flags_api.md
lyyder: Original intention here was maybe to describe a HTTP API that we could achieve by building a controller layer on top of this service. But from API driven perspective maybe it would be better to use the actual param names.

app/services/feature_flag_service/api/features.rb
lyyder: Agreed, this after all might be a clearer solution.

lyyder:
Discussed this with @ovan and decided to add `community_id` also to user-specific feature flags. Therefore, ditching single table inheritance and adding `community_id` and `person_id` to all API requests.

lyyder:
As the community_id is included in user-specific feature flags, the need for community related changes becomes rather small so I'll close this PR continue the work for user-specific changes in [2313](https://github.com/sharetribe/sharetribe/pull/2313).

PR #2331, opened 2016-07-26
Revert "Revert "Topbar style fixes""

Gnito:
Reverts sharetribe/sharetribe#2330
screen shot 2016-07-26 at 20 39 40

### DIFF
--- a/client/app/assets/styles/variables.js
+++ b/client/app/assets/styles/variables.js
@@ -111,6 +111,7 @@ module.exports = {
   '--Topbar_tabletHeight': '60px',
   '--Topbar_mobileHeight': '50px',
   '--Topbar_padding': `0 ${bodyPadding}`,
+  '--Topbar_paddingLanguageMenuVertical': '27.5px',
   '--Topbar_tabletPadding': `0 ${bodyPadding} 0 6px`,
   '--Topbar_mobilePadding': '0 1px 0 0',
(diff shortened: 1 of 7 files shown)


client/app/components/composites/Menu/Menu.css
```diff
@@ -35,7 +35,7 @@
   padding: var(--Menu_labelPaddingVertical) 0;
   color: var(--Menu_textColor);
   font-size: var(--Menu_fontSize);
-  line-height: 1.5em;
+  line-height: 1em;
```
Gnito: Line-height 1.5em aka 1.5 * 14px => 23px. 23 + 2*padding => odd number. 
It won't work well in topbar where height is even 80px.

client/app/assets/styles/variables.js
```diff
@@ -111,6 +111,7 @@ module.exports = {
   '--Topbar_tabletHeight': '60px',
   '--Topbar_mobileHeight': '50px',
   '--Topbar_padding': `0 ${bodyPadding}`,
+  '--Topbar_paddingLanguageMenuVertical': '27.5px',
```
Gnito: Language menu's font-size (13px) + 2 * --Topbar_paddingLanguageMenuVertical must be 70px (to be on same level as other dropdowns).

PR #2337, opened 2016-07-28
customizations: only current locales matter

Gnito:


app/services/admin/onboarding_wizard.rb
```diff
@@ -136,7 +136,8 @@ def load_setup_steps(community_id)
 
     def init_setup_steps(community_id)
       community = Community.find(community_id)
-      community_customizations = CommunityCustomization.where(community_id: community.id)
+      locales = community.locales
+      community_customizations = CommunityCustomization.where(community_id: community.id).select { |cc| locales.include?(cc.locale) }
```
lyyder: Does this change apply to the problems that have been encountered with other steps of the onboarding wizard, like custom fields.

Gnito: This slogan & description is the only bug I found from the example marketplace - I don't know if there's more bugs with different setup steps

PR #2349, opened 2016-08-01
Fix configurations api: nil values should'n affect updates

Gnito:


app/services/marketplace_service/api/configurations.rb
```diff
@@ -14,13 +14,21 @@ def get(community_id:)
...
+    # update({
+    #   community_id: 1,                # <community_id>,
+    #   configurations: {
+    #     main_search: :keyword,        # optional, one of: :keyword, :location, :keyword_and_location
+    #     distance_unit: :metric,       # optional, one of: :metric, :imperial
+    #     limit_search_distance: false, # optional, boolean
+    #     limit_priority_links: 4,      # optional, -1..5
+    #   }
+    # })
+    def update(community_id:, configurations:)
```
lyyder: Could the configurations hash have a default value: `def update(community_id:, configurations: {})`

Gnito: I thought it would be better if the call fails if it is called without any configuration hash

lyyder: Cool, fair enough.

Gnito: I added a type check for configurations. (The insides of that hash is checked in store layer)

PR #2354, opened 2016-08-01
Fix GA event sending about created listings

gnomet:
We noted with Vesa that the handler where Google Analytics event sending was, was only used when image uploader was still uploading, so now made another handler to be used in positive case, and put the GA event there.


Gnito:
👍

PR #2358, opened 2016-08-02
Reset cleanup strategy after executing a :no-transaction test

bladealslayer:
This resolves a rare bug where DatabaseCleaner is initialized with :truncation strategy
in a before(:all) block, if previous suite had :no-transaction set to true.

### DIFF
--- a/spec/support/database_cleanup.rb
+++ b/spec/support/database_cleanup.rb
@@ -29,6 +29,7 @@ RSpec.configure do |config|
 
   config.after(:each, :'no-transaction' => true) do
     clean_db.call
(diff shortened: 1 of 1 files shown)


spec/support/database_cleanup.rb
```diff
@@ -29,6 +29,7 @@
 
   config.after(:each, :'no-transaction' => true) do
     clean_db.call
+    DatabaseCleaner.strategy = :transaction
```
lyyder: Should the strategy be set before the `clean_db.call` is executed?

lyyder: Oh, the PR title pretty much explains it.

PR #2360, opened 2016-08-02
Add support for running tests with CircleCI

bladealslayer:


features/step_definitions/general_steps.rb
```diff
@@ -39,7 +39,7 @@
     with_scope(selector) do
       # nothing to do here, just try to search the selector and should fail on that
     end
-  }.should raise_error(Capybara::ElementNotFound)
+  }.should raise_error(RSpec::Expectations::ExpectationNotMetError)
```
lyyder: Why is the exception type changed here?

bladealslayer: Because with_scope was changed to first `expect` the selector, so now the exception comes from RSpec. However, I now see that there's a much cleaner way to write this test, so pushing update.

PR #2361, opened 2016-08-02
Slight copy tweaks

vsaarinen:
Some improvements to the onboarding copy. Also replaced references to "our" with "Sharetribe".

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -131,7 +131,7 @@ en:
         sender_address: "Sender address: %{sender_address}"
         sender_address_default: "Sender address: not set (using default address %{sender_address})"
         need_to_change: "If you need to change the email address, please %{contact_support_link}."
-        contact_support_link_text: "contact our support"
+        contact_support_link_text: "contact Sharetribe support"
         set_sender_address: "Set sender address"
         sender_name_label: Name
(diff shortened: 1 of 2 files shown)


vsaarinen:
Note: Now changed "Desktop logo" to "Logo" in the design settings page.

lyyder:
Should these come from WTI?

config/locales/en.yml
```diff
@@ -1995,7 +1995,7 @@ en:
       visible_to_registered_users: "(visible to users that have logged in)"
       default_in_listing: "(used only as a default when creating a new listing)"
       invisible: "(not shown to other users)"
-      image_is_processing: "We are processing your profile picture, it will be visible in a few seconds."
+      image_is_processing: "Hang on tight. We are processing your profile picture. It will be ready in a minute or two."
```
lyyder: Is the profile picture processing time more like minutes than seconds?

vsaarinen: Profile picture processing was quick, but processing a front page cover image took a minute or two for me. Thought it would be better to underpromise and overdeliver here, just in case it does sometimes take longer.

lyyder: Ok, cool.

PR #2368, opened 2016-08-03
Hide search from homepage main when using new topbar

Gnito:
Short paddings:
screen shot 2016-08-04 at 01 33 16
vs
expanded padding-top:
screen shot 2016-08-04 at 01 31 36

The latter takes same height as narrow cover image in other pages:
screen shot 2016-08-04 at 01 32 34

### DIFF
--- a/app/views/homepage/index.haml
+++ b/app/views/homepage/index.haml
@@ -23,12 +23,12 @@
         %a.inline-big-button{href: sign_up_path}
           .content
             = t("layouts.application.connect")
-      - else
+      - elsif !FeatureFlagHelper.feature_enabled?(:topbar_v1)
         - if(location_search_in_use)
           = render partial: "location_bar"
         - else
           = render partial: "search_bar"
-  - else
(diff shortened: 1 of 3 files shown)


app/views/homepage/index.haml
```diff
@@ -15,20 +15,22 @@
...
     %p= community_description.html_safe
-    .search-form
-      - if @current_community.private?
-        %a.inline-big-button{href: sign_up_path}
-          .content
-            = t("layouts.application.connect")
-      - else
-        - if(location_search_in_use)
-          = render partial: "location_bar"
+    - if !(is_homepage && !@current_community.private? && FeatureFlagHelper.feature_enabled?(:topbar_v1))
```
lyyder: Could `unless` be used instead of `if !`?

Gnito: Yes, but I rather keep if - else structures in if mode.

lyyder: But does this `if` have a matching `else`?

Gnito: Ah, good point - I was a bit blind here.

PR #2371, opened 2016-08-05
Allow more precise icon styling in CLP

bladealslayer:


app/views/landing_page/_info.erb
```diff
@@ -68,6 +68,12 @@
       .<%= section_id %>__info__column-icon > svg path  {
         stroke: rgb(<%= s["icon_color"]["value"].join(",") %>);
       }
+      .<%= section_id %>__info__column-icon > svg .no-stroke {
+        stroke: none;
+      }
+      .<%= section_id %>__info__column-icon > svg .fill {
```
kpuputti: The class names `.no-stroke` and `.fill` are a bit too  short and generic IMHO, and a similar class can easily come from other styles, resulting in weird styles appearing. Could they be something like `.icon-no-stroke` and `.icon-custom-color-fill`?

PR #2373, opened 2016-08-05
Fixed typo

mathijsdelangen:


thomasmalbaux:
Hey Mathijs,

thanks a lot for reporting and your contribution!

We actually use an online translation tool to translate languages in Sharetribe, instead of editing manually locale files. 
I've followed your suggestion and already made the change at https://webtranslateit.com/en/projects/2056-Sharetribe/locales/en..nl/strings/5217051

It should be deployed soon and it will be applied in this nl.yml file too.

Thanks again!

PR #2376, opened 2016-08-08
Send location search scale and offset parameters separately

mporkola:


app/controllers/homepage_controller.rb
```diff
@@ -288,18 +289,18 @@ def search_coordinates(latlng)
...
+  def location_search_params(latlng, distance_unit, distance, minimum_distance, scale_multiplier, offset_multiplier, limit_by_distance)
     # Current map doesn't react to zoom & panning, so we fetch all the results as before.
     if @view_type != 'map'
       Maybe(latlng)
         .map {
-          distance = [minimum_distance_max, distance_max.to_f].max
+          distance = [minimum_distance, distance.to_f].max
           distance_limit = distance if limit_by_distance
-          scale_or_distance = scale ? { scale: distance * scale } : { sort: :distance }
+          scale_or_sort = scale_multiplier && offset_multiplier ? { scale: distance * scale_multiplier, offset: offset_multiplier } : { sort: :distance }
```
Gnito: Maybe this should be something like
```
{ scale: [minimum_distance_for_scale, distance.to_f].max * scale_multiplier, offset: offset_multiplier }
```
or even 
```
{ scale: [minimum_distance_for_scale, distance.to_f].max * scale_multiplier, offset: [minimum_distance_for_offset, distance.to_f].max * offset_multiplier }
```

app/controllers/homepage_controller.rb
```diff
@@ -288,18 +289,18 @@ def search_coordinates(latlng)
...
   end
 
-  def location_search_params(latlng, distance_unit, distance_max, minimum_distance_max, scale, limit_by_distance)
+  def location_search_params(latlng, distance_unit, distance, minimum_distance, scale_multiplier, offset_multiplier, limit_by_distance)
     # Current map doesn't react to zoom & panning, so we fetch all the results as before.
     if @view_type != 'map'
       Maybe(latlng)
         .map {
-          distance = [minimum_distance_max, distance_max.to_f].max
+          distance = [minimum_distance, distance.to_f].max
```
Gnito: distance (_for_distance_limit) is now shadowing the original distance in parameters (i.e. viewport radius)

mporkola: Actually that was my purpose, but for clarity it's probably better to name it differently

app/controllers/homepage_controller.rb
```diff
@@ -177,8 +162,42 @@ def find_listings(params, listings_per_page, filter_params, includes, location_s
...
+
+
+    # location search params
+    distance = params[:distance_max].to_f
+    distance_system = marketplace_configuration ? marketplace_configuration[:distance_unit] : nil
+    distance_unit = (location_search_in_use && distance_system == :metric) ? :km : :miles
+    limit_search_distance = marketplace_configuration ? marketplace_configuration[:limit_search_distance] : true
+    distance_limit = [distance, APP_CONFIG[:external_search_distance_limit_min]].max if limit_search_distance
+
+    if @view_type != 'map' && params[:lc]
```
Gnito: location_search_in_use might be better here since it's already available (its a check for params[:lc] + search mode (:location && :keyword_and_location)

PR #2377, opened 2016-08-09
Removing the square around the pear icon for the CLP.

Luis-RG:
This just removes the square around one icon on the CLP.


lyyder:
Looks good, remember to test in staging!

PR #2379, opened 2016-08-09
Fix priority menu render blinking

kpuputti:
The menu is rendered minimised in the server side since the viewport size is not known, and then expanded in the client side when the space to render items is known. However, this results in first having the minimized menu visible and then swapping the expanded menu to replace it, which causes a frustrating blink on every page load.

The fix hides the menu when JavaScript is enabled (thus having the menu working when JS is disabled etc.) until the menu is expanded.

(built on top of topbar-search-submit, only the latest commit is relevant to this PR)


client/app/components/composites/MenuPriority/MenuPriority.css
```diff
@@ -1,8 +1,18 @@
...
+  @nest :global(.js-enabled) & {
+    visibility: hidden;
+
+    &.isMeasured {
+      visibility: visible;
+    }
+  }
 }
 
+.isMeasured {}
```
Gnito: Doesn't this need: /* stylelint-disable-line block-no-empty */

kpuputti: Yes, it should. I thought I ran the linter without errors, but that clearly shouldn't pass with our config. Thanks for noticing.

PR #2380, opened 2016-08-09
Merged ff get

lyyder:
Change feature flag service API so that `get` returns an combined set of person and community-specific features. For fetching person or community-specific feature flags specifically, there are `get_by_person_id` and `get_by_community_id` respectively.

Also `FeatureFlagService::API::Api.features.enabled?` is removed as it is never invoked.


Gnito:
At this point, looks good to me

PR #2389, opened 2016-08-11
Fix search button shadow leaking

mporkola:
Before:
screenshot 2016-08-11 15 29 58

After:
screenshot 2016-08-11 15 29 40


lyyder:
I think there is a slight _block_ visible under the shadow next to the spotlight box:
image

mporkola:
screenshot 2016-08-11 15 59 30

PR #2390, opened 2016-08-11
Fix missing locale string validation for omniauth callbacks

bladealslayer:


app/controllers/sessions_controller.rb
```diff
@@ -170,4 +172,13 @@ def terms_accepted?(user, community)
     user && community.consent.eql?(user.consent)
   end
 
+  def get_origin_locale
```
ovan: This should take request as input. The style to follow is to restrict access to top level globals to controller actions but have them passed to helper methods explicitly. This helps with both readability and helper method reusability/refactorability.

app/utils/url_utils.rb
```diff
@@ -35,6 +35,7 @@ def build_url(base_url, params)
   #
   # Returns the first "folder" in path. Does not validate the locale
   def extract_locale_from_url(url)
+    binding.pry
```
ovan: :)

PR #2394, opened 2016-08-12
Only allow valid coordinates to search by location

kpuputti:


mporkola:
Regex101.com says yes, so do I

Gnito:
Regexper looks nice : https://regexper.com/#%2F%5E-%3F%5Cd%2B(%3F%3A%5C.%5Cd%2B)%3F%2C-%3F%5Cd%2B(%3F%3A%5C.%5Cd%2B)%3F%24%2F

PR #2395, opened 2016-08-12
Adding Slovak to the list of unsupported languages.

Luis-RG:
Nothing more to add than what the title says.


thomasmalbaux:
@Luis-RG
Language name is not correct (now `slovenčina`, which is a copy paste from the above row ;-))
Also tests failed so good to check them.

mporkola:
The test failure is my bad, fixed in other commit

PR #2397, opened 2016-08-12
Add missing location search placholder back

kpuputti:


mporkola:
👍 even though [Object object] really gives the 2016 vibe

PR #2404, opened 2016-08-15
Update NPM packages

kpuputti:
- Update NPM packages
- Update linting config
- Fix new linting errors


client/app/assets/styles/variables.js
```diff
@@ -1,4 +1,5 @@
 /* eslint-env node */
+/* eslint-disable no-magic-numbers */
```
Gnito: :+1:

PR #2408, opened 2016-08-16
Disable Checkout Payment gateway

rap1ds:
This commit changes the communities that are using Checkout payment gateway in following ways:

1. The "transaction process" (i.e. free/postpay/preauthorize) is changed from `postpay` to `free` for all listings that are currently using `postpay` process
2. The transaction process is changed from `postpay` to `free` for all listing shapes (i.e. order types) that are currently using `postpay` process
3. The transaction process `postpay` is removed
4. The Checkout payment gateway is removed from the community.

So in practice these four changes will change the marketplace from Checkout marketplace to "free" marketplace, where all the transactions will go the "free" flow.

This commit doesn't try to migrate transaction that are ongoing (i.e. in pending/paid state). Accessing and changing the state of such transaction will probably cause an error.

### DIFF
--- a/db/migrate/20160816083020_migrate_checkout_postpay_listings_to_free.rb
+++ b/db/migrate/20160816083020_migrate_checkout_postpay_listings_to_free.rb
@@ -1,0 +1,18 @@
+class MigrateCheckoutPostpayListingsToFree < ActiveRecord::Migration
+  def up
+    # 1) Find all listings where transaction process is "postpay" and payment gateway is "Checkout"
+    # 2) Change the listings' transaction process to "none"
+
+    execute("
+      UPDATE listings l
+      LEFT JOIN transaction_processes txp_current ON (txp_current.id = l.transaction_process_id)
+      LEFT JOIN payment_gateways pg ON (pg.community_id = l.community_id)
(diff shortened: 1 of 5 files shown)


db/migrate/20160816083020_migrate_checkout_postpay_listings_to_free.rb
```diff
@@ -0,0 +1,18 @@
+class MigrateCheckoutPostpayListingsToFree < ActiveRecord::Migration
+  def up
+    # 1) Find all listings where transaction process is "postpay" and payment gateway is "Checkout"
+    # 2) Change the listings' transaction process to "none"
+
+    execute("
+      UPDATE listings l
+      LEFT JOIN transaction_processes txp_current ON (txp_current.id = l.transaction_process_id)
+      LEFT JOIN payment_gateways pg ON (pg.community_id = l.community_id)
+      SET l.transaction_process_id = (SELECT id FROM transaction_processes WHERE community_id = l.community_id AND process = 'none' AND author_is_seller = 1)
```
ovan: Does every community have a 'none' process? That's what all of these seem to rely on.

rap1ds: Yes, duoble checked.

PR #2423, opened 2016-08-18
Disable Braintree payments

kpuputti:
This PR disables Braintree by changing the preauthorize and postpay payment flows into free payment flows.

The PR doesn't disable Braintree entirely: If there's an existing non-finished transaction with Braintree, it can be finished with Braintree. The main purpose of the PR is to ensure that there will be **no new Braintree transactions**. This is to first step for disabling Braintree entirely.

In order to rollback the changes, this PR adds a new column `old_process` to the `transaction_processes` table. That column will be removed in the future releases.

### DIFF
--- a/db/migrate/20160818111724_change_braintree_preauthorization_processes_to_free.rb
+++ b/db/migrate/20160818111724_change_braintree_preauthorization_processes_to_free.rb
@@ -1,0 +1,18 @@
+class ChangeBraintreePreauthorizationProcessesToFree < ActiveRecord::Migration
+  def up
+    # Change all preauthorize processes to none unless they are in use with Paypal gateway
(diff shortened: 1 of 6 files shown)


db/migrate/20160818111724_change_braintree_preauthorization_processes_to_free.rb
```diff
@@ -0,0 +1,18 @@
+class ChangeBraintreePreauthorizationProcessesToFree < ActiveRecord::Migration
+  def up
+    # Change all preauthorize processes to none unless they are in use with Paypal gateway
+    execute "UPDATE transaction_processes tp
+             SET process = 'none'
```
ovan: `SET tp.process = 'none'` ?

PR #2424, opened 2016-08-18
Remove Braintree

kpuputti:
## TODO

 - [x] Restore Braintree translations that are using by PayPal
 - [x] Grep :paypal and check all if/else branches and possibly raise exceptions in else branches
 - [x] Remove VAT (communities.vat column and maybe some related code?)
 - [x] Remove PaymentsController
 - [x] Remove PostPayTransactionController
 - [x] Grep "post_pay" and "PostPay"
 - [x] Remove postpay steps from the statemachine
 - [x] Grep "escrow"
 - [x] Grep "invoice"
 - [x] Grep and remove "payment_reminder"
 - [x] Grep "community.payment_gateway"
 - [x] Remove transaction has_one payment
 - [x] Remove community.has_one payment_gateway
 - [x] Remove testimonials_in_use code


app/controllers/people_controller.rb
```diff
@@ -241,12 +241,10 @@ def update
           :email_when_conversation_accepted,
           :email_when_conversation_rejected,
           :email_about_new_received_testimonials,
-          :email_about_accept_reminders,
           :email_about_confirm_reminders,
           :email_about_testimonial_reminders,
           :email_about_completed_transactions,
           :email_about_new_payments,
-          :email_about_payment_reminders,
```
ovan: is this only about postpay too? There's payments in every flow, right?

ovan: No wait, I get it, in prepay there's no tx process without payment.

app/controllers/preauthorize_transactions_controller.rb
```diff
@@ -260,96 +239,13 @@ def booked
...
-
-      unless transaction_response[:success]
-        flash[:error] = "An error occured while trying to create a new transaction: #{transaction_response[:error_msg]}"
-        return redirect_to action: :preauthorize
-      end
-
-      transaction_id = transaction_response[:data][:transaction][:id]
-
-      redirect_to person_transaction_path(:person_id => @current_user.id, :id => transaction_id)
+    if (transaction_response[:data][:gateway_fields][:redirect_url])
```
ovan: If transaction_reponse is a Result then shouldn't this be done with on_success instead?

app/services/transaction_service/transaction.rb
```diff
@@ -6,19 +6,21 @@ module TransactionService::Transaction
...
   GATEWAY_ADAPTERS = {
     paypal: TransactionService::Gateway::PaypalAdapter.new,
-    braintree: TransactionService::Gateway::BraintreeAdapter.new,
-    none: TransactionService::Gateway::FreeAdapter.new
+    none: TransactionService::Gateway::FreeAdapter.new,
+
+    # Deprecated adapters.
+    # We need these to be here, otherwise reading the old transactions will fail
+    checkout: TransactionService::Gateway::FreeAdapter.new,
+    braintree: TransactionService::Gateway::FreeAdapter.new,
```
ovan: Did you think about solving this at the origin of the failure instead of providing fake adapters?

PR #2425, opened 2016-08-18
Make new top bar cacheable

mporkola:
- [x] don't use MarketplaceContext
- [x] cache with `props` as key
- [x] test on custom landing page (&fix)

### DIFF
--- a/app/views/layouts/application.haml
+++ b/app/views/layouts/application.haml
@@ -24,7 +24,8 @@
     - if(@view_type.eql?("map"))
       = javascript_include_tag "markerclusterer.js"
     - props = topbar_props
-    = react_component("TopbarApp", props: props, prerender: true)
+    - cache props do
+      = react_component("TopbarApp", props: props, prerender: true, id: 'topbar-container')
   - else
     = render partial: 'layouts/global_header', locals: header_props()
 
--- a/client/app/components/sections/Topbar/Topbar.js
+++ b/client/app/components/sections/Topbar/Topbar.js
@@ -5,7 +5,7 @@ import * as placesUtils from '../../../utils/places';
 import * as urlUtils from '../../../utils/url';
 
 import { t } from '../../../utils/i18n';
-import { routes as routesProp, marketplaceContext } from '../../../utils/PropTypes';
+import { routes as routesProp } from '../../../utils/PropTypes';
 import css from './Topbar.css';
 import styleVariables from '../../../assets/styles/variables';
 
@@ -154,7 +154,8 @@ const createQuery = (searchParams, queryString) => {
 
 class Topbar extends Component {
   render() {
-    const { location, marketplace_color1, loggedInUsername } = { ...DEFAULT_CONTEXT, ...this.props.marketplaceContext };
+    const { location, marketplace_color1: marketplaceColor1 } = { ...DEFAULT_CONTEXT, ...this.props.marketplace };
+    const { loggedInUsername } = this.props.user || {};
 
     const menuProps = this.props.menu ?
       Object.assign({}, this.props.menu, {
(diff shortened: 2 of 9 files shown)


client/app/utils/url.js
```diff
@@ -69,6 +69,7 @@ const parseSearchQueryParams = function parseSearchQueryParams(location, restric
  */
 const currySearchParams = function currySearchParams(restrict_to_params) {
   return function curryWrap(location) {
+    console.log("PARSING: ", location);
```
kpuputti: Remember to remove this.

app/views/layouts/application.haml
```diff
@@ -24,7 +24,8 @@
     - if(@view_type.eql?("map"))
       = javascript_include_tag "markerclusterer.js"
     - props = topbar_props
-    = react_component("TopbarApp", props: props, prerender: true)
+    - cache props do
```
ovan: Doesn't this mean the props now contain information about the currently logged in user and that ends up in the cache + rendered at server side too? Wasn't the plan to only render the user agnostic version at server side and add with user rendering client side?

client/app/startup/TopbarApp.js
```diff
@@ -20,6 +20,6 @@ export default (props, marketplaceContext) => {
     'sign_up',
   ], { locale });
 
-  const combinedProps = Object.assign({}, props, { marketplaceContext, routes });
+  const combinedProps = Object.assign({}, props, { routes });
```
ovan: I'm not sure but this change might break the landing page topbar dynamic rendering because it passes in the marketplaceContext too?

app/views/layouts/application.haml
mporkola: That's correct. We thought this is a good starting point, so that we can deploy this and implement the async personal data fetching later. Initially we thought that it can be accomplished without an additional http request by just rendering the user specific things on client side, but as the data for client side rendering is also output by the `react_component` call, it gets cached as well and doesn't work any better.

client/app/startup/TopbarApp.js
mporkola: Actually I tested this and it works, but I probably should remove the marketplace parameter from the call on CLP side

lyyder:
Even though the marketplace object is removed from topbar props here:
https://github.com/sharetribe/sharetribe/pull/2425/commits/d8c497a863f662f6e3e7cd34dc99780361f30548#diff-dffd979ece751a3641d088c9ec8aa160R52
, it is still passed through on server-side rendered non-landing pages and therefore there's a distinct cache entry for each path.

lyyder:
Answering my comment, `location` is passed in the `marketplace` hash and it is required for rendering the selected link at least, so it is required in the key.

client/app/components/sections/Topbar/Topbar.js
```diff
@@ -154,7 +154,8 @@ const createQuery = (searchParams, queryString) => {
 
 class Topbar extends Component {
   render() {
-    const { location, marketplace_color1, loggedInUsername } = { ...DEFAULT_CONTEXT, ...this.props.marketplaceContext };
+    const { location, marketplace_color1 } = { ...DEFAULT_CONTEXT, ...this.props.marketplace };
```
Gnito: I suggest renaming too:
```const { location, marketplace_color1: marketplaceColor1 }````

https://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&experimental=true&loose=false&spec=false&code=const%20%7B%20location%2C%20marketplace_color1%3A%20marketplaceColor1%20%7D%20%3D%20%7B%20location%3A%20'somedomain.com%2Fpath%2F'%2C%20marketplace_color1%3A%20'%23FF0000'%20%7D%0Aconsole.log(marketplaceColor1)%3B

PR #2427, opened 2016-08-19
Make external links open in new tab

lyyder:
Links that direct to a different domain are opened in a new tab.

### DIFF
--- a/app/controllers/landing_page_controller.rb
+++ b/app/controllers/landing_page_controller.rb
@@ -275,7 +275,8 @@ class LandingPageController < ActionController::Metal
       path_after_locale_change: path,
       search_placeholder: community_customization&.search_placeholder,
       locale_param: locale_param,
-      landing_page: landing_page)
+      landing_page: landing_page,
+      host_with_port: request.host_with_port)
   end
 
   # This is copied from the React on Rails source with our own rails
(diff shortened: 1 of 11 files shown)


lyyder:
Currently only menu links added by admin are considered.

app/controllers/landing_page_controller.rb
```diff
@@ -275,7 +275,8 @@ def topbar_props(community, community_customization, request_path, locale_param,
       path_after_locale_change: path,
       search_placeholder: community_customization&.search_placeholder,
       locale_param: locale_param,
-      landing_page: landing_page)
+      landing_page: landing_page,
+      host_with_port: request.host_with_port)
```
Gnito: You could probably use "railsContext" Location" too.

app/view_utils/topbar_helper.rb
```diff
@@ -127,4 +129,8 @@ def missing_profile_image_path
   def paths
     @_url_herlpers ||= Rails.application.routes.url_helpers
   end
+
+  def link_external?(url, host_with_port)
```
Gnito: how about ```is_link_external?``` or ```external_link?```

lyyder: Yeah, but `FeatureFlagHelper` has `feature_enabled?` which takes a feature and `Email` has `email_available?` which takes an email so I planned to follow the naming on those.

Gnito: Roger

PR #2428, opened 2016-08-19
Update React on Rails to version 6.0.5

mporkola:
- [x] Update dependencies
- [x] Update config to match
- [x] Deploy
- [x] Test extensively


rap1ds:
Looks cool! @mporkola have you checked the https://github.com/sharetribe/sharetribe/blob/master/lib/tasks/assets.rake file? Do we need it anymore? According to the React on Rails CHANGELOG, that file shouldn't be needed anymore. On the other hand, that's the place where we currently export the routes and translations.

mporkola:
Yes, we still need it because of the translations and routes. I thought migrating their generation to some other format isn't worth the hassle.

mporkola:
tested:
```
[x] deploy to staging
  [x] works on clp
  [x] works on R&A
[x] hot reloading:
  [x] CSS
  [x] JS
[x] local Procfile.static: builds assets after clobber, updates after js & css changes
  [x] landing page
  [x] R&A
[x] local Procfile.hot: builds assets after clobber, updates after js & css changes
  [x] landing page
  [x] R&A
```

config/initializers/react_on_rails.rb
```diff
@@ -18,10 +18,21 @@ def self.custom_context(view_context)
...
   config.server_bundle_js_file = "server-bundle.js"
 
+  # If you are using the ReactOnRails::TestHelper.configure_rspec_to_compile_assets(config)
+  # with rspec then this controls what npm command is run
+  # to automatically refresh your webpack assets on every test run.
+  config.npm_build_test_command = "npm run build:test"
+
+  # This configures the script to run to build the production assets by webpack. Set this to nil
+  # if you don't want react_on_rails building this file for you.
+  config.npm_build_production_command = nil #"npm run build:production"
```
Gnito: Mental note: if and when we start to use "npm run build:production", we need to check what can be done to duplicate asset handling in our webpack plugin: customfileloader

Gnito:
Looks good to me.
Seems to generate assets the old way (through customfileloader webpack plugin), which is good at this point. :+1:

PR #2440, opened 2016-08-23
Update Readme.md

pcm211:
the above changes added would be useful to compile Sharetribe for Production Mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,20 +268,10 @@ Before starting these steps, perform [steps 1-5 from above](#setting-up-the-deve
 1. Precompile the assets:
 
   ```bash
-  bundle exec rake assets:precompile
-  ```
-
- 1. Precompile the assets for Production Environment (else you may get squares instead of icons and other linking problems):
-
-  ```bash
   RAILS_ENV=production rake assets:precompile
   ```
 
-1. You may need to use the following cmd after the above cmd
-  ```bash
-  rm -rf public/assets
-  ```
- 
+
 1. Invoke the delayed job worker:
 
   ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -271,6 +271,17 @@ Before starting these steps, perform [steps 1-5 from above](#setting-up-the-deve
   bundle exec rake assets:precompile
   ```
 
+ 1. Precompile the assets for Production Environment (else you may get squares instead of icons and other linking problems):
+
+  ```bash
+  RAILS_ENV=production rake assets:precompile
```
rap1ds: Good catch! `rake assets:precompile` should have the `RAILS_ENV=production` environment set.

However, the `rake assets:precompile` is now there twice. This is unnecessary. Instead of adding a new step, could you just add the `RAILS_ENV=production` to the previous step?

README.md
```diff
@@ -271,6 +271,17 @@ Before starting these steps, perform [steps 1-5 from above](#setting-up-the-deve
...
+  ```
+
+1. You may need to use the following cmd after the above cmd
+  ```bash
+  rm -rf public/assets
```
rap1ds: I don't understand this. Why you think this is needed? You say that it _may_ be needed? When is it needed?

The purpose of the previous command `rake assets:precompile` is to generate precompiled files to `public/assets` directory, so removing that directory right after genereting the files doesn't make sense to me.

README.md
pcm211: @rap1ds   ya it was a typo error

README.md
pcm211: I would branch this out subsequently ...its reqd for users using S3 to load assets ...lightning fast speeds achievable by this ...

rap1ds:
Thanks for making the changes!

PR #2442, opened 2016-08-23
PayPal Capybara tests

rap1ds:
This PR adds an integration test for the full Paypal payment flow using the Fakepal payments implementation. Instead of writing the test as steps in a feature file, the whole test is written in the `paypal_steps.rb` file.

Test helpers are organised in a few directories under `features/support/feature_tests/`:

 - `action/`: UI flows than might span several pages
 - `page/`: abstracted selectors and actions for separate pages
 - `section/`: abstacted selectors and actions for UI components

### DIFF
--- a/features/paypal/paypal.feature
+++ b/features/paypal/paypal.feature
@@ -1,0 +1,5 @@
+Feature: Transaction with PayPal
+
+  @javascript
+  Scenario: Test transaction with PayPal
+    Then I expect transaction with PayPal test to pass
--- a/features/step_definitions/paypal_steps.rb
+++ b/features/step_definitions/paypal_steps.rb
@@ -21,12 +21,12 @@ Then("I expect transaction with PayPal test to pass") do
   paypal_actions.connect_seller_paypal
 
   # Add new listing
-  listing_actions.add_new_listing("Lörem ipsum")
+  listing_actions.add_new_listing("Snowman for sale: ☃")
   onboarding_wizard.dismiss_dialog
 
   # Member buys the listing
   login.logout_and_login_as(member[:username], member[:password])
-  paypal_actions.request_listing("Lörem ipsum")
+  paypal_actions.request_listing("Snowman for sale: ☃")
 
   # Adming accepts request
   login.logout_and_login_as(admin[:username], admin[:password])
--- a/features/step_definitions/paypal_steps.rb
+++ b/features/step_definitions/paypal_steps.rb
@@ -1,0 +1,48 @@
+# coding: utf-8
+
+def dismiss_onboarding_wizard_dialog
+  expect(page).to have_content("Woohoo, task completed!")
+  page.click_on("I'll do it later, thanks")
+end
+
+Then("I expect transaction with PayPal test to pass") do
+  navigation = FeatureTests::Navigation
+  data = FeatureTests::Data
+  login = FeatureTests::Action::Login
+  listing_actions = FeatureTests::Action::Listing
+  paypal_actions = FeatureTests::Action::Paypal
+
+  marketplace = data.create_marketplace(payment_gateway: :paypal)
+  admin = data.create_member(username: "paypal_admin", marketplace_id: marketplace[:id], admin: true)
+  member = data.create_member(username: "paypal_buyer", marketplace_id: marketplace[:id], admin: false)
+
+  navigation.navigate_in_marketplace!(ident: marketplace[:ident])
+
+  # Connect Paypal for marketplace and seller
+  login.login_as(admin[:username], admin[:password])
+  paypal_actions.connect_marketplace_paypal
+  dismiss_onboarding_wizard_dialog
+  paypal_actions.connect_seller_paypal
+
+  # Add new listing
+  listing_actions.add_new_listing("Lörem ipsum")
+  dismiss_onboarding_wizard_dialog
+
+  # Member buys the listing
+  login.logout_and_login_as(member[:username], member[:password])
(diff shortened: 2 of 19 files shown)


features/support/feature_tests/section/marketplace_paypal_preferences.rb
```diff
@@ -0,0 +1,27 @@
+module FeatureTests
+  module Section
```
ovan: I believe this is actually a page more than a section.

features/support/feature_tests/section/user_paypal_preferences.rb
```diff
@@ -0,0 +1,25 @@
+module FeatureTests
+  module Section
```
ovan: I believe this is actually a page more than a section.

features/support/feature_tests/action/login.rb
```diff
@@ -0,0 +1,33 @@
+module FeatureTests
+  module Action
+    module Login
+      extend Capybara::DSL
+      extend RSpec::Matchers
```
ovan: These two could be abstracted to a helper module that is then extended in every action/page/section. The helper module would define a Browser constant that is a shortcut to Capybara::DSL module. The helper module would also extend RSpec::Matchers to bring expect, raise_error, eql, etc. into scope.

features/paypal/paypal.feature
```diff
@@ -0,0 +1,5 @@
+Feature: Transaction with PayPal
+
+  @javascript
+  Scenario: Test transaction with PayPal
+    Then I expect transaction with PayPal test to pass
```
ovan: I'd love to get rid of this unnecessary boilerplate since we're not utilizing the Gherkin DSL anymore. That can be done in a later PR though as it requires a bit of thinking how to separate from non-Selenium based tests.

features/step_definitions/paypal_steps.rb
```diff
@@ -0,0 +1,48 @@
+# coding: utf-8
+
+def dismiss_onboarding_wizard_dialog
+  expect(page).to have_content("Woohoo, task completed!")
+  page.click_on("I'll do it later, thanks")
+end
```
rap1ds: Should this be in section module? Section::Wizard, Section::OnboardingPopup, or something like that?

features/support/feature_tests/action/paypal.rb
```diff
@@ -0,0 +1,112 @@
...
+
+        # Connect Paypal for admin
+        topbar.navigate_to_admin
+        admin_sidebar.click_payments_link
+        paypal_preferences.connect_paypal_account
+
+        expect(page).to have_content("PayPal account connected")
+
+        # Save payment preferences
+        paypal_preferences.set_payment_preferences("2.0", "5", "1.0")
```
rap1ds: keyword parameters would help readability a lot, something like

```
paypal_preferences.set_payment_preferences(commission: "2.0", min_commission: "5",  min_price: "1.0")
```

features/support/feature_tests/action/paypal.rb
```diff
@@ -0,0 +1,112 @@
...
+
+        expect(page).to have_content("Hooray, everything is set up!")
+      end
+
+      def request_listing(listing_title)
+        home = FeatureTests::Page::Home
+        listing = FeatureTests::Page::Listing
+        listing_book = FeatureTests::Page::ListingBook
+
+        home.click_listing(listing_title)
```
rap1ds: All the actions seem to start with some `topbar` action, which is cool, because the topbar is available in all pages and thus the actions are available, where ever you are. So that made me think, should there be a topbar action also here? E.g. `topbar.click_logo`.

Or if there's a hard-dependency, that some action can be only performed when we already are in some page, then we could add `expect { ...you are in this and that page } `.

Does that make sense?

features/support/feature_tests/action/paypal.rb
```diff
@@ -0,0 +1,112 @@
...
+        listing.fill_in_booking_dates
+        listing.click_request
+
+        expect(page).to have_content("Request #{listing_title}")
+        listing_book.fill_in_message("Trölölö #{listing_title}")
```
rap1ds: This might be a bit nitpicking (sorry! :) ) because this is just insignificant test content, but since we want to attract huge amount of contributors from all over the world, I would stick with ASCII characters and English language :)

features/paypal/paypal.feature
kpuputti: Let's postpone this to another PR to see if it's worth doing.

features/support/feature_tests/action/paypal.rb
kpuputti: This was the only actions that didn't start with the topbar that is visible in all pages, so I just added the logo click as you suggested.

That adds an extra request that wouldn't be needed for this test case, but maybe we can take that hit to avoid weird problems in the future?

features/support/feature_tests/action/paypal.rb
kpuputti: As I already mentioned, I like to use weird chars in test strings just to test that those are handled properly.

I made this more approachable with http://unicodesnowmanforyou.com/ :D

features/support/feature_tests/action/login.rb
kpuputti: I had this, but the tests didn't find the new `FeatureTests::Helper` module when loading all the files.

Dumb reason not to do this, but we were not able to solve the loading problem with @rap1ds and decided to leave the test helpers as they were.

PR #2446, opened 2016-08-24
Menuitem blur stopped

Gnito:
ToDo
- [ ] more testing: with win10 and android

Fix for Windows10 Chrome links not working inside topbar dropdowns.
(There Chrome thinks it's on a touch device even though it ain't - I'm not sure if there's good fix for that) 

Since Win10 Chrome thinks it's on touch device it uses 'click' pattern on opening drop downs. However, it sends _blur_ event before normal ```<a>``` tag navigation event kicks in. Therefore, we need to figure out if that event comes from child element of current dropdown element (in which case, we don't close the menu). RelatedTarget tells FocusEvents original source element. 

(FocusEvents behave differently than normal events - i.e. they don't bubble from child node to parent and their target is ```document``` instead of event source. Since focusable elements listen global document events, they can steal focus from each others despite their location on DOM hierarchy.) 

Cleaning related:
ref pattern: https://github.com/yannickcr/eslint-plugin-react/issues/678#issue-165177220

Check also commit comments.

### DIFF
--- a/client/app/components/composites/AvatarDropdown/AvatarDropdown.js
+++ b/client/app/components/composites/AvatarDropdown/AvatarDropdown.js
@@ -32,8 +32,12 @@ class AvatarDropdown extends Component {
     }
   }
 
-  handleBlur() {
-    this.setState({ isOpen: false });// eslint-disable-line react/no-set-state
+  handleBlur(event) {
+    // FocusEvent is fired faster than the link elements native click handler
+    // gets its own event. Therefore, we need to check the origin of this FocusEvent.
+    if (!this.profileDropdown.contains(event.relatedTarget)) {
+      this.setState({ isOpen: false });// eslint-disable-line react/no-set-state
+    }
   }
 
   render() {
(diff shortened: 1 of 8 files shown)


client/app/components/composites/AvatarDropdown/AvatarDropdown.js
```diff
@@ -32,8 +32,12 @@ class AvatarDropdown extends Component {
     }
   }
 
-  handleBlur() {
-    this.setState({ isOpen: false });// eslint-disable-line react/no-set-state
+  handleBlur(event) {
+    // FocusEvent is fired faster than the link elements native click handler
```
mporkola: This is `handleBlur`, but you're talking about FocusEvent. Just checking if this is correct!

client/app/utils/featureDetection.js
```diff
@@ -16,8 +16,13 @@ const canUseDOM = !!(typeof window !== 'undefined' &&
...
 
+const hasTouchEvents = !!(typeof window !== 'undefined' &&
+                          (('ontouchstart' in window) ||
+                            window.navigator.msMaxTouchPoints > 0));
+
 export {
   canUseDOM,
   canUsePushState,
   hasCSSFilters,
+  hasTouchEvents,
```
mporkola: 👍

client/app/components/composites/AvatarDropdown/AvatarDropdown.js
Gnito: "The FocusEvent interface represents focus-related events like focus, blur, focusin, or focusout."

PR #2447, opened 2016-08-24
Hover works better with mobile search if it's above it

Gnito:


client/app/components/composites/AvatarDropdown/ProfileDropdown.css
```diff
@@ -7,7 +7,7 @@
   background-color: white;
   box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.1);
   border-radius: 3px;
-  z-index: var(--ProfileDropdown_zIndex);
+  z-index: calc(var(--ProfileDropdown_zIndex) + 1);
```
Gnito: Instead of changing this, I'll increase variable :)

PR #2448, opened 2016-08-25
New listing button label

lyyder:
Add an editor for modifying the _Post a new listing_ button label.


app/views/admin/communities/edit_look_and_feel.haml
```diff
@@ -58,8 +58,8 @@
       .row
         .col-12
           - post_new_listing = "<i>#{t("homepage.index.post_new_listing")}</i>"
-          = form.label :custom_color2, t(".community_custom_color2", post_new_listing: post_new_listing).html_safe, :class => "input"
-          = render :partial => "layouts/info_text", :locals => { :text =>  t(".custom_color2_instructions_text", post_new_listing: post_new_listing).html_safe }
+          = form.label :custom_color2, t(".new_listing_button_custom_color").html_safe, :class => "input"
```
mporkola: At least in new top bar, we're using custom color #1 as the color for "Add listing" button: https://github.com/sharetribe/sharetribe/blob/b5a16211329d29fd921bd9fe3141bab3eaa6995e/client/app/components/sections/Topbar/Topbar.js#L324

sktoiva: Has this changed at some point? Are we getting rid of custom color #2? The instruction and the title text in admin panel say quite clearly that the custom color 2 controls only the Post a new listing button color.

mporkola: Yeah, I read forward and looks like it. Apparently it's different in new design.

PR #2450, opened 2016-08-25
Improve Fakepal so that it can be used with the application, not only…

rap1ds:
… in tests


app/services/paypal_service/paypal_service_injector.rb
```diff
@@ -32,30 +32,36 @@ def build_billing_agreements
       PaypalService::API::BillingAgreements.new(PaypalService::MerchantInjector.build_paypal_merchant)
     end
 
+    def events
+      Events.new({
```
ovan: The identations looks pretty bad like this. :(

PR #2452, opened 2016-08-25
For unpaid listing, the total price is unit price * quantity

rap1ds:
Steps to reproduce:

Prerequisite: PayPal setup ready for the marketplace

1. Add an Order Type with units enabled
2. Add a new listing "Bike" with price $10 per item
3. Log in as a seller
4. Go to the listing page of the "Bike", select that you want to buy 9 bikes and click the Buy button.
5. You are now being redirected to PayPal. Leave the PayPal page open but don't finish the payment. This leaves the transaction to `initiated` state.
6. Log in as an admin
7. Go to Admin > View Transactions and open the latest transaction

Expected:

You should see the details of the transaction: Price per item $10, quantity 9, subtotal 90, total 90.

screen shot 2016-08-25 at 14 32 14

Actual:

Total is 10.

screen shot 2016-08-25 at 14 31 37

### DIFF
--- a/app/services/transaction_service/gateway/paypal_adapter.rb
+++ b/app/services/transaction_service/gateway/paypal_adapter.rb
@@ -8,12 +8,6 @@ module TransactionService::Gateway
     end
 
     def create_payment(tx:, gateway_fields:, prefer_async:)
-      # Note: Quantity may be confusing in Paypal Checkout page, thus,
-      # we don't use separated unit price and quantity, only the total
-      # price for now.
-      shipping_total = Maybe(tx[:shipping_price]).or_else(0)
-      order_total = tx[:unit_price] * tx[:listing_quantity] + shipping_total
-
       create_payment_info = DataTypes.create_create_payment_request(
         {
          transaction_id: tx[:id],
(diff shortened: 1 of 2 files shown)


app/services/transaction_service/gateway/paypal_adapter.rb
```diff
@@ -65,7 +65,7 @@ def get_payment_details(tx:)
 
       payment_total = payment[:payment_total].or_else(nil)
       total_price = Maybe(payment[:payment_total].or_else(payment[:authorization_total].or_else(nil)))
-                    .or_else(tx[:unit_price])
+                    .or_else(tx[:unit_price] * tx[:listing_quantity])
```
ovan: I have a feeling this is the wrong place to default to tx based calculation. We are in payment service so the data returned is intentionally purely payment based. On top of this there's transaction_service/transaction.rb that return the final tx response. On line 183 there method to_tx_response already adds item_total into the response that is calculated using this exact logic. So ultimately defaulting to showing the item_total in place of actual payment total is the responsibility of the view (where it's possible also to tell the user that this is not the final payment price and only an estimate because we don't know all the details before an actual payment is made).

PR #2454, opened 2016-08-25
Adding several icons to the custom landing page

Luis-RG:
Adding several icons to the custom landing page to uniform all the SVG codes in the future.


mporkola:
Can't say much about the SVG content, but looks safe to merge 👍

PR #2460, opened 2016-08-26
When priority link limit was set to 0, menu didn't show up

Gnito:
Previous blinking menu fix, broke show-no-priority-links option. 
(i.e. when admin sets maximum number of links displayed in the top bar to 0)


kpuputti:
Oops, good catch.

PR #2462, opened 2016-08-26
Remove community ID from topbar menu url

lyyder:
- Remove id form top bar menu url
 - Redirect old menu links paths to new top bar paths


sktoiva:
Looks good, remember to test and wait circle.

PR #2463, opened 2016-08-26
Refactor PreauthorizeTransactionController

rap1ds:
This PR aims to refactor PreauthorizeTransactionController so that it's easier to reason about. In addition to that, some duplication was removed by combining `book` method with `initiate` method and `booked` method with `initiated` method.

Goals for refactoring:

* Get rid of early exists (e.g. `redirect_to :error if something_failed`), unless they are in the very beginning of the method (i.e. they are "guards"). For this, we use Result object for nicer control flow.
* Improve parameter handling, validation and coercion. For this, we use EntityBuilder.
* Extract validation logic out of controller action method
* Extract business logic (shipping price calculations e.g.) out of controller action method

### DIFF
--- a/app/controllers/preauthorize_transactions_controller.rb
+++ b/app/controllers/preauthorize_transactions_controller.rb
@@ -1,10 +1,10 @@
+# coding: utf-8
 class PreauthorizeTransactionsController < ApplicationController
 
   before_filter do |controller|
    controller.ensure_logged_in t("layouts.notifications.you_must_log_in_to_do_a_transaction")
   end
 
-  before_filter :fetch_listing_from_params
   before_filter :ensure_listing_is_open
   before_filter :ensure_listing_author_is_not_current_user
   before_filter :ensure_authorized_to_reply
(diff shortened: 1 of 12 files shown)


app/controllers/preauthorize_transactions_controller.rb
```diff
@@ -1,263 +1,382 @@
...
-    :listing_id
-   ).with_validations {
-    validates_presence_of :listing_id
-    validates :delivery_method, inclusion: { in: %w(shipping pickup), message: "%{value} is not shipping or pickup." }, allow_nil: true
+  PARSE_DATE = ->(v) {
+    return if v.nil?
+    begin
+      TransactionViewUtils.parse_booking_date(v)
+    rescue ArgumentError => e
+      e
```
ovan: Why do we return the error from rescue? What happens to it?

app/controllers/preauthorize_transactions_controller.rb
```diff
@@ -1,263 +1,382 @@
...
-      country_code: community_country_code,
-      listing: vprms[:listing],
-      delivery_method: delivery_method,
-      subtotal: vprms[:subtotal],
-      author: query_person_entity(vprms[:listing][:author_id]),
-      action_button_label: vprms[:action_button_label],
-      expiration_period: MarketplaceService::Transaction::Entity.authorization_expiration_period(vprms[:payment_type]),
-      form_action: booked_path(person_id: @current_user.id, listing_id: vprms[:listing][:id]),
-      price_break_down_locals: price_break_down_locals
+      listing_entity = ListingQuery.listing(params[:listing_id])
```
ovan: I think ListingQuery is just extra noise at this point. The controller is already using the `listing` method everywhere so this doesn't buy anything but is an unnecessary extra thing.

ovan: Too fast. It didn't realize that the listing hash is also passed to views.

ovan: The list of locals passed down to views is quite massive. I think this is enough at this point but let's look for opportunities to clean the views and reduce the locals list when we touch this stuff later.

app/controllers/preauthorize_transactions_controller.rb
```diff
@@ -1,263 +1,382 @@
+# coding: utf-8
```
ovan: ?

ovan:
Looks much more readable now. Good improvement, nothing to add at this point. Needs some selective squashing of commits to make the history look reasonable.

app/controllers/preauthorize_transactions_controller.rb
rap1ds: In error case, we need to return something that is not `Date` or `nil` so that the `date` validator can recognize that it's not valid date. We could also return something else like symbol `:error`, no specific reason for the decision to return the exception instead of the symbol.

I'll add a comment explaining this.

app/controllers/preauthorize_transactions_controller.rb
rap1ds: That's a hint to the interpreter to use UTF-8 instead of default US-ASCII. Emacs adds those shebangs automatically.

Issue #2449, opened 2016-08-25
Zero icon displayed since upgrading from 5.7.1 to v5.8


dwidoo:
##### Sharetribe Version:

_v5.8.0 ( 5252ff88a5e0a5e0f8890ce52a8bbca142f34a79 )

##### Environment:

_Development mode : Mac OS X 
AND 
Production mode : Ubuntu Server 14.04

##### Server mode:

_Development And Production mode!

##### Description:
I experience this in dev mode and prod mode.
Zero icon of font-awesome is displayed since I have upgraded from 5.7.1 to 5.8.0
The weird thing is that I can see with the browser inspector that the asset `font-awesome.min.css` is correctly served.
How could I check that the webfonts are correctly served please?

In Firefox:
![firefox](https://cloud.githubusercontent.com/assets/4923210/17962346/aae25f4c-6ab0-11e6-934f-50898b9603ea.jpg)

In Chrome:
![chrome](https://cloud.githubusercontent.com/assets/4923210/17962355/b89249ea-6ab0-11e6-91c3-56059ff52acb.jpg)

pcm211:
this is the reasons as the assets have been recompiled by you ... and the assets have changed after your last compile .... so for recompiling the assets do the following as see the icons reappearing 

$ rm -rf /public/assets    # remove the old precompiled assets from public dir 

$ RAILS_ENV=production rake assets:precompile 


Hope the above solves the problem

dwidoo:
hey @pcm211 
I have this issue in development and production mode.
Just tried the following but unfortunately, it didn't solve the problem 
`rm -rf /public/assets`
`RAILS_ENV=development rake assets:precompile`

pcm211:
oops its 

rm -rf public/assets

dwidoo:
@pcm211 thank you Vivek but this does not solve the issue.
I add that there is no error in the browser console!

pcm211:
@dwidoo ... it should have ... :(

dwidoo:
I can show you their url if there is a way to do this in private

On sam., août 27, 2016 at 7:23 AM, Vivek Bardia <notifications@github.com> wrote:
@dwidoo [https://github.com/dwidoo] ... it should have ... :(

dwidoo:
thank you Vivek, your instructions have helped me
This is SOLVED
Problem was that some files were customized to solve the icons issue and I had not replaced them with the 5.8 version.

PR #2465, opened 2016-08-26
Update README.md

pcm211:
this cmd would remove the precompiled assets from public/assets dir ... and leave a clean place for the subsequent command 

not using this command leaves stylesheets incorrectly linked and leads to problems in Modernizer, Squares instead of Icons etc etc 

PS : we would like to add a procfile cmd for the production envt too


mporkola:
Hi @pcm211, thanks again for contributing. I'm missing a bit of context here: is this related to production deployment, and https://github.com/sharetribe/sharetribe/issues/2449?

ovan:
Closing as inactive.

PR #2467, opened 2016-08-28
Use custom Post a new listing button text in header and email

thomasmalbaux:
The _Post a new listing_ button text can now be edited from the admin panel. For consistency, this custom text should also be used in two places:

- Listing creation form (header)
- CTA button in update emails (button label)

This PR also removes the old translation keys, as they are no longer relevant.

**One question**
The new text is set as `homepage.index.post_new_listing`. Because it is actually also used in other places than homepage (other pages and now listing creation page and email), should the translation key be changed?

### DIFF
--- a/app/views/community_mailer/community_updates.html.haml
+++ b/app/views/community_mailer/community_updates.html.haml
@@ -31,7 +31,7 @@
 
                             %tr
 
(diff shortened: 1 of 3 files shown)


sktoiva:
Probably in ideal case the key should be changed. But that requires now excess amount of work, as we would have to write a migration and be sure that all already entered translations are pointing to the new key.

sktoiva:
Moreover, it will break the existing UI for a pretty visible element as new translations would be required. So, I would not do it now.

PR #2468, opened 2016-08-28
Renamed "Payments" admin panel section to "Payment system"

thomasmalbaux:
The current name is confusing for some users and some understand that they'll be charge a fee by Sharetribe via their PayPal account. 

The new name should help to explain that this is about the online payment system, and not the Sharetribe subscription.


thomasmalbaux:
So small (only copy texts) and tests are OK so I'll merge directly.

PR #2469, opened 2016-08-29
Enable new layout admin page for all

mporkola:


mporkola:
Added `Blocked` as reminder: we want to debug something before merging this.

PR #2474, opened 2016-08-30
Initial Search Page draft: transit, immutable, redux

Gnito:
TODO:
- [x] Test: transit and immutable
- [x] Convert Transit to Immutable before storing it to store
- [x] Pass props through redux store 
- [x] Read transit data
- [x] Insert some React component to the page 
- [x] Clean subset routes from SearchPageApp.js
- [x] Redux data flow: more scrutinity needed
- [x] Clean ToDos

Initial draft for data flow: bootstrapped transit data -> Immutable format -> redux store -> props -> rendered content.

More stuff to be done in separate branches:
- Fetch Data through Rails side service (to be bootstrapped)
- Figure out how to deal with application.haml stuff (notifications to props etc.)
- Draft separate ListingCard composite
- Draft paging functionality (with separate components)
- Change current page
- Change thunk to saga?
- Add direct calls to DiscoveryAPI: authentication
- Lazy load next page
- Could we include Topbar component inside of SearchPage (and remove it from application.haml)

When testing data bootstrapping I've used following method in homepage controller:

```ruby
  # TODO this is just for testing purposes
  def bootstrap_data
    my_conn = Faraday.new(url: APP_CONFIG.external_search_url) do |c|
      c.request  :url_encoded             # form-encode POST params
      c.response :encoding                # use Faraday::Encoding middleware
      c.adapter  Faraday.default_adapter  # make requests with Net::HTTP
    end

    begin
      res = my_conn.get do |req|
        req.url("/discovery/1/listings/query", {
          "filter[distance_max]" => 5.0,
          "filter[omit_closed]" => true,
          "page[number]" => 1,
          "page[size]" => 24,
          "search[distance_unit]" => 'miles',
          "search[keywords]" => '',
          "search[lat]" => 60.167458,
          "search[lng]" => 24.931075,
          "search[locale]" => 'en',
          "sort" => 'distance',
          "search[scale]" =>  5,
          "search[offset]" => 0.5,
          "include-body" => true
        })
        req.headers['Authorization'] = "apikey key=#{APP_CONFIG.external_search_apikey}"
      end
      Result::Success.new(res)
    rescue StandardError => e
      Result::Error.new(e)
    end
  end
```

and in index:

```ruby
    # if FeatureFlagHelper.feature_enabled?(:search_page_v1)
    if true
      bootstrapped_data = bootstrap_data[:data].body
      render template: "search_page/search_page", locals: { bootstrapped_data: bootstrapped_data } and return
    end
```


app/views/search_page/search_page.erb
```diff
@@ -0,0 +1,12 @@
+<% props = {
+  discoveryAPI: {
+    url: 'somedomain.com/discovery/query'
```
mporkola: I know it's WIP, but maybe add the config var already?

client/app/components/sections/SearchPage/SearchPage.js
```diff
@@ -0,0 +1,41 @@
+import { Component, PropTypes } from 'react';
+import { div, img, p } from 'r-dom';
+
+import css from './SearchPage.css';
+
+class SearchPage extends Component {
+
+  constructor(props, context) {
+    super(props, context);
+    this.listings = this.props.searchPage.get('listings') || [];
```
kpuputti: The `this.listings` property is only used in the render. I don't think that's a reason to make this a class instead of a clean stateless functional component that should be the default choise.

client/app/components/sections/SearchPage/SearchPage.js
```diff
@@ -0,0 +1,41 @@
...
+        ])
+      )),
+    ]);
+  }
+}
+
+const { object } = PropTypes;
+
+SearchPage.propTypes = {
+  searchPage: object, // eslint-disable-line react/forbid-prop-types
```
mporkola: We can actually use immutable records here, e.g. `PropTypes.instanceOf(SearchPage)` (in case that's defined as a record, didn't go there yet)

client/app/constants/SearchPageConstants.js
```diff
@@ -0,0 +1,6 @@
+// Search page
+const CURRENT_PAGE_UPDATE = 'SEARCH::CURRENT_PAGE_UPDATE';
+
+export {
+  CURRENT_PAGE_UPDATE,
+};
```
kpuputti: We have such a few constants still that splitting those into multiple files is IMO cumbersome. Could they all live in a single file where you can easily see all possible action types?

kpuputti:
Loving all the Immutable.js you bring in this PR! <3

app/views/search_page/search_page.erb
Gnito: I think it should come in to this erb through locals

client/app/components/sections/SearchPage/SearchPage.js
Gnito: This is going to be page level component. There will be methods like shouldComponentUpdate etc. - so, I decided to keep this as class from the beginning and thus make things easier to test.

client/app/components/sections/SearchPage/SearchPage.js
Gnito: Sounds good :+1:

client/app/constants/SearchPageConstants.js
Gnito: I'd want to keep onboarding away from this data cycle => searchpage and everything inside of it could be added into this file (including topbar search inputs). So, I agree, but don't want to do any changes to current files.

PR #2476, opened 2016-08-30
Add guidelines for new UI tests

kpuputti:


docs/ui-testing.md
```diff
@@ -0,0 +1,24 @@
+# UI testing
+
+This document describes the UI testing guidelines for the Sharetribe platform.
+
+## Cucumber
```
rap1ds: Capybara?

rap1ds:
Looks good! I think we could now remove the old "Cucumber do's and dont's" document? That could be done in this PR.

In addition, we could replace the "Cucumber do's and dont's" document link in https://github.com/sharetribe/sharetribe/blob/master/docs/README.md with link to this doc.

PR #2480, opened 2016-08-30
Add service name to confirmation email subject and content

thomasmalbaux:


thomasmalbaux:
This looks pretty good:

- confirmation email subject now includes `%{service_name}`
- confirmation email content now includes `%{service_name}`
- tests are OK

But this is not perfect:

- this doesn't work for the very first confirmation email: the confirmation email sent to the admin at the end of the trial creation form on marketing site. In this first email, `%{service_name}` is *Sharetribe* and not the actual service name. Also in the non-white-label footer.
- All other confirmation emails (for regular users) work fine with `%{service_name}` included.

app/models/email.rb
```diff
@@ -56,7 +56,7 @@ def self.email_available?(email, community_id)
   end
 
   def self.send_confirmation(email, community)
-    MailCarrier.deliver_later(PersonMailer.email_confirmation(email, community))
+    Delayed::Job.enqueue(EmailConfirmationJob.new(email.id, community.id), priority: 2)
```
ovan: If I remember correctly there was some point why the mail carrier abstraction was put in place. The deliver_later is not implemented yet but it could be. Maybe @rap1ds remembers the details?

rap1ds: I guess this commit message explains the thinking: https://github.com/sharetribe/sharetribe/commit/b0a36a006c10e19900c4adb91f2f896723a67dda .

PR #2481, opened 2016-08-30
Listing night picker

kpuputti:
This branch adds calendar date picker to "per night" unit.

In additional to the code changes, this branch has migrations that change the existing transactions with the old "per night" unit to custom units.

### DIFF
--- a/app/services/transaction_service/store/transaction.rb
+++ b/app/services/transaction_service/store/transaction.rb
@@ -167,7 +167,7 @@ module TransactionService::Store::Transaction
     if m.booking
       booking_data = EntityUtils.model_to_hash(m.booking)
       hash.merge(booking: Booking.call(
-                  booking_data.merge(duration: booking_duration(booking_data))))
+                  booking_data.merge(duration: booking_duration(booking_data, hash[:unit_type].to_sym))))
     else
       hash
     end
@@ -203,8 +203,9 @@ module TransactionService::Store::Transaction
 
       # TODO What's the correct place for the booking calculation logic?
       # Make sure listing_quantity equals duration
-      if booking_duration(tx_data[:booking_fields]) != tx_model.listing_quantity
-        raise ArgumentException.new("Listing quantity (#{tx_listing_quantity}) must be equal to booking duration in days (#{booking_duration(tx_data)})")
+      duration = booking_duration(tx_data[:booking_fields], tx_data[:unit_type])
+      if duration != tx_model.listing_quantity
+        raise ArgumentException.new("Listing quantity (#{tx_listing_quantity}) must be equal to booking duration in days/nigths (#{duration})")
       end
 
       start_on = tx_data[:booking_fields][:start_on]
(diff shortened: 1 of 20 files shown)


app/services/transaction_service/store/transaction.rb
```diff
@@ -217,10 +218,17 @@ def is_booking?(tx_data)
...
     end_on = booking_data[:end_on]
-    DateUtils.duration_days(start_on, end_on)
+
+    if unit_type == :day
+      DateUtils.duration_days(start_on, end_on)
+    elsif unit_type == :night
+      DateUtils.duration_nights(start_on, end_on)
+    else
+      raise ArgumentError.new("Unknown unit_type for a booking: #{unit_type}")
+    end
```
ovan: The calculation here shouldn't happen at all according to our model that we just sketched. And even if it happens unit_type shouldn't matter because start_on and end_on here are always stored the same way (end exclusive).

app/views/transactions/_price_break_down.haml
```diff
@@ -2,18 +2,27 @@
   - if booking
     .initiate-transaction-per-unit-wrapper
       %span.initiate-transaction-per-unit-label
-        = t("transactions.initiate.price_per_day")
+        - if unit_type == :day
+          = t("transactions.initiate.price_per_day")
+        - else
+          = t("transactions.initiate.price_per_night")
```
ovan: Just a note for the future that label would be controlled by unit_label in the upcoming design, not pricing_unit

db/migrate/20160831054910_old_night_unit_to_custom_unit.rb
```diff
@@ -0,0 +1,270 @@
...
+    end
+  end
+
+  def up
+    # 1. Transactions that need change:
+    transaction_res =
+      execute("
+        SELECT tx.community_id, tx.id FROM transactions tx
+        LEFT JOIN bookings ON bookings.transaction_id = tx.id
+        WHERE unit_type = 'night' AND bookings.id IS NULL
```
ovan: tx.unit_type ?

db/migrate/20160831054910_old_night_unit_to_custom_unit.rb
```diff
@@ -0,0 +1,270 @@
...
+      "bg"=>"Number of nights:",
+      "mn"=>"Number of nights:",
+      "zh-TW"=>"夜晚数量:",
+      "zh-HK"=>"晚：",
+      "ka"=>"ღამეების რაოდენობა:",
+      "sl"=>"Number of nights:",
+      "sk-SK"=>"Počet nocí:"
+    }
+  end
+end
```
ovan: wow, this migration script underlines the reasons why a) custom units and b) community_translations are not easy to handle like they are today.

ovan:
One important thing to test before deploying to production is if any of the changes affect search api indexer assumptions about valid values. I quickly looked at the import description and couldn't spot anything (we handle unit_type as string and not enumerate the possible values in our input data validation there) but still it should be tested with staging data.

PR #2483, opened 2016-08-30
Better positioning for multi line profile link labels

mporkola:
Before:

After:

### DIFF
--- a/client/app/components/composites/AvatarDropdown/ProfileDropdown.css
+++ b/client/app/components/composites/AvatarDropdown/ProfileDropdown.css
@@ -25,6 +25,8 @@
   justify-content: center;
   cursor: pointer;
   color: var(--ProfileDropdown_textColor);
+  padding: 0 0.69em;
+  text-align: center;
 
   &:hover {
     background-color: rgb(249, 249, 249);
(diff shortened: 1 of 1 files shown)


client/app/components/composites/AvatarDropdown/ProfileDropdown.css
```diff
@@ -25,6 +25,8 @@
   justify-content: center;
   cursor: pointer;
   color: var(--ProfileDropdown_textColor);
+  padding: 0.69em 1.15em;
+  text-align: center;
```
Gnito: I ended up playing with this too.
Here's the same with better centering:
```
.profileActionIconWrapper {
  position: relative;
  flex-basis: 60%;
  display: flex;
  align-items: flex-end;
}

.notificationCount {
  position: absolute;
  bottom: 17px; /* The only row changed in .notificationCount */
  left: 0;
  width: 100%;
  text-align: center;
  color: var(--ProfileDropdown_colorLight);
  font-size: var(--ProfileDropdown_fontSizeNotification);
}

.profileActionLabel {
  flex-basis: 40%;
  text-align: center;
  line-height: 1.25em;
  padding: 0 0.5em;
  max-width: 100%;
  word-wrap: break-word;
}
```

PR #2486, opened 2016-08-31
Recover from MissingTemplate error with a 404

lyyder:
Previously we have encountered incidents where a template partial for a given response format has been missing and thus has resulted in a `ActionView::MissingTemplate` error and an internal server error response.

Instead of failing with an internal server error, a 404 status with a rendered 404 page is returned and the event is logged with `INFO` level.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -418,6 +418,20 @@ class ApplicationController < ActionController::Base
     @minutes_to_maintenance = NextMaintenance.minutes_to(now)
   end
 
+  # Respond to a MissingTemplate error with 404.
+  # This prevents the server from responding with a 500
+  # in cases where a template for a requested format is
+  # not found.
+  rescue_from(ActionView::MissingTemplate) do |e|
+    logger.info(
+      "An ActionView::MissingTemplate error occurred, responding with 404",
+      :missing_template,
+      requested_url: request.original_url
+    )
+    render "/errors/status_404", status: 404, content_type: "text/html", locals: { status: 404 }
+  end
+
+
   private
 
   # Override basic instrumentation and provide additional info for


app/controllers/application_controller.rb
```diff
@@ -418,6 +418,20 @@ def maintenance_warning
...
+  # This prevents the server from responding with a 500
+  # in cases where a template for a requested format is
+  # not found.
+  rescue_from(ActionView::MissingTemplate) do |e|
+    logger.info(
+      "An ActionView::MissingTemplate error occurred, responding with 404",
+      :missing_template,
+      requested_url: request.original_url
+    )
+    render "/errors/status_404", status: 404, content_type: "text/html", locals: { status: 404 }
```
mporkola: Is it easy to respond with the requested content type? E.g. `application/json` if `.json` was requested

rap1ds: So this will now catch all the `MissingTemplate` exceptions, even if we make a error in our code (e.g. typo the template name). I'm just thinking, is this kind of "catch all" a good idea?

lyyder: @mporkola Yes, another option would be just to respond with an empty 404 and not touch the content type. Or even return the html representation with some other content type, as the application works in normal cases if for example `.json` is requested.

@rap1ds This is a clear downside. The logging is added to track these but of course it does not notify us about the occurrences.

lyyder:
After looking at our error history and `MissingTemplate` occurrences I've decided to close this PR and not merge the changes. We have not encountered an error due to a missing partial for a specific format in quite a while and valid `MissingTemplate` errors have occurred every now and then.

Also Rails 5 might provide a better way of handling unsupported formats \[[1](http://blog.bigbinary.com/2016/07/03/controller-actions-default-no-content-in-rails-5-if-template-is-missing.html)\] \[[2](https://github.com/rails/rails/pull/23827)\], so if this keeps on generating invalid error notifications we can re-prioritize once we upgrade to Rails 5.

PR #2490, opened 2016-08-31
Fix feature flag id handling

sktoiva:
Changes FeatureFlagsHelper to use explicitly defined person_id and community_id instead of relying to session.

This pull request will fix the odd case where in some cases admins are not presented with correct feature flags.

### Steps to reproduce

  1. In master, create a new marketplace through int_api
  2. Enable feature flag: `topbar_v1` through console to the newly created user
  3. Confirm email and log in with the auth token from int_api response

*Expected*: New top bar is visible
*Actual*: Old top bar is still visible
*Reason*: Auth token login doesn't put session[:person_id] into place

### Next steps

Removal of session[:person_id] is investigated later as it seems that it's not read anywhere except the removed code in FeatureFlagsHelper, but this needs to be tested thoroughly.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -167,10 +167,13 @@ class ApplicationController < ActionController::Base
   def initialize_feature_flags
     # Skip this if there is no current marketplace.
     # This allows to avoid skipping this filter in many places.
-    return unless request.env[:current_marketplace]
+    return unless @current_community
 
-    FeatureFlagHelper.init(request, Maybe(@current_user).is_admin?.or_else(false),
-                           Maybe(@current_user).is_marketplace_admin?.or_else(false))
+    FeatureFlagHelper.init(community_id: @current_community.id,
+                           user_id: @current_user&.id,
+                           request: request,
+                           is_admin: Maybe(@current_user).is_admin?.or_else(false),
+                           is_marketplace_admin: Maybe(@current_user).is_marketplace_admin?.or_else(false))
   end
 
   # Ensure that user accepts terms of community and has a valid email
--- a/app/controllers/landing_page_controller.rb
+++ b/app/controllers/landing_page_controller.rb
@@ -221,7 +221,12 @@ class LandingPageController < ActionController::Metal
                          true)
     marketplace_context = marketplace_context(c, topbar_locale, request)
 
-    FeatureFlagHelper.init(request, false, false)
+    # Init feature flags with marketplace specific flags, skip personal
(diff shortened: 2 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -169,8 +169,11 @@ def initialize_feature_flags
     # This allows to avoid skipping this filter in many places.
     return unless request.env[:current_marketplace]
 
-    FeatureFlagHelper.init(request, Maybe(@current_user).is_admin?.or_else(false),
-                           Maybe(@current_user).is_marketplace_admin?.or_else(false))
+    FeatureFlagHelper.init(community_id: @current_community&.id,
```
rap1ds: We are anyways returning unless there is `request.env[:current_marketplace]`, so we shouldn't need to use safe-navigator here.

app/helpers/feature_flag_helper.rb
```diff
@@ -43,7 +42,9 @@ def fetch_feature_flags(request, is_admin, is_marketplace_admin)
   def fetch_flags_from_service(community_id, person_id, is_admin, is_marketplace_admin)
     # for admin users fetch combined feature flags,
     # for non-admin users only fetch the community specific feature flags
-    if is_admin || is_marketplace_admin
+    if !community_id
+      Set[]
```
rap1ds: This check is unnecessary

app/views/analytics/_google_tag_manager.haml
```diff
@@ -6,11 +6,12 @@
       dl.push({"event": "identify", "id": "mp-"+marketplaceId+"-admin" });
     }
 
-    dl.push({
-      event: 'feature flags',
-      featureFlags: #{feature_flags.to_json.html_safe}
-    });
-
+    if (#{feature_flags.present?}){
```
rap1ds: `feature_flags` seems to be undefined, so I guess this will break? Maybe worth testing again.

app/controllers/landing_page_controller.rb
```diff
@@ -221,8 +221,6 @@ def render_landing_page(default_locale:, locale_param:, structure:)
                          true)
     marketplace_context = marketplace_context(c, topbar_locale, request)
 
-    FeatureFlagHelper.init(request, false, false)
```
rap1ds: I guess the original purpose for `false, false` was to render only the community specific feature flags. I'm not sure if that's right thing to do, but good to acknowledge that this change changes what we had before.

rap1ds:
Some comments there. The `session[:person_id` was a good catch 👍

PR #2491, opened 2016-08-31
Enable topbar to all new trials

sktoiva:


app/controllers/int_api/marketplaces_controller.rb
```diff
@@ -49,6 +49,10 @@ def create
     auth_token = UserService::API::AuthTokens.create_login_token(user[:id])
     url = URLUtils.append_query_param(marketplace[:url], "auth", auth_token[:token])
 
+    # Enable specific features for all new trials
+    FeatureFlagService::API::Api.features.enable(community_id: marketplace[:id], person_id: user[:id], features: [:topbar_v1])
```
Gnito: This can be done, but "Enabled for all" should be enough.

sktoiva: Of course it should be enough, but it would look weird in the admin ui.

rap1ds: LGTM, even though the checkbox issues feels like an issue in the UI.

PR #2492, opened 2016-09-01
Finalize sitemap

rap1ds:
Cherry-picked commits from #2303

### DIFF
--- a/app/controllers/sitemap_controller.rb
+++ b/app/controllers/sitemap_controller.rb
@@ -84,13 +84,7 @@ class SitemapController < ActionController::Metal
   end
 
   def can_show_sitemap?(community)
-    if APP_CONFIG.enable_sitemap&.to_s != "true"
-      render_not_found
-      false
-    elsif community.nil?
-      render_not_found
-      false
-    elsif community.deleted?
+    if sitemap_enabled? || community.nil? || community.deleted?
       head :not_found
       false
     elsif community.private?
@@ -101,7 +95,7 @@ class SitemapController < ActionController::Metal
     end
   end
 
-  def render_not_found(msg = "Not found")
-    redirect_to "/404"
+  def sitemap_enabled?
+    APP_CONFIG.enable_sitemap&.to_s == "true"
   end
 end
--- a/app/controllers/sitemap_controller.rb
+++ b/app/controllers/sitemap_controller.rb
@@ -9,37 +9,39 @@ class SitemapController < ActionController::Metal
   include ActionController::Redirecting
 
   def sitemap
-    return unless can_show_sitemap?
+    community = request.env[:current_marketplace]
 
-    asset_host = APP_CONFIG.asset_host
+    return unless can_show_sitemap?(community)
 
-    if asset_host.present?
-      redirect_to ActionController::Base.helpers.asset_url("/sitemap/generate.xml")
+    if APP_CONFIG.asset_host.present?
+      redirect_to ActionController::Base.helpers.asset_url(
+                    "/sitemap/generate.xml?sitemap_host=#{request.host}")
     else
-      render_site_map
+      render_site_map(community)
     end
   end
 
   def generate
-    return unless can_show_sitemap?
+    community = CurrentMarketplaceResolver.resolve_from_host(
+      params[:sitemap_host], URLUtils.strip_port_from_host(APP_CONFIG.domain))
 
-    render_site_map
+    return unless can_show_sitemap?(community)
+
+    render_site_map(community)
   end
 
   private
 
-  def community
-    request.env[:current_marketplace]
-  end
+  def render_site_map(community)
+    default_host = community.full_domain(with_protocol: true)
 
-  def render_site_map
-    sitemap = from_cache(community.id, max_sitemap_links) do
+    sitemap = from_cache([community.id, max_sitemap_links, default_host]) do
       adapter = SitemapGenerator::NeverWriteAdapter.new
 
       open_listings = find_open_listings(community.id)
 
       SitemapGenerator::Sitemap.create(
-            default_host: request.base_url,
+            default_host: default_host,
             verbose: false,
             adapter: adapter) do
         open_listings.each do |l|
@@ -65,8 +67,8 @@ class SitemapController < ActionController::Metal
       }
   end
 
-  def from_cache(community_id, limit, &block)
-    key = "sitemaps/#{community_id}/#{limit}"
+  def from_cache(keys, &block)
+    key = "sitemaps/#{keys.join("-")}"
     Rails.cache.fetch(key, expires_in: 24.hours, &block)
   end
 
@@ -81,7 +83,7 @@ class SitemapController < ActionController::Metal
     end
   end
 
-  def can_show_sitemap?
+  def can_show_sitemap?(community)
     if APP_CONFIG.enable_sitemap&.to_s != "true"
       render_not_found
       false
--- a/app/controllers/sitemap_controller.rb
+++ b/app/controllers/sitemap_controller.rb
@@ -11,15 +11,14 @@ class SitemapController < ActionController::Metal
     render_site_map
   end
 
-  def self.find_open_listings(community_id)
+  def find_open_listings(community_id)
     Listing
       .currently_open
       .where(community_id: community_id)
-      .limit(SitemapGenerator::MAX_SITEMAP_LINKS)
+      .limit(max_sitemap_links)
       .pluck(:id, :title, :updated_at)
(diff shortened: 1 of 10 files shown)


config/config.defaults.yml
```diff
@@ -379,6 +379,14 @@ default: &default_settings
   #
   font_proximanovasoft_url:
 
+  # Maximum number of links in sitemap.xml.
+  #
+  # The default number is 5000. Increasing the limit may affect on performance.
```
ovan: It's either "affect performance" or "have effect on performance"

config/initializers/mime_types.rb
```diff
@@ -3,3 +3,4 @@
 # Add new mime types for use in respond_to blocks:
 # Mime::Type.register "text/richtext", :rtf
 # Mime::Type.register_alias "text/html", :iphone
+Mime::Type.register "application/x-gzip", :xml_gz, [], ["xml.gz"]
```
rap1ds: Is this needed anymore? The Rack::Deflater handles now the gzipping, and in the response we send:

```
Content-Encoding: gzip
Content-Type: application/xml
```

Full cURL:

```
➜  ~ curl 'http://test.lvh.me:3000/sitemap.xml' -H 'Pragma: no-cache' -H 'Accept-Encoding: gzip, deflate, sdch' -H 'Accept-Language: en-US,en;q=0.8,fi;q=0.6,de;q=0.4,fr;q=0.2' -H 'Upgrade-Insecure-Requests: 1' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36' -H 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8' -H 'Cache-Control: no-cache' -H 'Cookie: _sharetribe_session_newnewnew=2194b4c4245929eb8ac04fb7ae4ebf50; _sharetribe_session_newnewnew4=814a4b5ea25b5777676c9b8c5724415f; _st_session=c9d6795069a8a65b41f936ff5c1db67e' -H 'Connection: keep-alive' -i
HTTP/1.1 200 OK
X-Frame-Options: SAMEORIGIN
X-Xss-Protection: 1; mode=block
X-Content-Type-Options: nosniff
Content-Disposition: attachment; filename="sitemap.xml"
Content-Transfer-Encoding: binary
Content-Type: application/xml
Cache-Control: private
Etag: W/"e0002c666520520bb6e5ded9b894e758"
X-Meta-Request-Version: 0.4.0
X-Request-Id: 35c92e0a-d63f-4b7f-82f6-541dcb6116ce
X-Runtime: 0.048266
Vary: Accept-Encoding
Content-Encoding: gzip
Server: WEBrick/1.3.1 (Ruby/2.3.1/2016-04-26)
Date: Thu, 01 Sep 2016 13:57:19 GMT
Content-Length: 467
Connection: Keep-Alive
Set-Cookie: _st_session=c9d6795069a8a65b41f936ff5c1db67e; path=/; expires=Sat, 01 Oct 2016 13:57:19 -0000; HttpOnly
```

So to my understanding we can remove this, right?

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,107 @@
...
+  include ActionController::Redirecting
+
+  def sitemap
+    community = request.env[:current_marketplace]
+
+    return unless can_show_sitemap?(community)
+
+    if APP_CONFIG.asset_host.present?
+      redirect_to ActionController::Base.helpers.asset_url(
+                    "/sitemap/generate.xml?sitemap_host=#{request.host}")
```
ovan: Why do we need to pass the host instead of pulling it in generate?

config/initializers/mime_types.rb
ovan: The difference is that mime type applies to content being transferred and transfer encoding only applies to the transfer, i.e. the sitemap.xml is delivered over the wire as gzipped but the client delivers it to requesting app as xml. With application/x-gzip the content itself is gzipped so if you serve a sitemap.xml.gz file then that's appropriate mime type. I assume we only serve the xml and not a .gz file containing the xml? In that case application/xml is the correct mime type.

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,107 @@
...
+    elsif community.private?
+      head :forbidden
+      false
+    else
+      true
+    end
+  end
+
+  def render_not_found(msg = "Not found")
+    redirect_to "/404"
```
ovan: 404 should be returned directly instead of redirect. LandingPageController does it like this:

```ruby
    self.status = 404
    self.response_body = msg
  end
```

app/controllers/sitemap_controller.rb
ovan: Ok got it, we use different host for assets if asset_host is set.

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,107 @@
...
+            adapter: adapter) do
+        open_listings.each do |l|
+          add listing_path(id: l[:id]), lastmod: l[:lastmod]
+        end
+      end
+
+      adapter.data
+    end
+
+    send_data(sitemap, filename: "sitemap.xml")
```
ovan: We cache the content for 24 hours to memcached which is great. In addition to this we should also set the cache-control headers correctly. Right now we send back cache-control: private which means that no client is allowed to cache the response. The minimum we should do is to set expiry to 24h. This is easy to do by including ActionController::ConditionalGet and using expire_in method. Better still would be to handle conditional gets on controller level (now there's rack middleware but since we're not sending a string body I don't think it kicks in. This is a bit more work so I'll leave it up to you to consider if it's worth doing. LandingPageController can be checked for reference.

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,107 @@
...
+  def can_show_sitemap?(community)
+    if APP_CONFIG.enable_sitemap&.to_s != "true"
+      render_not_found
+      false
+    elsif community.nil?
+      render_not_found
+      false
+    elsif community.deleted?
+      head :not_found
+      false
```
ovan: Can you explain what's the idea behind handling community.deleted? differently from sitemap disabled and community.nil?

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,107 @@
+class SitemapController < ActionController::Metal
+
+  include AbstractController::Rendering
+  include ActionController::MimeResponds
+  include ActionController::DataStreaming
+  include ActionController::RackDelegation
+  include ActionController::Rescue
+  include ActionController::Head
+  include ActionController::Redirecting
+
```
ovan: With metal controller you want to add:

```ruby
  # Ensure ActiveSupport::Notifications events are fired
  include ActionController::Instrumentation
```

This triggers the structured request logging.

app/controllers/sitemap_controller.rb
rap1ds: The problem with that method is that it doesn't go to ErrorsController and it doesn't render the nice error page, just a blank page with the message.

I'll have a look how to fix this.

app/controllers/sitemap_controller.rb
ovan: Sorry, I checked my facts. Instrumentation alone is not enough. append_into_to_payload method also needs to be overridden. The version used in LandingPageController is:

```ruby
  # Override basic instrumentation and provide additional info for
  # lograge to consume. These are pulled and logged in environment
  # configs.
  def append_info_to_payload(payload)
    super
    payload[:host] = request.host
    payload[:community_id] = community(request)&.id
    payload[:current_user_id] = nil
    payload[:request_uuid] = request.uuid
  end
```

rap1ds: Good catch

app/controllers/sitemap_controller.rb
ovan: That's true but it's still the wrong way to do it. We should have a way to render the nice error page directly as a response when needed. In this case though I think this matters even less because the sitemap.xml is mainly for crawlers which presumably don't care about the prettiness much. :)

rap1ds: True. I guess that's the reason why `head :not_found` was used originally. No need to be fancy, correct status code is probably good enough for crawlers.

config/initializers/mime_types.rb
rap1ds: Makes sense, thanks for explanation!

PR #2495, opened 2016-09-01
Listing needs to be added by admin to count as onboarding progress

mporkola:


app/controllers/listings_controller.rb
```diff
@@ -312,12 +312,14 @@ def create
         ).html_safe
 
         # Onboarding wizard step recording
-        state_changed = Admin::OnboardingWizard.new(@current_community.id)
-          .update_from_event(:listing_created, @listing)
-        if state_changed
-          report_to_gtm({event: "km_record", km_event: "Onboarding listing created"})
+        if @current_user.has_admin_rights?
```
rap1ds: `has_admin_rights?` returns `true` for both marketplace admins AND super admins. Is that on purpose or should be include only marketplace admins and ignore super admins?

ovan: The intention of the design is that the OnboardingWizard is fed with different types of events and it's then responsible for deciding, by looking at the event data, whether or not the event means stepping forward in the wizard flow. So following that architecture there should be no conditionals here but instead the listing author would be passed as part of the event data. This is important e.g. when the wizard flow is bootstrapped from existing marketplace data. If it handles all the logic it can then do the bootstrap following the same logic as the gradual recording does. The bootstrapping functionality was originally put in place because we retroactively applied the wizard to previously created marketplace. It's still useful to have even now when wizard is enabled for all because this way the state can be cleared (because of a bug fix, because of wanting to change the logic, etc.) leading the wizard to rebuild it's state correctly.

app/services/admin/onboarding_wizard.rb
```diff
@@ -103,7 +103,7 @@ def paypal_preferences_updated(setup_status, community)
 
     def listing_created(setup_status, listing)
       if !setup_status[:listing] &&
-         listing
+         listing&.author&.is_marketplace_admin?
```
ovan: 👍 

This now affects the bootstrap logic a bit. The way it works now is that it pulls the first listing for the community and checks that. But since we no longer accept any listing to mean :listing step is complete we should now pull all listings in the bootstrap phase (in batches) and process them through this callback as long as we either get a hit or we run out of listings.

app/services/admin/onboarding_wizard.rb
```diff
@@ -139,7 +139,9 @@ def init_setup_steps(community_id)
       locales = community.locales
       community_customizations = CommunityCustomization.where(community_id: community.id).select { |cc| locales.include?(cc.locale) }
       custom_field = CustomField.find_by(community_id: community.id)
-      listing = Listing.find_by(community_id: community.id)
+      listing = Listing.where(community_id: community.id)
+                  .joins(author: :community_membership)
+                  .where('community_memberships.admin' => true).first
```
ovan: Looks good. It's a different solution from what I had in mind but feels simpler this way. The check logic is no duplicated from here to event handler but maybe that's reasonable as it speeds up and simplifies the bootstrap data processing side vs. processing all listings.

PR #2496, opened 2016-09-01
iOS Chrome creates resize event when keyboard shows up

Gnito:
iOS Chrome creates resize event when keyboard shows up and it closes the mobile search form.
However, we need to change the state before mobile mode changes to desktop mode


kpuputti:
It took me a while to understand the solution. It looks like it only closes the mobile search in desktop mode, which doesn't seem to make sense since the mobile  search isn't used in desktop mode.

But apparently the change from mobile to desktop mode has to set the mobile menu closed, which is why the fix is done as it is. Maybe this could be clarified in a comment in the code itself?

Other than that and the small ESLint error, looks good!

PR #2497, opened 2016-09-01
Add translation fallback mechanism

rap1ds:
This PR adds a fallback mechanism for untranslated translation keys.

The default fallback mechanism in Rails provides a way to first look key A in language 1, then look key A in language 2.

This fallback mechanism lets us look key A in language 1 and if that is not found, then key B in language 1. Only after that, the default fallback mechanism tries to look for key B in language 2.

This way we can start using new translations and "deprecate" the old translations. Without this mechanism, users would see wrong language (most likely English) in the UI when we change translations before translators have had change to translate them.

### DIFF
--- a/app/helpers/translation_helper.rb
+++ b/app/helpers/translation_helper.rb
@@ -1,0 +1,55 @@
+module TranslationHelper
+  class NewTranslationOrFallback
+
+    def initialize(new_translation = nil)
+      @new_translation = new_translation
+    end
+
+    def or_fallback_to(tr_key, opts = {})
+      @new_translation || I18n.translate(tr_key, opts)
+    end
+  end
+
+  # Uses translations key or a fallback, if the first given key is translated in the current language
+  #
+  # Usage:
+  #
+  # # yml:
+  # build_a_vehicle_with_four_wheels: Build a vehicle with four wheels.
+  # build_a_car: Build a car.
+  # build_a_vehicle_that_flies: Build a vehicle that flies.
+  # build_an_airplane: ~
+  # count_apples: "There are #{count} appels"
+  # apples:
+  #   one: "#{count} apple"
+  #   other:
+  #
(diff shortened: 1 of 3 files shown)


app/helpers/translation_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+  #
+  # # haml:
+  #
+  # use_new_translation("build_a_car").or_fallback_to("build_a_vehicle_with_four_wheels")
+  # => "Build a car."
+  #
+  # use_new_translation("build_an_airplane").or_fallback_to("build_a_vehicle_that_flies")
+  # => "Build a vehicle that flies."
+  #
+  # use_new_translation("apples", count: "1").or_fallback_to("count_apples")
```
lyyder: Should the interpolation parameter `count` be passed also to `.or_fallback_to`?

app/helpers/translation_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+  # => "1 apple"
+  #
+  # use_new_translation("apples", count: "5").or_fallback_to("count_apples")
+  # => "There are 5 apples"
+  #
+  def use_new_translation(tr_key, opts = {})
+    begin
```
sktoiva: This `begin` can be removed as it is redundant with the method body.

app/helpers/translation_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+      translation =
+        I18n.translate(tr_key, opts.merge(
+                         # Throw error, if not found
+                         # Disable fallbacks (no idea why the value needs to be `true`
+                         # instead of `false`. Feels counter intuitive)
+                         throw: true,
+                         fallback: true,
```
sktoiva: Remove trailing comma or disable the trailing comma check in `rubocop.yml`

PR #2498, opened 2016-09-02
Fix an error where an email is sent to nil address

sktoiva:
- Refactor `confirmations_controller#create`
 - At the same time fix an issue where user has already confirmed an email address but is still on the confirm page and clicks confirm again

## How to reproduce

1. Sign up as a new user
2. Leave the confirmation page open
3. Confirm email
4. In the still open confirmation page click `Resend confirmation instructions`

*Expected:* User is redirected to search page
*Actual*: An exception is thrown

### DIFF
--- a/app/controllers/confirmations_controller.rb
+++ b/app/controllers/confirmations_controller.rb
@@ -5,36 +5,22 @@ class ConfirmationsController < Devise::ConfirmationsController
               :ensure_consent_given,
               :ensure_user_belongs_to_community
 
-  # This is directly copied from Devise::ConfirmationsController
+  # This is overridden from Devise::ConfirmationsController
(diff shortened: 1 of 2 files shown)


app/controllers/confirmations_controller.rb
```diff
@@ -76,4 +63,29 @@ def show
...
+  private
+
+  def create_or_find_email(email)
+      if new_email_address_sent?(email)
+        # If user submitted the email change form, change the email before sending again.
+        create_new_email(email)
+      else
+        email_to_confirm = @current_user.latest_pending_email_address(@current_community)
+        Email.find_by_address_and_community_id(email_to_confirm, @current_community.id)
+      end
```
lyyder: Indentation is a bit off here.

PR #2502, opened 2016-09-05
This change allows to add Google Search Console verification tag in CLP.

Luis-RG:
Right now it is not possible to verify Google Search console for our CLP customers, adding this change in the meta file and updating the structure file for the CLP that require it, will allow us to add this tag. Very simple change that does not affect the customers that don't have it.

### DIFF
--- a/app/services/custom_landing_page/example_data.rb
+++ b/app/services/custom_landing_page/example_data.rb
@@ -309,7 +309,8 @@ JSON
       "description": {"type": "marketplace_data", "id": "description"},
       "publisher": {"type": "marketplace_data", "id": "name"},
       "copyright": {"type": "marketplace_data", "id": "name"},
-      "facebook_site_name": {"type": "marketplace_data", "id": "name"}
+      "facebook_site_name": {"type": "marketplace_data", "id": "name"},
+      "google_site_verification": {"value": "CHANGEME"}
     },
     "sections": [
         {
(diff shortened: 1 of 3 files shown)


app/views/landing_page/_meta.erb
```diff
@@ -61,3 +61,6 @@
   <meta name="twitter:description" content="<%= p["description"]["value"] %>" />
 <% end %>
 
+<% if p["google_site_verification"] %>
+  <meta name="google_site_verification" content="<%= p["google_site_verification"]["value"] %>" />
```
Gnito: I'm not entirely sure, but I think this might cause "undefined" problems, and therefore it would be better to have this line instead:
```  <meta name="google_site_verification" content="<%= p.dig("google_site_verification", "value") %>" />```

ping @rap1ds and @ovan

Comment (Luis-RG) [inline app/views/landing_page/_meta.erb]: @Gnito Since I am not very good at code I just "copied" it and modified from the line above for copyright that is : 
`<% if p["copyright"] %>
  <meta name="copyright" content="<%= p["copyright"]["value"] %>" />
<% end %>`
I tried this code in preview and it works... but I don't know if it will work in prod :)

File: app/services/custom_landing_page/example_data.rb [hunk 2]
```diff
@@ -19,7 +19,8 @@ module ExampleData
     "description": {"type": "marketplace_data", "id": "description"},
     "publisher": {"type": "marketplace_data", "id": "name"},
     "copyright": {"type": "marketplace_data", "id": "name"},
-    "facebook_site_name": {"type": "marketplace_data", "id": "name"}
+    "facebook_site_name": {"type": "marketplace_data", "id": "name"},
+    "google_site_verification": {"value": "CHANGEME"}
```
Gnito: Does this need to be also on template_string (~line 313)?

PR #2503, opened 2016-09-05
Add sitemap host to the path

rap1ds:


config/routes.rb
```diff
@@ -11,8 +12,19 @@
...
+  # this is because the dot is used as a separator for formatted
+  # routes. If you need to use a dot within a dynamic segment add a
+  # constraint which overrides this – for example :id => /[^\/]+/
+  # allows anything except a slash.
+  #
+  # That's why there's the constraints in generate URL to accept host
+  # parameter with dots
+  #
+  get "/sitemap.xml.gz"                        => "sitemap#sitemap", format: :xml
+  get "/sitemap/:sitemap_host/generate.xml.gz" => "sitemap#generate", format: :xml, :constraints => { :sitemap_host => /[^\/]+/ }
```
bladealslayer: How about making the regex a bit more tight? E.g. `/[.-\w]+/` (dash, alphanum and dot)?

rap1ds: Good idea!

PR #2504, opened 2016-09-05
More space for keyword search

Gnito:
We wanted to show at least 20 characters on default mode and search was squeezed narrower on responsive layout between 1200px - 1270px (63% wasn't enough).
1200px - 1270px: no slide effect
1270++: slide effect


client/app/components/composites/SearchBar/SearchBar.css
```diff
@@ -184,14 +184,18 @@
   }
 
   &:focus {
-    @media (--large-viewport) {
+    @media (--search-desktop-wide) {
```
mporkola: What's this media query, and why the change?

Gnito: We need to squeeze the search so that it fits 1200px layout. 
(So, that's 1270px breakpoint where we take 70 pixels away from Search component)

PR #2505, opened 2016-09-06
Add rel=next and rel=prev links

rap1ds:
For better crawlability, Google recommends adding `rel=next` and `rel=prev` links to paginated content. This is especially important for infinite scroll, because the crawlers may not trigger the proper scroll events to see the content from the next pages.

This Pull Request adds those links to the "homepage" a.k.a "search page".

Read more:

* https://webmasters.googleblog.com/2014/02/infinite-scroll-search-friendly.html
* https://support.google.com/webmasters/answer/1663744


mporkola:
I'm guessing the failing tests aren't related, but otherwise ok 👍

PR #2506, opened 2016-09-06
Add sitemap to robots.txt

rap1ds:


lib/rack_middleware/robots_generator.rb
```diff
@@ -5,13 +5,16 @@ class RobotsGenerator
   # Use the config/robots.txt in production.
   # Disallow everything for all other environments.
   def self.call(env)
+    return [404, {}, []] if env[:current_marketplace].nil?
```
ovan: Why not render the no_index_content in this case?

lib/rack_middleware/robots_generator.rb
```diff
@@ -5,13 +5,16 @@ class RobotsGenerator
   # Use the config/robots.txt in production.
```
ovan: This comment is no longer up to date.

lib/rack_middleware/robots_generator.rb
rap1ds: Isn't 404 more accurate information? "This page does not exists" vs "This page does exist but I don't want you to index it"

ovan: Yes, I think it makes sense. The env[:current_marketplace] shouldn't be nil in any sensible scenario.

PR #2507, opened 2016-09-06
Fetch home page data from Discovery API

mporkola:
Use data from Discovery API to populate new browse view.

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -157,8 +157,6 @@ class HomepageController < ApplicationController
     dropdowns = filter_params[:custom_dropdown_field_options].map { |dropdown_field| dropdown_field.merge(type: :selection_group, operator: :or) }
     numbers = numeric_search_params.map { |numeric| numeric.merge(type: :numeric_range) }
 
-    marketplace_configuration = MarketplaceService::API::Api.configurations.get(community_id: @current_community.id).data
-
     search = {
       # Add listing_id
       categories: filter_params[:categories],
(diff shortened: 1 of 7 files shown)


app/controllers/homepage_controller.rb
```diff
@@ -7,28 +7,32 @@ class HomepageController < ApplicationController
...
-    # only 1 sub share type, as that would make the listing type menu visible and it would look bit silly
-    listing_shape_menu_enabled = all_shapes.size > 1
-    @show_categories = @categories.size > 1
-    show_price_filter = @current_community.show_price_filter && all_shapes.any? { |s| s[:price_enabled] }
+    redirect_to landing_page_path and return if no_current_user_in_private_clp_enabled_marketplace?
 
-    filters = @current_community.custom_fields.where(search_filter: true).sort
-    @show_custom_fields = filters.present? || show_price_filter
-    @category_menu_enabled = @show_categories || @show_custom_fields
+    @homepage = true
```
mporkola: No change in functionality above, just reorganised to not execute needless code when using Discovery API view.

app/controllers/homepage_controller.rb
```diff
@@ -61,13 +65,17 @@ def index
...
     end
 
-    if request.xhr? # checks if AJAX request
+    if FeatureFlagHelper.search_engine == :discovery
+      search_result.on_success { |listings|
+        render template: "search_page/search_page", locals: { bootstrapped_data: listings }
+      }.on_error {
+        render nothing: true, status: 500
+      }
+    elsif request.xhr? # checks if AJAX request
```
mporkola: New browse view paging is not designed yet, but it probably won't use this code path.

app/services/listing_index_service/search/discovery_adapter.rb
```diff
@@ -0,0 +1,73 @@
...
+       :'filter[price_max]' => Maybe(original[:price_cents]).map{ |p| p.max }.or_else(nil),
+       :'filter[omit_closed]' => !original[:include_closed],
+       :'filter[listing_shape_ids]' => Maybe(original[:listing_shape_ids]).join(",").or_else(nil),
+       :'filter[category_ids]' => Maybe(original[:categories]).join(",").or_else(nil),
+       :'search[locale]' => original[:locale],
+       :sort => original[:sort]
+      }.merge(location_params).merge(custom_fields).compact
+    end
+  end
+end
```
mporkola: Same functionality as in `ZappyAdapter` on query side. Extracted to separate adapter for cleaner response handling, also the query will change. `ZappyAdapter` will probably be deprecated in the future.

Gnito: 👍 
We need to write something about this change to changelog.md. 
- DiscoveryAdapter might still change when Discovery API changes. 
- Change to pure DiscoveryAdapter (full deprecation) might still take a year or so (depends on how long we allow marketplaces to use current search page over the new one).

app/services/feature_flag_service/store.rb
```diff
@@ -19,6 +19,7 @@ class FeatureFlag
     FLAGS = [
       :export_transactions_as_csv,
       :topbar_v1,
+      :discovery_api
```
Gnito: This flag could be named to 'searchpage_v1' or something else since we are not going to be able to only turn discovery_api _on_. I.e. There's nothing that would be able to handle/render the results separately.

app/controllers/homepage_controller.rb
```diff
@@ -7,28 +7,32 @@ class HomepageController < ApplicationController
...
+      @show_custom_fields = filters.present? || show_price_filter
+      @category_menu_enabled = @show_categories || @show_custom_fields
+    end
 
-    # This assumes that we don't never ever have communities with only 1 main share type and
-    # only 1 sub share type, as that would make the listing type menu visible and it would look bit silly
-    listing_shape_menu_enabled = all_shapes.size > 1
-    @show_categories = @categories.size > 1
-    show_price_filter = @current_community.show_price_filter && all_shapes.any? { |s| s[:price_enabled] }
+    redirect_to landing_page_path and return if no_current_user_in_private_clp_enabled_marketplace?
```
Gnito: These kind of redirections should be the first line I think - no need to execute the function any further

PR #2509, opened 2016-09-06
This PR fixes warning: already initialized constant warnings

rap1ds:


spec/services/paypal_service/store/paypal_payment_spec.rb
```diff
@@ -1,6 +1,5 @@
 describe PaypalService::Store::PaypalPayment do
   order = {
-    order_total: 100,
```
bladealslayer: Why was this changed?

rap1ds: This fixes a warning also. There's a `order_total` redefined in line 12, and that line overrides this definition.

PR #2511, opened 2016-09-07
Sharetribe branding component

mporkola:
React-based branding component for non-whitelabel marketplaces

### DIFF
--- a/client/app/components/composites/Branding/Branding.css
+++ b/client/app/components/composites/Branding/Branding.css
@@ -1,0 +1,10 @@
+.brandingContainer {
+  border-top: 1px solid var(--Branding_borderColor);
+  margin: 40px 0;
+  padding-top: 40px;
+}
+
+.brandingContent {
+  color: var(--Branding_textColor);
+  text-align: center;
+}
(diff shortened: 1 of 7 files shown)


Gnito:
Do you have link to styleguide? It could be added here and shared with Juho & Janne too.

client/app/components/composites/Branding/Branding.js
```diff
@@ -0,0 +1,26 @@
...
+import { t } from '../../../utils/i18n';
+
+import css from './Branding.css';
+
+export default function Branding({ linkToSharetribe }) {
+  const learnMoreLink = a({ href: linkToSharetribe }, t('web.branding.learn_more'));
+  const sharetribeLink = a({ href: linkToSharetribe }, 'Sharetribe');
+
+  return div({
+    className: css.brandingContainer,
```
Gnito: By not giving classname handles, it makes harder for endusers to style or hide this stuff - which is probably OK?

client/app/components/composites/Branding/Branding.css
```diff
@@ -0,0 +1,10 @@
+.brandingContainer {
+  border-top: 1px solid #EBEBEB;
```
Gnito: Also: 
app/components/composites/Branding/Branding.css
 2:25  ✖  Expected "#EBEBEB" to be "#ebebeb"   color-hex-case

client/app/components/composites/Branding/Branding.js
mporkola: Yeah, didn't think custom styling this is something we want :)

PR #2514, opened 2016-09-07
Listing shape availability

rap1ds:
This PR adds availability calendar options to Order Types admin UI.

**Changes to datamodel**

A new column `availability` is added to `listing_shapes` table. Possible values for `availability` are `none` (default) and `booking`.

**UI changes**

A new section "Availability" is now added to Order Types admin UI. The availability can be enabled/disabled and doing so will enable/disable pricing units. In addition to choosing to use availability, the admin can choose whether "per day" or "per night" availability is used.

In the UI, we have two different "units". We have the "availability unit" (per day, per night, only one can be chosen) and "pricing units" (day, night, hour, week, etc. many can be chosen). However, in the data model, we have only the "pricing units". If availability is enabled (value is `booking`), we know that the unit `night` means the "per night availability unit". Because of this, we do some mapping on the controller, where we map UI value to data model values and vice versa.

**Screenshot**

screen shot 2016-09-08 at 10 24 18

### DIFF
--- a/app/assets/javascripts/admin/listing_shapes.js
+++ b/app/assets/javascripts/admin/listing_shapes.js
@@ -153,7 +153,11 @@ window.ST.initializeListingShapeForm = function(formId) {
   var toggleRadio = function(el, state) {
     if(state) {
       el.prop('disabled', false);
-      el.first().prop('checked', true);
+
+      // Check the first one if none of the radiobuttons is checked
+      if (!el.is(":checked")) {
+        el.first().prop('checked', true);
+      }
     } else {
       el.prop('disabled', true);
       el.prop('checked', false);
--- a/app/assets/javascripts/admin/listing_shapes.js
+++ b/app/assets/javascripts/admin/listing_shapes.js
@@ -86,11 +86,9 @@ window.ST.initializeListingShapeForm = function(formId) {
     if(enabled) {
       toggleAvailabilityUnitsEnabled(true);
       toggleUnitsEnabled(false);
-      $('.js-pricing-units-disabled-info').show();
     } else {
       toggleAvailabilityUnitsEnabled(false)
       toggleUnitsEnabled(true);
-      $('.js-pricing-units-disabled-info').hide();
     }
   }
 
@@ -107,6 +105,7 @@ window.ST.initializeListingShapeForm = function(formId) {
   var toggleUnitsEnabled = function(enabled) {
     toggleCheckbox($(".js-unit-checkbox"), enabled);
     toggleLabel($(".js-unit-label"), enabled);
+    toggleInfo($('.js-pricing-units-info'), enabled);
   };
 
   var toggleAvailabilityEnabled = function(enabled) {
@@ -117,6 +116,7 @@ window.ST.initializeListingShapeForm = function(formId) {
   var toggleAvailabilityUnitsEnabled = function(enabled) {
     toggleRadio($(".js-availability-unit"), enabled);
     toggleLabel($(".js-availability-unit-label"), enabled);
+    toggleInfo($('.js-pricing-units-disabled-info'), enabled)
   };
 
   var removeCustomUnit = function() {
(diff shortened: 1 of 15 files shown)


app/assets/javascripts/admin/listing_shapes.js
```diff
@@ -116,15 +141,37 @@ window.ST.initializeListingShapeForm = function(formId) {
     this.parentElement.remove();
   };
 
-  var toggle = function(el, state) {
+  var toggleCheckbox = function(el, state) {
```
kpuputti: I don't understand this function. Is it only there to enable/disable the checkbox for user to check/uncheck, and unchecks when disabled to avoid submitting the value?

The name leads me to think it would toggle the value, not the disabled/enabled state.

db/migrate/20160907095103_add_availability_to_listing_shape.rb
```diff
@@ -0,0 +1,5 @@
+class AddAvailabilityToListingShape < ActiveRecord::Migration
+  def change
+    add_column :listing_shapes, :availability, :string, limit: 32, default: "none", after: :shipping_enabled
```
kpuputti: This column accepts only a predefined set of strings, is it possible to make it an enumeration?

app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -344,14 +341,57 @@ def params_to_shape(params)
...
+
+    # Combines availability units and pricing units to just "units".
+    #
+    # In the backend we have only one concept, units. However, in the UI
+    # we are showing pricing unit checkboxes and availability unit radio
+    # buttons. This method maps the UI units to backend units.
+    #
+    def parse_units_from_params(form_params)
+      selected_predefined_units =
+        if form_params[:availability] == "booking"
```
ovan: Why is the comparsion against a string "booking" instead of symbol :booking? Shouldn't the service always return these as symbols?

ovan: Noticed now this is actually coming from the form post. In that case, the form validator should transform to symbols, right?

rap1ds: The caller of this function puts the return value through the Entity, which will make the conversion: https://github.com/sharetribe/sharetribe/pull/2514/files#diff-710ec5b3d9350434e1486006c0540ebfL341

db/migrate/20160907095103_add_availability_to_listing_shape.rb
rap1ds: You mean [ActiveRecord Enum](http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html)? Hmm, interesting. I have never used those. I'll check it out!

ovan: A MySQL ENUM? We haven't been using that anywhere. Quick google revealed it's a bit controversial topic. The main worries are perf and hard to add new types. I'd feel it's not worth it at least right now but would take more studying if it's a column type we want to use or not.

rap1ds: I read quickly about ActiveRecord Enum. It works so that it emulates the native enum type by converting the allowed values to integers. That's quite ok I guess, but then we lose some clarity (i.e. not easy to remember by just looking at the database value what does 3 or 4 stand for).

I don't thing the native MySQL enum type is supported by Rails.

app/assets/javascripts/admin/listing_shapes.js
rap1ds: Good point. I'll change the name.

app/controllers/admin/listing_shapes_controller.rb
ovan: Ideally it should be in different order so that the "logic" can always consistently deal with symbols. Now it adds a bit of complexity that mapping to one way uses symbols and to other strings. But it's not a big deal.

db/migrate/20160907095103_add_availability_to_listing_shape.rb
ovan: It loses the self describing nature of the enum which affects things like migrations and adhoc debugging. I don't see a perf gain reason here to do so (we don't load listing types by enum value and even then it's just a short varchar which works ok with indexes).

PR #2517, opened 2016-09-08
Add migration for square listing card images

lyyder:
New listings browsing view requires two new image sizes:

 - square: __408x408__
 - square_2x: __816x816__

These images are created with the following actions:

 - Add the new image sizes for every new listing image [#2518](https://github.com/sharetribe/sharetribe/pull/2518)
 - Create new images for all listings that are open and valid
 - If a listing is reopened, reprocess it's images to make sure it has all the required image sizes [#2518](https://github.com/sharetribe/sharetribe/pull/2518)

__Notice!__  [#2518](https://github.com/sharetribe/sharetribe/pull/2518) has to be merged before this PR.


db/migrate/20160908062011_process_images_for_listing_cards.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def values(ids)
+    ids.map { |id|
+      "(1, #{handler(id)}, NULL, NOW(), NOW(), NOW(), 'image_reprocess')"
+    }.join(",")
+  end
+
+  def handler(id)
+    "'--- !ruby/struct:CreateSquareImagesJob\nimage_id: #{id}\n'"
+  end
```
ovan: This feels a bit manual. There's utils in YAML module that do serialization in standard way.

db/migrate/20160908062011_process_images_for_listing_cards.rb
```diff
@@ -0,0 +1,25 @@
+class ProcessImagesForListingCards < ActiveRecord::Migration
+
+  def up
+    select_values("SELECT id FROM listing_images").each_slice(1000) { |ids|
+
+      exec_insert("INSERT INTO delayed_jobs
+                  (priority, handler, last_error, run_at, created_at, updated_at, queue)
+                  VALUES #{values(ids)}", "create_delayed_jobs", [])
```
rap1ds: I'm wondering... What's the maximum size for one MySQL insert query? This will be massive. I hope it doesn't crash because it goes over some limit.

rap1ds:
The reprocessing process should be documented in UPGRADE.md so that people know how to do it.

Procfile
```diff
@@ -1,2 +1,3 @@
 web:         bundle exec passenger start -p $PORT --max-pool-size $PASSENGER_MAX_POOL_SIZE
 worker:      QUEUES=default,paperclip,mailers bundle exec rake jobs:work
+image_reprocess_worker: QUEUE=image_reprocess bundle exec rake jobs:work
```
rap1ds: Is this necessary? Why not just add to the normal worker queue with low enough priority? (11)

db/migrate/20160908062011_process_images_for_listing_cards.rb
rap1ds: Is there a risk that the massive insert locks the table for too long time? That would have immediate effect on the application performance.

rap1ds: Ooookey... forget everything, noticed now the `each_slice` there.

db/migrate/20160908062011_process_images_for_listing_cards.rb
lyyder: 👍

Procfile
lyyder: @rap1ds A specific queue makes it easier to recognize when all images have been processed in order to minimize the time extra workers are required. Necessary? Hardly.

lyyder: @sktoiva should we drop the reprocess-specific queue to simplify upgrade?

sktoiva: @lyyder, your show, what do you think?

lyyder: Right on, I'll remove the queue and use the default one.

PR #2518, opened 2016-09-08
Add new listing image sizes for listing cards

lyyder:
Add two new image sizes:

 - square: __408x408__
 - square_2x: __816x816__

which are used in the listing cards view.


app/controllers/listings_controller.rb
```diff
@@ -945,4 +946,12 @@ def delivery_price_hash(delivery_type, price, shipping_price_additional)
         default: true
       }
   end
+
+  # Create image sizes that might be missing
+  # from a reopened listing
+  def create_images(listing)
+    listing.listing_images.pluck(:id).each { | image_id |
```
lyyder: This seems to generate all images styles, regardless of the style parameters.

PR #2519, opened 2016-09-08
Add UUIDs for listings

kpuputti:
This PR adds a Version 1 [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) (sortable due to timestamp) to listings. UUIDs are stored as `BINARY(16)` in the Database to be efficient with storage.

UUIDs are generated automatically for new listings, and reading a UUID from a listing makes the binary -> `UUIDTools::UUID` conversion automatically. Migrations to add UUIDs to existing listings will come in separate PRs as well as constraints for uniqueness and non-nullness of the column value.


app/utils/uuid_utils.rb
```diff
@@ -0,0 +1,12 @@
+module UUIDUtils
+  module_function
+
+  def bin16_uuid
+    # TODO: Use https://github.com/assaf/uuid
+    s = SecureRandom.uuid.gsub("-", "")
```
ovan: Did you @kpuputti check somewhere that these are v1 UUIDs? My understanding has been they'd be random (v4), which doesn't work for clustering by time.

kpuputti: @ovan That's exactly why there is the TODO comment above. These are indeed v4 UUIDs and should be changed to use the linked library or something else.

db/schema.rb
```diff
@@ -509,6 +509,7 @@
   add_index "listing_units", ["listing_shape_id"], name: "index_listing_units_on_listing_shape_id", using: :btree
 
   create_table "listings", force: :cascade do |t|
+    t.binary   "uuid",                            limit: 16
```
kpuputti: This will most likely be a problem. Using the `:binary, limit: 16` type in the migration generated a column of type `VARBINARY`, which is not what we want. However, when setting up a new DB from this schema, I expect Rails to create a `VARBINARY` column again, assuming that the `:binary` -> `VARBINARY` mapping always holds.

kpuputti: Here is the relevant MySQL adapter code that converts the Rails types to MySQL types:

https://github.com/rails/rails/blob/4b5570bd1bdd6c6750ae06ace1806f9cafd68f12/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb#L573

@ovan  Should we just use the `VARBINARY` type generated by the `:binary, limit: 16` Rails type?

PR #2520, opened 2016-09-08
Harmony Client

rap1ds:


app/utils/service_client/client.rb
```diff
@@ -0,0 +1,46 @@
...
+      endpoint_url = @_endpoints[endpoint]
+
+      raise ArgumentError.new("Couldn't find endpoint '#{endpoint}'") if endpoint_url.nil?
+
+      ctx = @_context_runner.execute(
+        url: endpoint_url,
+        method: method,
+        params: params,
+        opts: opts
+      )
```
ovan: (sorry if I'm commenting too early work to be commented)

There's no reason to lose the logical endpoint name on this level. Down the middleware chain some middleware might like to use it (e.g. validate params based on operation id in swagger schema).

app/utils/service_client/client.rb
```diff
@@ -0,0 +1,46 @@
...
+        Result::Success.new(
+          status: res[:status],
+          body: res[:body]
+        )
+      else
+        Result::Error.new(res[:body],
+          status: res[:status],
+          body: res[:body]
+        )
+      end
```
ovan: I think I'd like to see the result wrapping as a middleware instead of as hard coded top level code. That way you can also generically handle errors raised in request processing (raising triggers the error flow and middleware can handle both leave and error), so whether the error is because of an error response from remote or because of timeout it comes out as Result::Error.

app/utils/service_client/middleware/logger.rb
```diff
@@ -0,0 +1,24 @@
...
+        @logger.info("Leave", :leave, ctx)
+        ctx
+      end
+
+      def error(ctx)
+        @logger.info("Error", :error, ctx)
+        ctx
+      end
+    end
+  end
```
ovan: I think a more useful logger implementation would be something that understands the request and response maps and can actually log what's happening. How about placing it close to end of queue where response is already built so you can log what kind of request will be made to which endpoint. I would also appreciate logging the time it takes from request to response. This is easy to do by putting the timestamp in the ctx in enter and logging diff in leave/error

app/utils/service_client/middleware/middleware_base.rb
```diff
@@ -0,0 +1,36 @@
+module ServiceClient
+  module Middleware
+
+    # Base class for middlewares.
+    #
+    # Default implementations for `enter`, `leave` and `error`
+    # return the context object `ctx` without manipulating it.
+    # The subclasses can override all or some of these methods.
```
ovan: Before invoking a lifecycle method like enter, leave or error the context runner implementation checks if a middleware implements it. At each stage, only middleware that respond to the lifecycle method in question is invoked. Inheriting from this class breaks the idea because now everybody implement every lifecycle method. I would suggest making this a marker class only and either not inherit from it or then leave out these default implementations of leave, enter and error.

app/utils/service_client/middleware/request_id.rb
```diff
@@ -0,0 +1,11 @@
+module ServiceClient
+  module Middleware
+    class RequestID < MiddlewareBase
+      def enter(ctx)
+        headers = ctx.fetch(:req).fetch(:headers)
```
ovan: fetch doesn't buy much here because the next line already expected :reg to be defined. I feel it's good to let middleware like this fail when ctx is not of expected shape. That can only mean a configuration error (middleware is incorrectly ordered or missing) so it's good to expose it in the dev phase as early as possible.

app/utils/service_client/middleware/body_encoder.rb
```diff
@@ -0,0 +1,90 @@
...
+        ctx[:req][:body] = @_encoder.encode(body)
+        ctx[:req][:headers]["Accept"] = @_encoder.mime_type
+        ctx[:req][:headers]["Content-Type"] = @_encoder.mime_type unless body.nil?
+
+        ctx
+      end
+
+      def leave(ctx)
+        res = ctx.fetch(:res)
+        body = res[:body]
```
ovan: The encoder is now chosen statically instead of using the responses content-type header. I think this is ok because we expect the API to work like that when we specifically set the Accept header ourselves. Still, I suggest adding a validation here that checks the responses content-type value and ensures it is of expected type. That will fail more meaningfully than having the decoding process raise some random parsing error.

app/utils/service_client/middleware/middleware_base.rb
rap1ds: We could of course change the runner so that it always calls enter, leave and error method, without checking. That would work, if all the middleware inherit the base class. Or do you prefer keeping the runner implementation as it is now?

app/utils/service_client/middleware/body_encoder.rb
```diff
@@ -0,0 +1,90 @@
...
+
+        begin
+          ctx[:res][:body] = @_encoder.decode(body)
+        rescue StandardError => e
+          raise ParsingError.new("Parsing error, msg: '#{e.message}', body: '#{body}'")
```
ovan: I was about to comment that it makes it difficult for debugging to swallow the original exception and re-raise as something else. But then I googled a bit and it seems that as of Ruby 2.1 there's something called cause which links an error to the original causing error. And also that the cause is automatically put in place by raise by looking at the context where raise happens. Magic.

app/utils/service_client/middleware/result_mapper.rb
```diff
@@ -0,0 +1,32 @@
+module ServiceClient
+  module Middleware
+    class ResultMapper < MiddlewareBase
+
+      def leave(ctx)
+        res = ctx.fetch(:res)
+
+        res_object =
+          if res.fetch(:success)
```
ovan: res.fetch makes no sense because in both branches of if-else we rely on res being defined and call res[:status] etc. directly. So just `res[:success]` here.

app/utils/service_client/middleware/request_id.rb
rap1ds: There are couple benefits:

1. If `req` is `nil`, we get more informative error message: `KeyError: key not found: :req` vs `NoMethodError: undefined method `[]' for nil:NilClass`
2. If `headers` is `nil`, we get more informative error message: ``KeyError: key not found: :headers` vs `NoMethodError: undefined method `merge' for nil:NilClass`

It's true that the method is so shorts that even a bad error message would point easily to the issue, but does fetch do some harm?

app/utils/service_client/middleware/middleware_base.rb
ovan: I prefer middleware that only implements what it needs. It's unnecessary to call the pass though implementations and it makes is possible to look at middleware and dynamically tell what type of middleware it is (might be useful for debugging, logging, validating something, etc.).

app/utils/service_client/middleware/request_id.rb
ovan: Right, makes sense. I had an incorrect assumption that fetch wouldn't raise but instead returns a nil when the key is not defined (that's how get works in Clojure). But the fact that it raises when no default is given changes things.

app/utils/service_client/middleware/body_encoder.rb
rap1ds: Wow. Magic. I didn't know that.

I added the begin-rescue there and the custom exception so that I can have the original body in the error message. This makes debugging easier when you see the string that caused the parsing to fail.

PR #2521, opened 2016-09-08
Add support for passing resource limits to ImageMagick 'convert'

bladealslayer:


config/config.defaults.yml
```diff
@@ -253,6 +253,18 @@ default: &default_settings
   # Max filesize for uploaded image (bytes)
   max_image_filesize: 104857600
 
+  # Memory limit for image processing - passed to ImageMagick -limit memory (MAGICK_MEMORY_LIMIT)
+  # See http://www.imagemagick.org/script/resources.php
+  # Example value:
+  # image_processing_memory_limit: 512MiB
+  image_processing_memory_limit:
```
rap1ds: Could we add here a sensible default value?

config/config.defaults.yml
```diff
@@ -253,6 +253,18 @@ default: &default_settings
...
+
+  # Memory map limit for image processing - passed to ImageMagick -limit map (MAGICK_MAP_LIMIT)
+  # See http://www.imagemagick.org/script/resources.php
+  # Example value:
+  # image_processing_map_limit: 512MiB
+  image_processing_map_limit:
```
rap1ds: Same here

config/config.defaults.yml
bladealslayer: I had trouble deciding what the "sensible" defaults would be so left the default behaviour unchanged. What do you think?

rap1ds: Understood. Leave it empty then.

PR #2528, opened 2016-09-12
Update React Storybook to 2.13.0

mporkola:


Gnito:
Looks good - although, the version seems to have progressed last week:
https://github.com/kadirahq/react-storybook/blob/master/CHANGELOG.md

PR #2530, opened 2016-09-13
ListingCard

Gnito:
TODO:
- [x] Calculate distance to 2 significant numbers
- [x] Icon for no-image & errored image
- [x] Icon for distance
- [x] ```<a>``` tag to title (and image)
- [x] styles to variables.js
- ~~2x images + picture element~~ (in different branch)
- ~~refactor passed props~~ (in different branch when new data is ~ready)
- ~~Format price & unit according to our own specifications~~ (in different branch)
- ~~Avatar icon missing.~~ (in different branch)


client/app/components/composites/ListingCard/ListingCard.js
```diff
@@ -0,0 +1,144 @@
...
+import noImageIcon from './images/noImageIcon.svg';
+import distanceIcon from './images/distanceIcon.svg';
+
+const MINIMUM_DISTANCE = 0.1;
+const PRECISION = 2;
+const TINT_PERCENTAGE = 20;
+
+const sigFigs = function sigFigs(n, sig) {
+  return parseFloat(n.toPrecision(sig));
+};
```
mporkola: We should probably use numeral.js or something for formatting, to get thousands separators etc correct. But we can do that also later.

client/app/components/composites/ListingCard/ListingCard.js
```diff
@@ -0,0 +1,144 @@
...
+  handleImageErrored() {
+    this.setState({ imageStatus: 'failed' }); // eslint-disable-line react/no-set-state
+  }
+
+  render() {
+
+    const tintedRGB = tint(this.props.color, TINT_PERCENTAGE);
+    const hasDistance = !!this.props.distance;
+    const precision = (hasDistance && this.props.distance < 1) ? 1 : PRECISION;
+    const distanceFormatted = (hasDistance && this.props.distance < MINIMUM_DISTANCE) ? `< 0.1${this.props.distanceUnit}` : `${sigFigs(this.props.distance, precision)}${this.props.distanceUnit}`;
```
mporkola: space between value and unit!

client/app/components/composites/ListingCard/ListingCard.js
Gnito: Thomas et al. were creating some kind of country based format mapping. When we implement that, we need to figure out if numerals.js is good or extra weight for page load.

client/app/components/composites/ListingCard/ListingCard.css
```diff
@@ -0,0 +1,125 @@
...
+  left: 0;
+  right: 0;
+  width: 100%;
+  height: 100%;
+  border-top-left-radius: inherit;
+  border-top-right-radius: inherit;
+}
+
+.noImageContainer,
+.noImagaWrapper {
```
mporkola: I'm guessing it should be `noImageWrapper`

client/app/components/composites/ListingCard/ListingCard.story.js
```diff
@@ -0,0 +1,89 @@
...
+              distance: 0.02,
+              distanceUnit: 'km',
+              color: '#347F9D',
+              className: css.listing,
+            },
+          )),
+        ]),
+        containerStyle
+      ))
+  ));
```
mporkola: I think the component story should include simply the component, one state per story. Now, this looks more like a Listing card container story, and makes it a bit hard to find out what the actual component by itself looks like.

client/app/components/composites/ListingCard/ListingCard.story.css
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  @media screen and (min-width: 1440px) {
+    flex-basis: calc(20% - 24px);
+  }
+
+  @media screen and (min-width: 1724px) {
+    flex-basis: calc(100% / 6 - 24px);
+  }
+}
```
mporkola: This looks like it should be in container, instead of card story

client/app/components/composites/ListingCard/ListingCard.story.js
Gnito: I'm trying to show the different states the component has. 
It is much easier to notice problems when you make changes to component and you see different components with one glance.
See ref: http://backpack.prod.aws.skyscnr.com/docs/atoms/buttons

Gnito: So, I actually suggest that we go to that direction.

client/app/components/composites/ListingCard/ListingCard.story.css
Gnito: We could turn this story to be part of container story when we have container.

PR #2531, opened 2016-09-13
Switch to :sql schema format

kpuputti:
This PR changes the `config.active_record.schema_format` configuration to use `:sql` instead of `:ruby` and thus generating the DB schema dump in `db/structure.sql` instead of `db/schema.rb`.

Using the :sql schema format instead of :ruby allows us to use MySQL specific functionality in migrations etc. and since we will only support MySQL, the abstraction provided by the schema.rb is not needed.

One use case for this is the need to use the `BINARY` datatype in MySQL instead of the `VARBINARY` datatype that Rails generates when using the `:binary` type within Ruby code.


rap1ds:
TODO: Update README.md installation instructions to use `rake db:structure:load`

HuntBurdick:
Writing SQL in migration files is a truly sadistic thing to do.. Completely breaks expected behavior of Rails. IMHO

PR #2537, opened 2016-09-13
Cleanup marketplace router

rap1ds:
Clean up the MarketplaceRouter:

- [x] Remove SSL redirection (this is already handled in EnforceSSL middleware)
- [x] Remove domain validation awareness (we don't use this method for domain validation anymore)
- [x] Remove domain validation controller
- [x] Mark the domain validation columns as ignored
- [x] Separate the redirect reason and redirect target. This is needed for example if we want to render 404 with no body for /robots.txt but redirect to some "Oops... not found" page for other cases.


app/utils/marketplace_router.rb
```diff
@@ -119,87 +120,77 @@ def needs_redirect(request:, community:, paths:, configs:, other:, &block)
...
   end
 
-  def should_use_https?(request:, configs:, community:, is_domain_verification:)
-    from_proxy = (request[:headers]["HTTP_VIA"] && request[:headers]["HTTP_VIA"].include?("sharetribe_proxy"))
-    robots = request[:fullpath] == "/robots.txt"
-
-    configs[:always_use_ssl] && !from_proxy && !robots && !is_domain_verification
+  def redirect_reason(request:, community:, configs:, other:)
+    if other[:community_search_status] == :not_found && other[:no_communities]
+      :new_marketplace
```
ovan: :no_marketplaces ?

app/utils/marketplace_router.rb
```diff
@@ -119,87 +120,77 @@ def needs_redirect(request:, community:, paths:, configs:, other:, &block)
...
+    elsif other[:community_search_status] == :not_found && !other[:no_communities]
+      :not_found
+    elsif community && community[:deleted]
+      :deleted
+    elsif community && community[:closed]
+      :closed
+    elsif community && community[:domain].present? && community[:use_domain] && request[:host] != community[:domain]
+      :domain
```
ovan: :use_domain

app/utils/marketplace_router.rb
```diff
@@ -119,87 +120,77 @@ def needs_redirect(request:, community:, paths:, configs:, other:, &block)
...
+    elsif community && community[:domain].present? && !community[:use_domain] && request[:host] == community[:domain]
+      :no_domain
```
ovan: :use_ident

PR #2538, opened 2016-09-14
Listing card 2x

Gnito:


client/app/components/composites/ListingCard/ListingCard.js
```diff
@@ -56,12 +57,12 @@ class ListingCard extends Component {
...
         href: this.props.listingURL,
       }, this.props.imageURL && this.state.imageStatus !== 'failed' ?
-        img({
+        img(Object.assign({}, {
           className: css.thumbnail,
           src: this.props.imageURL,
           onLoad: this.handleImageLoaded,
           onError: this.handleImageErrored,
-        }) :
+        }, higherRes)) :
```
mporkola: We should use spread operator (`...`) for consistency

client/app/components/composites/ListingCard/ListingCard.js
```diff
@@ -41,6 +41,7 @@ class ListingCard extends Component {
   render() {
 
     const tintedRGB = tint(this.props.color, TINT_PERCENTAGE);
+    const higherRes = this.props.image2xURL != null ? { srcSet: `${this.props.image2xURL} 2x` } : null;
```
mporkola: wouldn't `this.props.image2xURL ?` be the same, I think we use that more

client/app/components/composites/ListingCard/ListingCard.js
Gnito: I'm not sure we want to prefer spread on every cases.
Here it would mean something like:
```
{
  ...{
    className: css.thumbnail,
    src: this.props.imageURL,
    onLoad: this.handleImageLoaded,
    onError: this.handleImageErrored,
  },
  ...higherRes,
}
```
Where it is very easy to forget to do that mutation wrapping spread (object assign makes it definitely more clear.)

client/app/components/composites/ListingCard/ListingCard.js
Gnito: I'll change

client/app/components/composites/ListingCard/ListingCard.js
Gnito: But since I want to push this forward - I'll change this.

PR #2544, opened 2016-09-14
Marketplace routing

rap1ds:
This PR adds routing logic to controllers that previously didn't do any routing, i.e. `SitemapController`, `LandingpageController` (both `Metal` controllers) and to the `RobotsGenerator` routing.

The main routing logic is still in `MarketplaceRouter`. The `MarketplaceRouter` has three main entry-points:

- `redirect_reason`: This method returns a reason why we need to redirect the user. It returns `nil` if no redirection is needed.
- `redirect_target`: This method takes `reason` and some other parameters, and returns either URL or named route where the user should be redirected
- `perform_redirect`: This method is a dirty method (i.e. it knows about global `APP_CONFIG`). It's main purpose is to be a convenience method for controllers to add routing. Currently `ApplicationController` and `LandingPageController` are using this method.

This PR affects (remember to test all these):

- ApplicationController (and all controllers inheriting that)
- SitemapController
- RobotsGenerator
- LandingpageController

`ApplicationController` and `LandingPageController` use same redirection logic. They redirect use to HTML pages.

Both `RobotsGenerator` and `SitemapController` have pretty similar redirection logic. Instead of redirecting user to "Not found" HTML page, they just return 404 status.


app/utils/marketplace_router.rb
```diff
@@ -160,37 +139,126 @@ def redirect_target(reason:, request:, community:, paths:, configs:, protocol:)
...
-    elsif community && community[:domain].present? && !community[:use_domain] && request[:host] == community[:domain]
+    elsif community && community[:domain].present? && !community[:use_domain] && host == community[:domain]
       :use_ident
-    elsif community && request[:host] == "www.#{community[:ident]}.#{configs[:app_domain]}"
+    elsif community && host == "www.#{community[:ident]}.#{app_domain}"
       :www_ident
     end
   end
+
+  # This is a dirty method that is aware of the global application
```
ovan: The word dirty in my mind refers to some data being modified but not flushed. I would call this side effecting in the comment text.

lib/rack_middleware/marketplace_redirect_reason.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    redirect_reason = ::MarketplaceRouter.redirect_reason(
+      community: ::MarketplaceRouter.community_hash(marketplace, plan),
+      host: request.host,
+      no_communities: no_marketplaces,
+      app_domain: ::URLUtils.strip_port_from_host(APP_CONFIG.domain)
+    )
+
+    @app.call(env.merge!(redirect_reason: redirect_reason))
+  end
```
ovan: I would merge this to the other middleware above because this cannot exists independently.

PR #2549, opened 2016-09-15
Create bookable in Harmony API when creating a listing

kpuputti:
This PR is the first step in integrating the new Harmony API to the application.

Using a new Harmony service client, a bookable is created in Harmony before a listing is created in the application. The following scenarios are accounted for:

1. Bookable creation succeeds -> listing creation continues
2. Bookable already created -> listing creation continues
3. Bookable creation fails -> listing is not created, error is shown to the user

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -261,7 +261,20 @@ class ListingsController < ApplicationController
     params[:listing].delete("origin_loc_attributes") if params[:listing][:origin_loc_attributes][:address].blank?
 
     shape = get_shape(Maybe(params)[:listing][:listing_shape_id].to_i.or_else(nil))
+    listing_uuid = UUIDTools::UUID.timestamp_create
 
+    if FeatureFlagHelper.feature_enabled?(:availability) && shape.present? && shape[:availability] == :booking
+      bookable_res = create_bookable(@current_community.uuid, listing_uuid, @current_user.uuid)
+      unless bookable_res.success
+        flash[:error] = t("listings.error.something_went_wrong_plain")
+        return redirect_to new_listing_path
+      end
+    end
+
+    create_listing(shape, listing_uuid)
+  end
+
+  def create_listing(shape, listing_uuid)
     listing_params = ListingFormViewUtils.filter(params[:listing], shape)
     listing_unit = Maybe(params)[:listing][:unit].map { |u| ListingViewUtils::Unit.deserialize(u) }.or_else(nil)
     listing_params = ListingFormViewUtils.filter_additional_shipping(listing_params, listing_unit)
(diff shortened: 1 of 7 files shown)


app/controllers/listings_controller.rb
```diff
@@ -262,6 +262,17 @@ def create
 
     shape = get_shape(Maybe(params)[:listing][:listing_shape_id].to_i.or_else(nil))
 
+    listing_uuid = UUIDTools::UUID.timestamp_create
+    author_uuid = UUIDTools::UUID.parse_raw(Base64.urlsafe_decode64(@current_user.id))
```
rap1ds: I guess we could move the `author_uuid` inside the `if FeatureFlagHelper.feature_enabled?(:availability) ...` branch since it's not used elsewhere?

app/services/harmony_service/api/api.rb
```diff
@@ -0,0 +1,16 @@
+module HarmonyService::API
+  Api = ServiceClient::Client.new(APP_CONFIG.harmony_api_url,
```
rap1ds: It's true that we currently are using the naming format you have here: `HarmonyService::API::Api`. However, I have always had a feeling that the `::API::Api` part is a bit cumbersome and unnecessary.

So how about changing the name to just `HarmonyClient`? And then move the file to `app/services/harmony_client.rb`?

app/controllers/listings_controller.rb
ovan: How about we add a method to person class to get the uuid? I imagine this will be needed in quite many places.

app/controllers/listings_controller.rb
```diff
@@ -484,6 +496,20 @@ def verification_required
 
   private
 
+  def create_bookable(listing_uuid, author_uuid)
+    res = HarmonyService::API::Api.post(:create_bookable,
+                                        body: {
+                                          marketplaceId: @current_community.uuid,
+                                          refId: listing_uuid,
+                                          authorId: author_uuid
+                                        })
```
ovan: The indentation doesn't look very nice to read. How about an enter before :create_bookable?

app/utils/transit_utils.rb
```diff
@@ -1,9 +1,27 @@
...
+    end
+
+    def rep(u)
+      Transit::UUID.new(u.to_s)
+    end
+
+    def string_rep(u)
+      u.to_s
+    end
+  end
```
ovan: To keep the data types symmetrical we must also provide a read handler that returns UUIDTools::UUID objects instead of Transit::UUID. The recursive write handling is an ok simplification.

app/controllers/listings_controller.rb
```diff
@@ -484,6 +498,22 @@ def verification_required
...
   private
 
+  def create_bookable(listing_uuid)
+    res = HarmonyClient.post(
+      :create_bookable,
+      body: {
+        marketplaceId: @current_community.uuid,
+        refId: listing_uuid,
+        authorId: @current_user.uuid
+      })
```
ovan: One principle we've tried to keep with controllers is that only the top level method, i.e. the action, pulls data from global scope and all delegated to methods get things like marketplace and auhor as parameters. This makes it at least in theory easier to refactor the code to separate modules later and also makes it easier to follow the data flow.

PR #2550, opened 2016-09-16
Storybook specs

mporkola:
- [x] UI tests in Storybook ([example](https://sharetribe.github.io/sharetribe/storybook-specs/?selectedKind=Search%20results&selectedStory=ListingCard%20-%20image%20fail&full=0&down=1&left=1&panelRight=0&downPanel=storybook-addon-specifications%2Fspecifications-panel))
- [x] Run in CI


client/webpack.storybook.config.js
```diff
@@ -37,6 +37,14 @@ config.module.loaders.push(
   }
 );
 
+config.externals = {
```
Gnito: This could be added as a comment here:  https://github.com/airbnb/enzyme/blob/master/docs/guides/webpack.md

client/webpack.storybook.config.js
```diff
@@ -37,6 +37,14 @@ config.module.loaders.push(
...
+  jsdom: 'window',
+  cheerio: 'window',
```
Gnito: From the link above: we could possibly leave cheerio away (needs testing though)

client/webpack.storybook.config.js
```diff
@@ -37,6 +37,14 @@ config.module.loaders.push(
...
+  'react/lib/ExecutionEnvironment': true,
+  'react/lib/ReactContext': 'window',
```
Gnito: The way I understand this is that ReactContext can be required from window. The examples I found, treated this as pure external request ``` 'react/lib/ReactContext': true, ```. So, it probably works that way too.
https://webpack.github.io/docs/configuration.html#externals

client/mochaConfig.js
```diff
@@ -0,0 +1,21 @@
...
+import { jsdom } from 'jsdom';
+
+global.document = jsdom('<!doctype html><html><body></body></html>', {
+ headers: {
+   'User-Agent':
+   'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_6_7)' +
+   ' AppleWebKit/534.24 (KHTML, like Gecko) Chrome/11.0.696.71 Safari/534.24'
+ }
+});
+global.window = document.defaultView;
```
Gnito: If I understand this right, this line would be more explicit as:
``` global.window = global.document.defaultView; ```

File: client/mochaConfig.js [hunk 5]
```diff
@@ -0,0 +1,21 @@
...
+global.action = action;
+global.linkTo = linkTo;
+global.describe = describe;
+global.it = it;
+global.expect = expect;
+global.specs = specs;
+
+import { jsdom } from 'jsdom';
+
+global.document = jsdom('<!doctype html><html><body></body></html>', {
```
Gnito: Adding comment like:
// Enzyme fix: load a document into the global scope before requiring React

client/mochaConfig.js
```diff
@@ -0,0 +1,21 @@
...
+ headers: {
```
Gnito: Indentation

client/mochaConfig.js
Gnito: We also might need to change this to a version where window's properties are exposed.
( http://airbnb.io/enzyme/docs/guides/jsdom.html )
```
Object.keys(document.defaultView).forEach((property) => {
  if (typeof global[property] === 'undefined') {
    exposedProperties.push(property);
    global[property] = document.defaultView[property];
  }
});
```

client/package.json
```diff
@@ -19,7 +19,8 @@
     "print-phantomjs-version": "phantomjs --version",
     "start-phantomjs": "phantomjs --webdriver=8910",
     "styleguide": "start-storybook -p 9001",
-    "deploy-storybook": "storybook-to-ghpages"
+    "deploy-storybook": "storybook-to-ghpages",
+    "test": "mocha --reporter mocha-circleci-reporter --require babel-register --require ignore-styles --require ./mochaConfig.js './app/components/composites/ListingCard/ListingCard.story.js'"
```
Gnito: Maybe we could move babel-register etc. to mochaConfig.js
See: https://github.com/lelandrichardson/enzyme-example-mocha/blob/master/test/.setup.js

client/app/components/composites/ListingCard/ListingCard.story.js
```diff
@@ -159,25 +160,38 @@ storiesOf('Search results')
...
+          distanceUnit: 'km',
+          color: '#347F9D',
+          className: css.listing,
+        },
+      ));
+
+    specs(() => describe('Failing image', function() {
+      const output = mount(story);
+      it('Should display "No picture"', function() {
+        expect(output.text()).to.include('No picture');
```
Gnito: This is async test and the version of jsdom Enzyme includes ("jsdom": "^6.1.0"), doesn't know how to deal with image loading. Even if we had a newer version, I'm not sure we want to set jsdom to fetch images (it most likely makes these test slow).

client/app/components/composites/Menu/Menu.story.js
```diff
@@ -1,5 +1,5 @@
 import r from 'r-dom';
-import { storiesOf } from '@kadira/storybook';
+const { storiesOf } = storybookFacade;
```
Gnito: This could be moved under the import section

client/webpack.storybook.config.js
Gnito: I added that

client/mochaConfig.js
Gnito: I added that

client/mochaConfig.js
Gnito: I added this one

client/app/components/composites/ListingCard/ListingCard.story.js
Gnito: This is changed to test internal state instead of actual image loading fail.

client/mochaConfig.js
Gnito: [x] added

client/app/components/sections/Topbar/Topbar.story.js
```diff
@@ -120,17 +121,60 @@ const loggedOut = (props) => ({
...
+    expect(component.text()).to.contain('Sign up');
+  });
+  it('shouldn\'t contain logout link', () => {
+    expect(component.text()).to.not.contain('Log out');
+  });
+};
+
+const hasLogo = (component) => {
+  it('should contain logo', () => {
+    expect(component.find('.Logo')).to.exist;
```
mporkola: `eslint` fails here, see https://github.com/eslint/eslint/issues/2102

The solution seems to be either disabling `no-unused-expressions` rule, or https://github.com/prodatakey/dirty-chai. Not sure which I prefer.

mporkola: 👍

PR #2555, opened 2016-09-19
Sitemap redirect fixes

rap1ds:
- Copy logic to define the redirect reason from MarketplaceLookup middleware
- Generate returns 404 if CDN is not in use
- Split `can_show_sitemap?` method, because part of that logic is already in the `redirect_reason` method

### DIFF
--- a/app/controllers/sitemap_controller.rb
+++ b/app/controllers/sitemap_controller.rb
@@ -16,13 +16,13 @@ class SitemapController < ActionController::Metal
   CACHE_HEADER = "X-Sitemap-Cache"
 
   def sitemap
-    return if render_sitemap_disabled
+    return if render_sitemap_disabled!
 
     com = community_from_request(request)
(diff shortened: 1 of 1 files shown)


app/controllers/sitemap_controller.rb
```diff
@@ -16,12 +16,14 @@ class SitemapController < ActionController::Metal
...
-    reason = redirect_reason(request)
+    return if render_sitemap_disabled
+
+    com = community_from_request(request)
+    reason = redirect_reason_from_request(request)
 
-    return if do_host_redirect!(com, reason)
     return if render_not_found!(reason)
-    return unless can_show_sitemap?(com)
+    return if render_private_community(com)
```
bladealslayer: `render_not_found!` has `!`, while the other `render_*` methods don't. I think it's better if all have the exclamation mark, as the behaviour is similar.

PR #2559, opened 2016-09-20
Validate booking slots

kpuputti:
This PR checks that the booking dates (between start date inclusive, end date exclusive) are available in the Harmony API before moving into payment.

## TODO

- [X] Date query param encoder to HarmonyClient
- [x] Query timeslots from Harmony
- [x] Validate all days/nights are available in Harmony response
- [x] Only validate timeslots with feature flag and availability turned on
- [x] Git cleanup

### DIFF
--- a/app/controllers/preauthorize_transactions_controller.rb
+++ b/app/controllers/preauthorize_transactions_controller.rb
@@ -90,13 +90,26 @@ class PreauthorizeTransactionsController < ApplicationController
 
     module_function
 
-    def validate_initiate_params(tx_params:,
+    def validate_initiate_params(marketplace_uuid:,
+                                 listing_uuid:,
+                                 tx_params:,
                                  quantity_selector:,
                                  shipping_enabled:,
-                                 pickup_enabled:)
+                                 pickup_enabled:,
+                                 availability_enabled:)
 
       validate_delivery_method(tx_params: tx_params, shipping_enabled: shipping_enabled, pickup_enabled: pickup_enabled)
         .and_then { validate_booking(tx_params: tx_params, quantity_selector: quantity_selector) }
+        .and_then { |result|
+          if FeatureFlagHelper.feature_enabled?(:availability) && availability_enabled
+            validate_booking_timeslots(tx_params: tx_params,
+                                       marketplace_uuid: marketplace_uuid,
+                                       listing_uuid: listing_uuid,
+                                       quantity_selector: quantity_selector)
+          else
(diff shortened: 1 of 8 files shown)


app/controllers/preauthorize_transactions_controller.rb
```diff
@@ -146,6 +159,52 @@ def validate_booking(tx_params:, quantity_selector:)
...
 
+    def all_days_available(timeslots, start_on, end_on)
+      # Take all the days except the exclusive last day
+      requested_days = (start_on..end_on).to_a[0..-2]
+
+      available_days = timeslots
+                         .select { |s| s[:unitType] == :day }
+                         .map { |s| s[:start].to_date }
+
+      requested_days.all? { |d|
```
rap1ds: This is perfectly fine, no need to change, but If you want to make this without iterating, you could turn these arrays to Sets and use set operations.

app/utils/service_client/middleware/param_encoder.rb
```diff
@@ -0,0 +1,27 @@
...
+    class ParamEncoder < MiddlewareBase
+      def enter(ctx)
+        params = ctx[:req][:params]
+        ctx[:req][:params] = encode(params)
+        ctx
+      end
+
+      private
+
+      def encode(params)
```
kpuputti: Thanks for a great suggestion! The `is_a?` felt awkward but I didn't know any better. Your suggestion sounds good, I'll change to that.

config/locales/en.yml
```diff
@@ -1599,6 +1599,8 @@ en:
       night: night
       nights: nights
       invalid_parameters: "Invalid values for new transaction"
+      dates_not_available: "Selected dates not available"
```
rap1ds: Is this missing _are_ ?

app/services/result.rb
```diff
@@ -96,6 +114,14 @@ def and_then(&block)
       self
     end
 
+    # Error a -> Result b
+    def rescue(&block)
+      result = block.call(self)
```
rap1ds: The Result object is now passed to the handler block is as, which is different what `and_then` does. `and_then` passes unwrapped `data` to the handler. So I think `rescue` should also pass unwrapped `data` and the `error_msg` the same way as `on_error` does: https://github.com/sharetribe/sharetribe/pull/2559/files#diff-3806624eafa2c3ce0e7273410ae8e337R131

Does it make sense?

PR #2562, opened 2016-09-21
Add sync/async logic to TransactionService layer

rap1ds:
TODO:

- [x] Save UUIDs as binary

### DIFF
--- a/app/controllers/transactions_controller.rb
+++ b/app/controllers/transactions_controller.rb
@@ -155,7 +155,23 @@ class TransactionsController < ApplicationController
     }
   end
 
-  def op_status
+  def transaction_op_status
+    process_token = params[:process_token]
+
+    resp = Maybe(process_token)
+      .map { |ptok| transaction_process_tokens.get_status(ptok) }
+      .select(&:success)
+      .data
+      .or_else(nil)
+
+    if resp
+      render :json => resp
+    else
+      redirect_to error_not_found_path
+    end
+  end
+
+  def paypal_op_status
     process_token = params[:process_token]
 
     resp = Maybe(process_token)
@@ -179,6 +195,10 @@ class TransactionsController < ApplicationController
 
   def paypal_process
     PaypalService::API::Api.process
+  end
+
+  def transaction_process_tokens
+    TransactionService::API::Api.process_tokens
   end
 
   private
--- a/app/controllers/transactions_controller.rb
+++ b/app/controllers/transactions_controller.rb
@@ -167,7 +167,7 @@ class TransactionsController < ApplicationController
     if resp
       render :json => resp
     else
-      redirect_to error_not_found_path
+      head :not_found
     end
   end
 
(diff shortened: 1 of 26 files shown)


app/controllers/transactions_controller.rb
```diff
@@ -155,7 +155,23 @@ def show
...
+    resp = Maybe(process_token)
+      .map { |ptok| transaction_process_tokens.get_status(ptok) }
+      .select(&:success)
+      .data
+      .or_else(nil)
+
+    if resp
+      render :json => resp
+    else
+      redirect_to error_not_found_path
```
ovan: I know this is copy&paste code but it really does not make any sense to redirect to a 404 since this is a JSON endpoint and nobody will even see the error page.

app/services/transaction_service/process/preauthorize.rb
```diff
@@ -8,12 +10,28 @@ class Preauthorize
...
+      end
+    end
+
+    def do_create(tx, gateway_fields)
+      gateway_adapter = TransactionService::Transaction.gateway_adapter(tx[:payment_gateway])
+
+      completion = gateway_adapter.create_payment(
+        tx: tx,
+        gateway_fields: gateway_fields,
+        prefer_async: false)
```
ovan: We now have an option prefer_async which is named so that it hints what's the preferred behavior but cannot fully control it. Does this still make sense? When we're already in a worker context it feels to me we should be able to just force the adapter to go sync so don't end up scheduling a job from a job. Also, we wouldn't have to deal with the async completion case at all then. prefer_async already works prescriptively so it's almost purely about naming and communicating the intent correctly (to force and not hint).

app/controllers/transactions_controller.rb
rap1ds: True, we should probably just send empty 404.

app/utils/uuid_utils.rb
```diff
@@ -0,0 +1,37 @@
+module UUIDUtils
+
+  module_function
+
+  def create_raw
+    to_raw(UUIDTools::UUID.timestamp_create)
```
ovan: raw? or rename raw to to_raw

spec/utils/uuid_utils.rb
```diff
@@ -0,0 +1,43 @@
...
+      mysql_conn.prepare("
+        CREATE TEMPORARY TABLE `uuid_utils_test` (
+          `id` int(11) NOT NULL AUTO_INCREMENT,
+          `uuid` binary(16) NOT NULL,
+          PRIMARY KEY (`id`)
+        )").execute
+    end
+
+    it "parses back and forth" do
+      sorted_uuids = (1..100).map { UUIDTools::UUID.timestamp_create }
```
ovan: The timestamps are created with a 100ns granularity so if this code would happen to run faster than that you can end up with 2 or more UUIDs with the exact same timestamp component. In this case the rearranging does not specify the sort order explicitly, i.e. it's not something that matters. One option is to ensure the UUID timestamps don't overlap by adding a 1ms sleep between generating two UUIDs.

rap1ds: `timestamp_create` take the timestamp as an argument, so I can use that.

PR #2573, opened 2016-09-26
Rearranged marketplace UUIDs

kpuputti:
Currently marketplaces have the `UUIDTools::UUID` instances of their UUID values accessible through the `.uuid` method and the raw binary value is stored to the DB as UUIDTools encodes it.

This PR changes the Communities so that the raw binary DB presentation is rearranged for efficient insertion, and the `UUIDTools::UUID` instance is accessible through the `.uuid_object` method.

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -122,7 +122,7 @@ FactoryGirl.define do
     automatic_confirmation_after_days 14
     listing_quantity 1
     listing_uuid { listing.uuid.raw }
-    community_uuid { community.uuid.raw }
+    community_uuid { community.uuid } # raw UUID
   end
 
   factory :conversation do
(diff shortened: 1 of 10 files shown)


spec/factories.rb
```diff
@@ -122,7 +122,7 @@ def build_association(association, opts = {})
     automatic_confirmation_after_days 14
     listing_quantity 1
     listing_uuid { listing.uuid.raw }
-    community_uuid { community.uuid.raw }
+    community_uuid { community.uuid } # raw UUID
```
kpuputti: This will now be the rearranged binary UUID, is that ok?

spec/services/transaction_service/paypal_events_spec.rb
```diff
@@ -64,7 +64,7 @@ def create_test_transaction(opts)
       payment_process: :preauthorize,
       payment_gateway: :paypal,
       community_id: @cid,
-      community_uuid: @community.uuid.raw,
+      community_uuid: @community.uuid, # raw UUID
```
kpuputti: Like above, this will now be the rearranged binary UUID, is that ok?

spec/factories.rb
rap1ds: Yeah I guess so

PR #2575, opened 2016-09-26
ListingCardPanel (container to show search results)

Gnito:
- Added ListingCardPanel
- Composite and elements test inside some section topic (to achieve "Summary" styleguide page).
e.g. SearchPage can include ListingCard stories as sub-stories.
- Stories grouped under Branding, Onboarding, Topbar, Search,
- Cleaned no-brainer eslint warnings

https://sharetribe.github.io/sharetribe/listing-card-panel/


client/app/components/composites/ListingCardPanel/ListingCardPanel.css
```diff
@@ -0,0 +1,49 @@
+.panel {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-around;
+  padding: calc(var(--ListingCardPanel_gutterSpacing) / 2);
+
+  &::after {
+    content: '';
+    flex: auto;
+  }
```
mporkola: what's this?

client/app/components/composites/ListingCardPanel/ListingCardPanel.css
```diff
@@ -0,0 +1,49 @@
...
+}
+
+.card {
+  margin: calc(var(--ListingCardPanel_gutterSpacing) / 2);
+  min-width: 264px;
+  max-width: 550px;
+  flex-basis: calc(100% - 24px);
+
+  /* 24 + 264 + 24 + 264 + 24 = 600 */
```
mporkola: It'd be worth mentioning at least once what the calculation is about

client/app/utils/numbers.js
```diff
@@ -18,4 +19,9 @@ const formatDistance = function formatDistance(distance, unit, precision = PRECI
 
 const formatPrice = (price, unit) => `${unit} ${price}`;
 
-export { formatDistance, formatPrice };
+const toFixedNumber = (number, limit, base) => {
```
mporkola: What does this function do? `toFixed` is standard JS, and this apparently does something else?

client/app/components/composites/ListingCardPanel/ListingCardPanel.js
```diff
@@ -0,0 +1,38 @@
...
+class ListingCardPanel extends Component {
+
+  shouldComponentUpdate(nextProps) {
+    return this.props.children !== nextProps.children;
+  }
+
+  render() {
+    const childrenWithColumnStyle = Children.map(
+      this.props.children,
+      (child) => React.cloneElement(child, {
```
mporkola: Does this mean we have two copies of the element in memory? Just wondering if there was a better way to pass elements as children.

client/app/components/composites/ListingCardPanel/ListingCardPanel.story.js
```diff
@@ -0,0 +1,63 @@
...
+import ListingCard from '../ListingCard/ListingCard';
+
+const LISTINGS_COUNT = 24;
+
+const { storiesOf, specs, expect } = storybookFacade;
+const containerStyle = { style: { background: 'white' } };
+
+
+const listingCardTemplate = (title, perUnit, price, distance) => (
+  r(ListingCard, Object.assign({},
```
mporkola: Why is `Object.assign` needed here?

client/app/components/composites/ListingCardPanel/ListingCardPanel.css
Gnito: If last row is not full, those listings get too much space and are not stacked to left side of the row.

client/app/components/composites/ListingCardPanel/ListingCardPanel.css
Gnito: Roger. (it's about columns & gutters)

client/app/utils/numbers.js
Gnito: ```toFixed``` returns string, but ListingCard expects number.

client/app/components/composites/ListingCardPanel/ListingCardPanel.story.js
Gnito: Actually it is not. (We have stories with default props merged with overwritten props - this was copy-paste blindness...) I'll remove.

client/app/components/composites/ListingCardPanel/ListingCardPanel.js
Gnito: It seems so. That's due immutable nature of current React:
https://facebook.github.io/react/blog/2015/03/03/react-v0.13-rc2.html

However, I don't think it matters much whether we pass in pure data (props) or React components, since those components are a lightweight versions of DOM elements. 

In this case passing in child components makes more sense than props of child components (since that would mean that the panel needs to create and understand what's inside of it).

PR #2584, opened 2016-09-28
Searchpage with ListingCards

Gnito:


client/app/components/composites/ListingCard/ListingCard.js
```diff
@@ -38,31 +40,33 @@ class ListingCard extends Component {
 
   clickHandler() {
     if (canUseDOM) {
-      window.location = this.props.listingURL;
+      window.location = this.props.listing.get('listingURL');
```
mporkola: with records, you can use native-like accessor `listing.listingURL`

app/views/search_page/search_page.erb
```diff
@@ -6,6 +6,11 @@
     locale: I18n.locale,
     defaultLocale: I18n.default_locale
   },
+  marketplace: {
+    marketplace_color1: CommonStylesHelper.marketplace_colors(@current_community)[:marketplace_color1],
```
mporkola: I know I'm behind the original naming, but `marketplace.marketplace_color1` feels a bit redundant

Gnito: Good point - let's do that when we decide how to go forward with Topbar (will it be included to search page). Then we have a good time to do the naming change also helper files and topbar's props naming.

PR #2586, opened 2016-09-28
Listing UUID API

kpuputti:
Builds on top of #2585 

UUID handling for different models will be changed to follow the convention of .uuid accessing the raw binary UUID and .uuid_object accessing the UUIDTools::UUID instance.

This PR changes the Listing model to use the rearranged raw binary UUIDs as the storage format, but offers a convenience API on top of those.

### DIFF
--- a/db/migrate/20160928080819_copy_rearranged_uuids_to_transactions.rb
+++ b/db/migrate/20160928080819_copy_rearranged_uuids_to_transactions.rb
@@ -1,0 +1,6 @@
+class CopyRearrangedUuidsToTransactions < ActiveRecord::Migration
+  def up
+    execute "UPDATE transactions, listings SET transactions.listing_uuid = listings.uuid WHERE transactions.listing_id = listings.id"
+    execute "UPDATE transactions, communities SET transactions.community_uuid = communities.uuid WHERE transactions.community_id = communities.id"
+  end
+end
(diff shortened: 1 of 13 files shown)


db/migrate/20160928080819_copy_rearranged_uuids_to_transactions.rb
```diff
@@ -0,0 +1,10 @@
+class CopyRearrangedUuidsToTransactions < ActiveRecord::Migration
+  def up
+    execute "UPDATE transactions, listings SET transactions.listing_uuid = listings.uuid WHERE transactions.listing_id = listings.id"
+    execute "UPDATE transactions, communities SET transactions.community_uuid = communities.uuid WHERE transactions.community_id = communities.id"
+  end
+
+  def down
+    execute "UPDATE transactions SET listing_uuid = NULL, community_uuid = NULL"
+  end
```
rap1ds: NULL is not allowed for `listing_uuid` and `community_uuid`.

I'd suggest you just remove the `down` migration. The original state where we were before running the migrations is something we can't go back to, because we overrode the original values.

kpuputti: Good point, thanks.

kpuputti: Fixed.

PR #2587, opened 2016-09-28
Add new Result object

rap1ds:
This PR uses the Result object gem instead. It also contains an "adapter" (i.e. initializer with bunch of monkey patches) from old to new Result object. When ever we use the old methods, a deprecation warning is shown.


rap1ds:
No need to keep this PR open.

PR #2594, opened 2016-09-29
Remove migrations that could not be run.

kpuputti:
Changing the UUID wasn't efficient enough and the migrations could not
be run.

### DIFF
--- a/db/migrate/20160928080048_rearrange_listing_uuid_bytes.rb
+++ b/db/migrate/20160928080048_rearrange_listing_uuid_bytes.rb
@@ -1,14 +1,7 @@
 class RearrangeListingUuidBytes < ActiveRecord::Migration
   def up
-    mysql_conn = ActiveRecord::Base.connection.raw_connection
-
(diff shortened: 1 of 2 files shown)


db/migrate/20160928080048_rearrange_listing_uuid_bytes.rb
```diff
@@ -1,14 +1,7 @@
 class RearrangeListingUuidBytes < ActiveRecord::Migration
   def up
```
sktoiva: Hm. Does this work? Should it have `down` also? Or is it inherited from ActiveRecord::Migration as no-op?

Other option could be `change`.

kpuputti: It should work, we've had migration before that only have the up method.

kpuputti: Example: https://github.com/sharetribe/sharetribe/blob/master/db/migrate/20160926111847_rearrange_community_uuid_bytes.rb

PR #2597, opened 2016-09-30
Parse user profiles and attach them to listings

mporkola:
Parse user profile from Transit data to Profile record, and attach that to Listing record.

### DIFF
--- a/client/app/startup/SearchPageApp.js
+++ b/client/app/startup/SearchPageApp.js
@@ -28,11 +28,6 @@ export default (props) => {
 
   const bootstrappedData = TransitImmutableConverter.fromJSON(props.data);
 
-  const listingWithAuthor = (listing, profiles) => {
-    const author = profiles.get(listing.authorId);
-    return listing.set('author', author);
-  };
-
   const profilesToMap = (includes) =>
     includes.reduce((acc, val) => {
       const type = val.get(':type');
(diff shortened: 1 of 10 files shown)


client/app/startup/SearchPageApp.js
```diff
@@ -27,18 +28,36 @@ export default (props) => {
...
+    includes.reduce((acc, val) => {
+      const type = val.get(':type');
+      if (type === ':profile') {
+        const profile = parseProfile(val);
+        const id = val.get(':id');
+        return acc.set(id, profile);
+      } else {
+        return acc;
+      }
+    }, new Immutable.Map());
```
mporkola: I'm starting to think we need to plan some kind of structure for these conversions.

client/app/startup/SearchPageApp.js
```diff
@@ -27,18 +28,36 @@ export default (props) => {
 
   const bootstrappedData = TransitImmutableConverter.fromJSON(props.data);
 
-  const listingsToMap = (listings) =>
+  const listingWithAuthor = (listing, profiles) => {
+    const author = profiles.get(listing.authorId);
+    return listing.set('author', author);
```
mporkola: Should we just store the Listing to Redux store normalised, without the Author attached, and then attach the Author only in `mapStateToProps`? Let's plan.

client/app/utils/transitImmutableConverter.js
```diff
@@ -31,6 +31,7 @@ const createReader = function createReader() {
     handlers: {
       ':': (rep) => `:${rep}`,
       list: (rep) => Immutable.List(rep).asImmutable(),
+      u: (rep) => rep.toString(),
```
mporkola: This was a nasty surprise: Immutable maps didn't do well with UUID objects as keys, so I converted them to string representation.

client/app/utils/transitImmutableConverter.js
```diff
@@ -16,6 +16,10 @@ const toImage = (data) => {
   return styles;
 };
 
+// Outside of this file we should only pass UUID references, no need to export
+const UUID = Immutable.Record({ stringRep: '' });
+const toUUID = (transitUuid) => new UUID({ stringRep: transitUuid.toString() });
```
Gnito: If Record wrapper is what we want, then also ListingModel's id should be updated to use it (+ [keys](https://github.com/sharetribe/sharetribe/blob/master/client/app/components/sections/SearchPage/SearchPage.js#L28) and related stories/tests)

mporkola: That way we don't lose the type information: it's not a string, it's an UUID. It pretty much only matters when we're serialising the data back to Transit, but parsing it separately everywhere would be a hassle. `ListingModel`'s id is parsed by this same function, so no need to do it separately.

I'll change the field name though, it's not good

PR #2598, opened 2016-09-30
Fakepal PStore

rap1ds:
This PR adds PStore to the fake PayPal implementation (FakePal). With the help of PStore, the Fakepal that is running in the app process and the Fakepal that is running in the worker process and share information. Due to this, we can make Fakepal to support asynchronous operations.

Now that the Fakepal supports asynchronous operations, we can remove the synchronous option for PayPal operations. This is great, since those operations weren't used in production, but only for testing. This change brings the FakePal and the real PayPal implementation closer to each other.


features/support/feature_tests/worker.rb
```diff
@@ -0,0 +1,20 @@
+module FeatureTests
+  module Worker
+    module_function
+
+    def work_until(&block)
+      worker = Delayed::Worker.new(queit: true)
```
ovan: quiet?

features/support/feature_tests/worker.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      default_max_wait_time = Capybara.default_max_wait_time
+      Capybara.default_max_wait_time = 0.2
+
+      begin
+        Timeout.timeout(default_max_wait_time) do
+          worker.work_off until block.call
```
ovan: Will this work synchronously in the same thread as where the test is running?

app/services/paypal_service/api/payments.rb
```diff
@@ -37,7 +36,7 @@ def request(community_id, create_payment, force_sync: true)
       @lookup.with_active_account(
         community_id, create_payment[:merchant_id]
       ) do |m_acc|
-        if (use_async?(force_sync))
+        if !force_sync
```
ovan: Is the idea to remove the async path for these from paypal service altogether now that the transaction service level handles these operations asynchronously?

ovan: I'm not entirely on board anymore how we use this flag, is it always passed as true (or left out and default is true)?

features/support/feature_tests/worker.rb
rap1ds: This is running synchronously in the test runner thread.

app/services/paypal_service/api/payments.rb
rap1ds: I removed the parts that I could easily and safely remove.

rap1ds: So what this changes is that is removes the ability to run PayPal payments "synchronously", which in this case means, run payments without Delayed Jobs. Synchronized execution was previously then only way to use FakePal, because FakePal didn't have a way to share memory between app process and the worker process. Now that the implementation uses the file-based storage, FakePal can share state with two processes and thus FakePal can be run in asynchronous manner, the same way as the real implementation.

I hate the work sync/async in this context! They are ambiguous: In the context of the transaction layer, it means synchronous as in Braintree (you just enter the CC details in the payment view and that's it) or asynchronous as in PayPal (you're redirected to 3rd-party site, payment will be finalized when the user comes back). In PayPal, the word sync/async means the way how the code is executed: synchronously in app-process or asynchronously in worker process.

PR #2601, opened 2016-09-30
This PR standardizes most icons SVG's.

Luis-RG:
Several of the initial icon SVG codes were written with different specifications than we do now. Some of them were also added with a "wrong" name. This PR standardizes the icon's SVG codes. the next step is deleting the icons that have the wrong name and are not used anymore in CLP's


Gnito:
I assume that this can wait for Luis to return. Looks good, but I'd like to know a bit more about the process. There are still some icons remaining whose viewboxes are quite 'off' (like viewBox="1079 4804 61 53") - I guess those are picked from some sprite file.

Luis-RG:
@Gnito The boxes that are quite "off" are from initial icons. They will be deleted in the future. I just want to make sure that nothing breaks (no icons disappear from CLP's). I could delete this icons now. But prefer to delete them after this PR is merged and deployed.

PR #2602, opened 2016-09-30
Retry middleware

rap1ds:
This Pull Request adds Retry middleware

### DIFF
--- a/app/utils/context_runner.rb
+++ b/app/utils/context_runner.rb
@@ -77,48 +77,15 @@ class ContextRunner
 
   private
 
-  def try_execute_mw(ctx, mw_name, stage, &block)
+  def try_execute_mw(ctx, mw, stage, &block)
     block.call
   rescue StandardError => e
     new_ctx = ctx.dup
     new_ctx[:error] = e
-    new_ctx[:error_middleware] = mw_name
+    new_ctx[:error_middleware] = mw.class.name
     new_ctx[:error_stage] = stage
     new_ctx
   end
-
-
-  def execute_enter(ctx, mw)
-    try_execute_mw(ctx, mw.class.name, :enter) do
-      if mw.respond_to?(:enter)
-        mw.enter(ctx)
-      else
-        ctx
-      end
-    end
-  end
-
-  def execute_leave(ctx, mw)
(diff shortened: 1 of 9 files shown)


app/utils/context_runner.rb
```diff
@@ -112,7 +112,8 @@ def execute_leave(ctx, mw)
   end
 
   def execute_error(ctx, mw)
-    try_execute_mw(ctx, mw.class.name, :exit) do
+    try_execute_mw(ctx, mw.class.name, :error) do
```
ovan: Good catch. How has this stayed hidden so far?

app/utils/context_runner.rb
```diff
@@ -131,6 +127,8 @@ def next_mw(ctx)
       [new_ctx, mw, :error]
     elsif !new_ctx[:enter_queue].empty?
       mw = new_ctx[:enter_queue].pop
+      new_ctx[:leave_stack].unshift(mw)
```
ovan: This makes a lot of sense. I also confirmed from this is how Pedestal [handles it](https://github.com/pedestal/pedestal/blob/master/interceptor/src/io/pedestal/interceptor/chain.clj#L151).

app/utils/context_runner.rb
```diff
@@ -132,6 +135,8 @@ def next_mw(ctx)
       [new_ctx, mw, :enter]
     else
       mw = new_ctx[:leave_stack].shift
+      new_ctx[:complete_stack].unshift(mw)
```
ovan: Is there any way we could implement the desired behaviour without needing to add a new stack? My objection is that this:
* Adds complexity by adding one more moving part to be handled
* Is not obvious; the name complete_stack is not nearly as intuitive as enter_queue and leave_stack are which correspond to execution phases and order
* This is the only case where it has been needed so far (and I'm thinking if it's really needed here?) so it feels like adding this is only to support one specific use case and not a generally useful concept
* Pedestal doesn't have this

How about an approach where the retry middleware makes a copy of the remaining enter_queue at enter phase and stores it to ctx. And then in leave phase copies back the enter_queue to trigger a retry. Would this work?

ovan:
Beautiful! :)

PR #2604, opened 2016-10-03
Refactor ordering community categories

otlaitil:
- Remove invalid SQL injection guard (faulty regex)
- Simplify controller action
- Use AR to generate the mass update query, for enhanced security (generated SQL itself stays the same)
- Add controller tests for the order action

### DIFF
--- a/app/controllers/admin/categories_controller.rb
+++ b/app/controllers/admin/categories_controller.rb
@@ -57,23 +57,8 @@ class Admin::CategoriesController < ApplicationController
   end
 
   def order
-    sort_priorities = params[:order]
-                      .reject { |o| !o.match /[0-9]+/} #Guard against sql injection
-                      .each_with_index
-                      .map do |category_id, index|
-      [category_id, index]
-    end.inject({}) do |hash, ids|
-      category_id, sort_priority = ids
-      hash.merge(category_id.to_i => sort_priority)
-    end
-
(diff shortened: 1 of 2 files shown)


app/controllers/admin/categories_controller.rb
```diff
@@ -57,23 +57,8 @@ def update
   end
 
   def order
-    sort_priorities = params[:order]
-                      .reject { |o| !o.match /[0-9]+/} #Guard against sql injection
```
otlaitil: `.reject { |o| !o.match /[0-9]+/} #Guard against sql injection`

This regex allowed potential SQL injections to pass the `reject`. Correct regex in Ruby would be `/\A[0-9]+\z/` to match the string to contain only numbers.

The category id's are casted to integers with `to_i` later though, in the original code, so no injection _should_ actually be possible. 

I refactored this mainly for clarity and readability.

otlaitil:
@rap1ds what is the status with this?

rap1ds:
@otlaitil Thanks for the ping, and sorry for slow reaction! I'll check this today/tomorrow.

rap1ds:
Thanks @otlaitil ! The code looks good. Tested this on my local dev environment and it seems to work well!

otlaitil:
Great! Thanks @rap1ds 👍

Issue #2603, opened 2016-10-02
S3 images not loading with 5.12


thecrabnebula:
##### Sharetribe Version:
5.12

##### Environment:
Heroku

##### Server mode:
development

##### Description:
Since updating to 5.12, images will not load from us-west-2. This has been added to the s3_region variable specified in the documentation.

No other changes have been made to the server.

pcm211:
change the bucket location to US East (N Virginia) or use Cloudfront .... to make it work ...
PS See naming conventions of AWS ... 

Hope that solves your issue

rap1ds:
What was the previous working version you were running before the update?

thecrabnebula:
The previous version was 5.11.

thecrabnebula:
@pcm211: The naming convention used for the variable is the same as that in the config.default.yml. 

In the first instance, us-east-1 was tried, without success. 

The solution should not be "use Cloudfront" for something that has worked the way it is for a long time.

rap1ds:
@thecrabnebula Thanks for reporting the previous version number.

"Use CloudFront" is not the solution here. We don't do changes that requires you to change your hosting infra between version updates without specifically mentioning it in the upgrade notes.

rap1ds:
@bladealslayer Any ideas?

pcm211:
@thecrabnebula .... I have faced this problem ... code is not the problem it works the same ...the problem is the naming convention CALL ...US East has the config which calls as ... XXX.amazonaws.com ... rest calls are a bit longer ...  moreover do check that your bucket name does not have a dot "." in it ... because it changes the complete naming gambit 

@rap1ds  ... I feel we should add another param in the config file .. for the region in which the bucket is made and ammend the aws calling accordingly to seamlessly integrate all region buckets easily

bladealslayer:
@thecrabnebula do you have separate `s3_upload_bucket_name`? If yes, is that bucket also in `us-west-2`?

It seems we have possible issue with S3 uploads when there is separate uploads bucket in region other than `us-east-1`. However, I didn't see any issue with displaying images from `us-west-2` when testing. @thecrabnebula are previously uploaded images working and just uploads failing or all images don't display? We generate image URLs in the form `bucket-name.s3.amazonaws.com`, which normally gets redirected to the correct region endpoint.

Ideally, we should add `:s3_host_name` to paperclip options, in order to avoid the redirect and should support other regions in the uploader.

thecrabnebula:
[...]

PR #2605, opened 2016-10-03
Improve S3 region support

bladealslayer:
- [x] avoid unnecessary redirects when S3 region is not `us-east-1`

Some discussion already in #2603.


config/application.rb
```diff
@@ -158,12 +158,19 @@ class Application < Rails::Application
 
     if (APP_CONFIG.s3_bucket_name && APP_CONFIG.aws_access_key_id && APP_CONFIG.aws_secret_access_key)
       # S3 is in use for uploaded images
+      s3_domain = "amazonaws.com"
+      s3_host_name = if APP_CONFIG.s3_region == "us-east-1"
```
sktoiva: Should here be a comment for the special case of "us-east-1"?

PR #2606, opened 2016-10-03
Create bookable also when editing listing

rap1ds:
Create booking when editing listing, if the listing shape supports booking and the old listing availability is something else than booking.

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -398,6 +398,18 @@ class ListingsController < ApplicationController
 
     shape = get_shape(params[:listing][:listing_shape_id])
 
+    if FeatureFlagHelper.feature_enabled?(:availability) &&
+       shape.present? &&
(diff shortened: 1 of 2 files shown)


app/controllers/listings_controller.rb
```diff
@@ -398,6 +398,18 @@ def update
...
     shape = get_shape(params[:listing][:listing_shape_id])
 
+    if FeatureFlagHelper.feature_enabled?(:availability) &&
+       shape.present? &&
+       shape[:availability] == :booking &&
+       @listing.availability.to_sym != :booking
+
+      bookable_res = create_bookable(@current_community.uuid_object, @listing.uuid_object, @current_user.uuid_object)
+      unless bookable_res.success
+        flash[:error] = t("listings.error.something_went_wrong_plain")
```
ovan: Would there be a reason to use a more specific error message in this case or is it all the same to users if our BE services have connectivity issues?

PR #2607, opened 2016-10-03
Add info about assets for tests in README

nicolaracco:
It seems to me there's a missing step in the README, needed to successfully run tests (otherwise `spec/requests/https_redirects_spec.rb` fails).

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,8 @@ To update your local database schema to the newest version, run database migrati
 
 Tests are handled by [RSpec](http://rspec.info/) for unit tests and [Cucumber](https://cucumber.io/) for acceptance tests.
 
+Remember to follow *all* the steps listed in the [Setting up the development environment](#setting-up-the-development-environment) paragraph before running tests because some tests depend on webpack assets.
+
 1. Navigate to the root directory of the sharetribe project
 1. Initialize your test database:
 


rap1ds:
Interesting. The tests should depend on assets. What's the error you are seeing?

nicolaracco:
So these tests were failing:

```bash
rspec ./spec/requests/https_redirects_spec.rb[1:1:1:4] # Redirect to HTTPS for custom domain when always_use_ssl is true: does not happen over HTTPS
rspec ./spec/requests/https_redirects_spec.rb[1:1:1:5] # Redirect to HTTPS for custom domain when always_use_ssl is true: does not happen when SSL is terminated on proxy
rspec ./spec/requests/https_redirects_spec.rb[1:1:2:1] # Redirect to HTTPS for custom domain when always_use_ssl is false does not happen over HTTP
rspec ./spec/requests/https_redirects_spec.rb[1:1:2:2] # Redirect to HTTPS for custom domain when always_use_ssl is false does not happen over HTTPS
rspec ./spec/requests/https_redirects_spec.rb[1:2:1:4] # Redirect to HTTPS for marketplace subdomain when always_use_ssl is true: does not happen over HTTPS
rspec ./spec/requests/https_redirects_spec.rb[1:2:1:5] # Redirect to HTTPS for marketplace subdomain when always_use_ssl is true: does not happen when SSL is terminated on proxy
rspec ./spec/requests/https_redirects_spec.rb[1:2:2:1] # Redirect to HTTPS for marketplace subdomain when always_use_ssl is false does not happen over HTTP
rspec ./spec/requests/https_redirects_spec.rb[1:2:2:2] # Redirect to HTTPS for marketplace subdomain when always_use_ssl is false does not happen over HTTPS
```

On each one of these the test was expecting a 200 while it received a 500. When I checked the response body, it was reporting a `Sprockets::FileNotFound` exception because it couldn't find file `vendor-bundle`.

Please notice this was a fresh repo so I hadn't started the application in local.

rap1ds:
Ok, got it! Thanks for reporting! I can reproduce this in my local environment with a fresh install if I don't complete all the steps in the development environment setup. Which makes me think, instead of adding this additional step, would it make sense to add something like "Before starting these steps, perform all the steps in the development environment setup"? We want to encourage people to follow those steps. We also have similar note in the production setup.

Does that make sense?

nicolaracco:
Indeed I followed the development setup only till point 11 because point 12 was about "starting the development server", so I originally though it wasn't useful to run tests.

nicolaracco:
I added a brief statement for that

rap1ds:
@nicolaracco Thanks!

PR #2611, opened 2016-10-04
Move welcome_email translation keys to their own translation files

thomasmalbaux:
Welcome email (sent after trial creation) copy texts will be updated a bit. This PR prepares things so that those copy texts are no longer translated by contributors.

This removes welcome_email translation keys from main translation file (en.yml) and moves them to dedicated file for some translations.


config/locales/welcome_email/de.yml
```diff
@@ -0,0 +1,25 @@
+en:
```
rap1ds: The root needs to match to the language code, so here it should be `de:` in es-ES it should be `es-ES:` etc.

PR #2612, opened 2016-10-04
Avoid DB connections during assets:precompile

nicolaracco:
I'm unable to compile the assets when building a docker image because the `assets:precompile` task attempts to establish a connection to the database. The problem is related to the [initializer of money-rails 1.3.0](https://github.com/RubyMoney/money-rails/blob/v1.3.0/lib/money-rails/hooks.rb#L18) that invoking `ActiveRecord::Base.connection` triggers a new connection to the database.

Since the 1.4.0, this bug has been fixed in money-rails ([see the new initializer](https://github.com/RubyMoney/money-rails/blob/v1.4.1/lib/money-rails/hooks.rb#L20)). It seems there are no regressions in the test suite, and the `assets:precompile` task is now running fine.

Given I'm still not confident with the codebase I didn't dare to update money-rails to the latest version (1.7.0).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,7 +60,7 @@ gem "omniauth-facebook", "~> 3.0.0"
 # Dynamic form adds helpers that are needed, e.g. error_messages
 gem 'dynamic_form', "~>1.1.4"
 gem "truncate_html", "~>0.9.1"
-gem 'money-rails', "~>1.3.0"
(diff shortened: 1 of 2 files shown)


rap1ds:
@nicolaracco Thanks! I tried this locally and indeed, it fixes the issue with the MySQL connection when compiling assets. Thanks!

PR #2613, opened 2016-10-04
Distance & Price added to ListingCard

Gnito:
This PR adds price and distance to ListingCard. 

screen shot 2016-10-06 at 20 51 31

- currency_iso.json is from money gem (since that's what we use in backend and it's best not to change conversions between base units and fractional units)
- locale_info is needed to translate simple language code ('en') to full language code ('en-US')
- Numbro is used to formatting numerals (both distance & money)
- Those locales that are missing from Numbro's language rebository are using default ('en-US')

[Styleguide](https://sharetribe.github.io/sharetribe/transit-distance-price-data/?selectedKind=Search%20results&selectedStory=Summary&full=0&down=1&left=1&panelRight=0&downPanel=storybook-addon-specifications%2Fspecifications-panel)


client/app/assets/json/currency_iso.json
```diff
@@ -0,0 +1,924 @@
+{
```
Gnito: This file is money gem's configuration.
I've only added some of the data points: name, subunit_to_unit, symbol, [disambiguate_symbol]

client/app/utils/numbers.js
```diff
@@ -1,27 +1,119 @@
...
+  const currencyData = currencies[price.currency.toLowerCase()];
+  if (currencyData == null) {
+    throw new Error('Unknown currency');
+  }
+
+  const localizedNumbro = localizeNumbro(locale);
+  const amount = price.fractionalAmount / currencyData.subunit_to_unit;
+  const cultureSpecificFormat = localizedNumbro.cultureData().formats.fullWithTwoDecimals;
+
+  return localizedNumbro(amount, price.currency).formatForeignCurrency(currencyData.symbol, cultureSpecificFormat);
```
Gnito: Instead of using ```currencyData.symbol```, we could also create variable *symbol*:
```javascript
const symbol = currencyData.disambiguate_symbol ? currencyData.disambiguate_symbol : currencyData.symbol;
```
That way users would always see *US$*, *CA$* instead of bit confusing *$*.

client/app/utils/numbers.js
```diff
@@ -1,23 +1,121 @@
...
+const localizeNumbro = function localizeNumbro(locale) {
+
+  // Ensure that numbro has required all the localization languages and select correct locale
+  if (KNOWN_LOCALES.includes(locale)) {
+    const isServer = typeof window === 'undefined';
+    const numbroMultiLang = initializeNumbro(numbro, isServer ? null : locale);
+    numbroMultiLang.culture(locale);
+    return numbroMultiLang;
+  }
+  numbro.culture(DEFAULT_LOCALE);
```
mporkola: The control flow would be more obvious with an `else` block

client/app/utils/numbers.js
```diff
@@ -1,23 +1,121 @@
...
+    KNOWN_LOCALES.forEach((localeCode) => {
+      if (localeCode !== 'en-US') {
+        n.culture(localeCode, require(`numbro/languages/${localeCode}`));
+      }
+    });
+  }
+  return n;
+}, (n, locale) => (locale ? locale : 'all'));
+
+const localizeNumbro = function localizeNumbro(locale) {
```
mporkola: Does this return a new instance of numbro? If so, `localizedNumbro` would convey the meaning better, `localizeNumbro` sounds like state mutation

client/app/utils/numbers.js
```diff
@@ -1,23 +1,121 @@
...
   }
 
-  const precisionWithCloseBy = (distance < 1) ? 1 : precision;
-  return (distance < minimumDistance) ?
-    `< 0.1 ${unit}` :
-    `${sigFigs(distance, precisionWithCloseBy)} ${unit}`;
+  const localizedNumbro = localizeNumbro(locale);
+  const dist = localizedNumbro(sigFigs(distance.value, precision));
+  const localizedUnit = translateDistanceUnit(distance.unit);
+  const formatted = (distance.value < 1) ? dist.format('0,0.0') : dist.format('0,0');
```
mporkola: If I read the docs on this correctly, `'0,0.0'`/`'0,0'` specifies comma as thousands separator, and dot as decimal separator, but shouldn't this be specified by locale?

client/app/utils/numbers.js
```diff
@@ -1,23 +1,121 @@
...
+
+  return (distance.value < minimumDistance) ?
+    `< 0.1 ${localizedUnit}` :
```
mporkola: Same here, decimal delimiter by locale

client/app/components/composites/ListingCard/ListingCard.story.js
```diff
@@ -107,7 +119,7 @@ const testPrice = function priceTest(card, mountedCard) {
 };
 const testDistance = function priceTest(card, mountedCard) {
   it('Should display formatted distance', () => {
-    expect(mountedCard.text()).to.include(formatDistance(card.props.listing.distance, card.props.listing.distanceUnit));
+    expect(mountedCard.text()).to.include(formatDistance(card.props.listing.distance));
```
mporkola: Maybe also a test for currency formatting? E.g. locale specific separators and positioning, weird amount of fractionals

client/app/components/composites/ListingCard/ListingCard.js
```diff
@@ -53,7 +53,12 @@ class ListingCard extends Component {
 
     const localeCode = fullLocaleCode();
     const distanceFormatted = formatDistance(listing.distance, localeCode);
-    const priceFormatted = listing.price ? formatMoney(listing.price.get(':money'), localeCode): null;
+    const price = listing.price;
+    const moneyFormatted = price ? formatMoney(price.get(':money'), localeCode) : null;
+    const pricingUnitTranslations = price && price.get(':pricingUnit').size > 0 ?
+      price.get(':pricingUnit') :
+      null;
+    const pricingUnit = `/ ${localizedString(pricingUnitTranslations, 'pricing unit')}`;
```
mporkola: Maybe this could happen in a different `PriceUtils`? We're going to need it in many places at least later

client/app/components/composites/ListingCard/ListingCard.story.js
```diff
@@ -114,7 +120,7 @@ const ListingCardImageError =
 
 const testPrice = function priceTest(card, mountedCard) {
   it('Should display formatted price', () => {
-    expect(mountedCard.text()).to.include(formatMoney(card.props.listing.price, card.props.listing.priceUnit));
+    expect(mountedCard.text()).to.include(formatMoney(card.props.listing.price.get(':money'), card.props.listing.price.get(':priceUnit')));
```
mporkola: I think these tests would be better if we'd match against a literal formatted value, in this case `'199,00 €'` (?)

client/app/utils/numbers.js
Gnito: numbro is unfortunately a global object - so it's not returning new instance.

client/app/utils/numbers.js
```diff
@@ -1,23 +1,121 @@
...
+  'et-EE',
+  'pl-PL',
+  'hu-HU',
+  'cs-CZ',
+  'th-TH',
+  'zh-TW',
+  'sk-SK',
+];
+
+const initializeNumbro = _.memoize((n, locale) => {
```
Gnito: Ah, true. can't use numbro as it clashes with our eslint rules (shadowing). 
I wanted to make this function a bit more functional style than what actually happens here (i.e. passing numbro as a parameter even though it is global object).

client/app/utils/numbers.js
Gnito: Well, it's harder to make mistakes this way while refactoring code (that's why eslint prefers this style)

client/app/utils/numbers.js
Gnito: Nope, those are just placeholders for locale specific rules. So, that *'0,0.0'* should be read as "use thousand separator and decimal separator [according to installed culture rules]"

client/app/utils/numbers.js
Gnito: This is a good catch! 
This should be ```\`< ${formattedMinimumDistance} ${localizedUnit}\````

client/app/components/composites/ListingCard/ListingCard.story.js
Gnito: They are harder to write and maintain, if we spread literals everywhere.
But good point, we should have separate tests for utility functions.

client/app/components/composites/ListingCard/ListingCard.story.js
Gnito: 👍 
(Also type in the declaration: priceTest -> testDistance )

client/app/components/composites/ListingCard/ListingCard.js
Gnito: Pricing unit "/ day" is usually showed with different styles and "/ unit" or "per unit" texts. I'd not generalize this (at least yet).

mporkola: I meant the three lines starting from `const moneyFormatted`

client/app/utils/numbers.js
mporkola: confusing notation!

client/app/utils/numbers.js
mporkola: I wouldn't like us to do that kind of decision for the user: you rarely see "US$" anywhere. If plain "$" is considered confusing somewhere, in my opinion it should be an admin switch. But this is a design issue, let's not solve it in code review.

mporkola:
👍 besides eslint failures

client/app/components/composites/ListingCard/ListingCard.js
```diff
@@ -55,10 +55,9 @@ class ListingCard extends Component {
     const distanceFormatted = formatDistance(listing.distance, localeCode);
     const price = listing.price;
     const moneyFormatted = price ? formatMoney(price.get(':money'), localeCode) : null;
-    const pricingUnitTranslations = price && price.get(':pricingUnit').size > 0 ?
-      price.get(':pricingUnit') :
-      null;
-    const pricingUnit = `/ ${localizedString(pricingUnitTranslations, 'pricing unit')}`;
+    const hasPricingUnit = price && price.get(':pricingUnit') != null;
```
Gnito: The format of unknown pricing unit is not yet defined in DiscoveryAPI, therefore we check the existence of key (':pricingUnit') and possible null value.

PR #2614, opened 2016-10-04
Save booking_uuid after booking is initialized

rap1ds:
Save booking_uuid after the booking is initialized.


rap1ds:
This is already merged, see #2620

PR #2615, opened 2016-10-04
Accept booking

rap1ds:
Accept booking after the preauthorize payment is completed.

Built on top of #2614 

**Open questions**

- For simplicity, the transaction is now moved to `paid` state even if the booking accept fails. Is this the right thing to do? If not, what's the state where the transaction is left if the preauthorized payment is completed, but the booking is still in `initiated` state?

### DIFF
--- a/app/services/transaction_service/process/preauthorize.rb
+++ b/app/services/transaction_service/process/preauthorize.rb
@@ -152,7 +152,35 @@ module TransactionService::Process
     end
 
     def finalize_complete_preauthorization(tx:, gateway_adapter:)
-      Transition.transition_to(tx[:id], :paid)
+      ensure_can_execute!(tx: tx, allowed_states: [:preauthorized, :pending_ext, :paid])
+
+      if tx[:current_state] == :paid
+        Result::Success.new()
+      else
+        Transition.transition_to(tx[:id], :paid)
+
+        if tx[:availability] != :booking
+          Result::Success.new()
+        else
+          HarmonyClient.post(
+            :accept_booking,
+            params: {
+              id: tx[:booking_uuid]
+            },
+            body: {
+              actorId: UUIDUtils.base64_to_uuid(tx[:listing_author_id]),
+              reason: "provicer accepted"
+            },
+            opts: {
+              max_attempts: 3
+            }).on_error { |error_msg, data|
+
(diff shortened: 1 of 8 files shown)


app/services/transaction_service/process/preauthorize.rb
```diff
@@ -152,7 +152,35 @@ def complete_preauthorization(tx:, message:, sender_id:, gateway_adapter:)
     end
 
     def finalize_complete_preauthorization(tx:, gateway_adapter:)
-      Transition.transition_to(tx[:id], :paid)
+      ensure_can_execute!(tx: tx, allowed_states: [:preauthorized, :pending_ext, :paid])
+
+      if tx[:current_state] == :paid
+        Result::Success.new()
+      else
+        Transition.transition_to(tx[:id], :paid)
```
ovan: Why does this :paid state sound like a new thing to me. Have we always had it?

app/services/transaction_service/process/preauthorize.rb
```diff
@@ -152,7 +152,35 @@ def complete_preauthorization(tx:, message:, sender_id:, gateway_adapter:)
...
+            opts: {
+              max_attempts: 3
+            }).on_error { |error_msg, data|
+
+            logger.error("Failed to accept booking",
+                         :failed_accept_booking,
+                         tx.slice(:community_id, :id).merge(error_msg: error_msg))
+          }
+        end
+      end
```
ovan: Do I get this right that we only log this error and don't show anything to user but still mark the transaction to have moved forward? Can't the provide retry accept if it fails?

app/services/transaction_service/process/preauthorize.rb
rap1ds: Yes we have

app/services/transaction_service/process/preauthorize.rb
rap1ds: Well that exactly is the big open question...

What would happen if we end up accepting the booking from the `pending_ext` status? E.g. accepting transaction failed because of multi-currency but then user goes to PayPal.com and resolves the issue, which sends us an IPN message, but then the Harmony call fails. What's the intermediate state for the transaction at this point?

rap1ds:
Closed, see #2620

PR #2618, opened 2016-10-05
Fix overlapping items on frontpage list

mporkola:
Live example of fixes at http://aalto.sharetribe.us/

Before:
screenshot 2016-10-05 10 17 38

After:
screenshot 2016-10-05 10 17 44

### DIFF
--- a/app/views/homepage/_list_item.haml
+++ b/app/views/homepage/_list_item.haml
@@ -32,6 +32,22 @@
             %a.home-share-type-link{:href => search_path(:transaction_type => shape_name_map[listing.listing_shape_id], :view => :list)}
               = icon_tag(listing.icon_name, ["icon-fix"])
               = shape_name(listing)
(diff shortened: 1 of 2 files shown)


app/views/homepage/_list_item.haml
```diff
@@ -32,6 +32,22 @@
             %a.home-share-type-link{:href => search_path(:transaction_type => shape_name_map[listing.listing_shape_id], :view => :list)}
               = icon_tag(listing.icon_name, ["icon-fix"])
               = shape_name(listing)
+      .home-list-price-mobile{:class => (listing.listing_images.size > 0 ? "home-list-price-mobile-with-listing-image" : "home-list-price-mobile-without-listing-image")}
```
mporkola: No change in content, just moved mobile price block `.home-list-price-mobile` inside the div containing listing title. The price block used to be absolutely positioned, which caused overlap. Now it's in normal page layout flow.

PR #2619, opened 2016-10-06
Display user initials if no avatar image is found

mporkola:
screenshot 2016-10-06 12 51 59

### DIFF
--- a/client/app/components/elements/Avatar/Avatar.css
+++ b/client/app/components/elements/Avatar/Avatar.css
@@ -6,4 +6,14 @@
 .link {
   display: inline-block;
   height: 100%;
+  width: 100%;
+}
+
+.textAvatar {
+  width: 100%;
+  border-radius: 50%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  color: var(--Avatar_textColor);
 }
--- a/client/app/models/ProfileModel.js
+++ b/client/app/models/ProfileModel.js
@@ -4,7 +4,7 @@ export const Profile = Immutable.Record({
   familyName: 'family name',
   givenName: 'given name',
   description: 'product author',
-  avatarURL: 'https://placehold.it/40x40',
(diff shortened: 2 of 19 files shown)


client/app/components/elements/Avatar/Avatar.css
```diff
@@ -6,4 +6,14 @@
...
+  width: 100%;
+}
+
+.textAvatar {
+  width: 100%;
+  border-radius: 50%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  color: white;
```
Gnito: Color could come from variables.js
Have you thought about font-sizes? Should the font size be what ever is defined outside of Avatar or should it be proportional to avatar components width (or height). That way it would scale always depending on how big avatar we are going to use.

(The latter would mean either direct % value, or [css lock formula](http://blog.typekit.com/2016/08/17/flexible-typography-with-css-locks/) The idea behind the formula is that there's defined min & max, but the value can increase within those bounds.)

I'm not sure which one is better. In Topbar, it makes sense that the font-size is the same as in other elements. However, in ListingCard it might make more sense that the size is proportional (and in ListingPage design there was some 3rd size of Avatar used).

client/app/models/ProfileModel.js
```diff
@@ -4,14 +4,14 @@ export const Profile = Immutable.Record({
...
   profileURL: 'https://example.com/anonym',
 });
 
 export const parse = (profile) => new Profile({
   familyName: profile.getIn([':attributes', ':familyName']),
   givenName: profile.getIn([':attributes', ':givenName']),
   description: profile.getIn([':attributes', ':description']),
-  avatarURL: 'https://placehold.it/40x40',
-  profileURL: 'https://example.com/anonym',
+  avatarURL: null,
```
mporkola: It's going to be a `ResponsiveImage` reference

PR #2620, opened 2016-10-06
Accept and reject booking

rap1ds:
This PR implements booking accept/reject.

When the transaction status is moved to `paid`, we will accept the booking. In addition, when the status is moved to `rejected`, we will reject the booking.

The implementation is now optimistic. There's no error handling mechanism if the booking status change fails. Failed booking status change doesn't prevent the transaction from moving on. If the booking status change fails, we will only log and trigger Airbrake error, but nothing else. In the future, we will need to build a proper error handling mechanism.

### DIFF
--- a/app/services/marketplace_service/transaction.rb
+++ b/app/services/marketplace_service/transaction.rb
@@ -377,7 +377,7 @@ module MarketplaceService
           },
           body: {
             actorId: UUIDUtils.base64_to_uuid(transaction[:listing_author_id]),
-            reason: "provicer accepted"
+            reason: :provider_accepted
           },
           opts: {
             max_attempts: 3
@@ -402,7 +402,7 @@ module MarketplaceService
             # Passing the reason to the event handler is a bit
             # cumbersome. We decided to skip it for now. That's why
             # we always set the reason to "unknown"
-            reason: "unknown"
+            reason: :unknown
           },
           opts: {
             max_attempts: 3
--- a/app/services/marketplace_service/transaction.rb
+++ b/app/services/marketplace_service/transaction.rb
@@ -3,6 +3,9 @@ module MarketplaceService
     TransactionModel = ::Transaction
     ParticipationModel = ::Participation
 
+    class BookingStateChangeError < StandardError
+    end
+
     module Entity
       Transaction = EntityUtils.define_entity(
         :id,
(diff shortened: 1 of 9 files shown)


app/services/marketplace_service/transaction.rb
```diff
@@ -354,6 +367,52 @@ def preauthorized(transaction, payment_type)
...
+          params: {
+            id: transaction[:booking_uuid]
+          },
+          body: {
+            actorId: UUIDUtils.base64_to_uuid(transaction[:listing_author_id]),
+
+            # Passing the reason to the event handler is a bit
+            # cumbersome. We decided to skip it for now. That's why
+            # we always set the reason to "unknown"
+            reason: "unknown"
```
ovan: :unknown, I'd prefer these to be keywords.

app/services/marketplace_service/transaction.rb
```diff
@@ -354,6 +367,52 @@ def preauthorized(transaction, payment_type)
...
+        return unless transaction[:availability].to_sym == :booking
+
+        HarmonyClient.post(
+          :accept_booking,
+          params: {
+            id: transaction[:booking_uuid]
+          },
+          body: {
+            actorId: UUIDUtils.base64_to_uuid(transaction[:listing_author_id]),
+            reason: "provicer accepted"
```
ovan: :provider-accepted ?

app/services/transaction_service/paypal_events.rb
```diff
@@ -118,8 +118,11 @@ def pending_ext?(pending_reason)
   end
 
   def pending_ext_to_paid(tx)
-    MarketplaceService::Transaction::Command.transition_to(tx[:id], :paid)
+    # Commission charge is synchronous and must complete before we
+    # transition to paid so that we have the full payment info
+    # available at the time we send payment receipts.
     TransactionService::Transaction.charge_commission(tx[:id])
+    MarketplaceService::Transaction::Command.transition_to(tx[:id], :paid)
```
ovan: Why did you want to change the order here? If we follow the current logic charging a commission should be a reaction to transaction having moved to paid, so it too should move out of the paypal event handler similar to harmony notifications.

rap1ds: This comment was so convincing that I thought the order here was wrong: https://github.com/sharetribe/sharetribe/pull/2620/files#diff-df4b745a9b64b2271a4aad63144569beL105

ovan: :) Ok, I guess you better follow it then!

PR #2622, opened 2016-10-07
Avatar image

mporkola:
Display avatar image on listings. Refactored avatar images to use Image record instead of plain URL.

### DIFF
--- a/client/app/models/ImageModel.js
+++ b/client/app/models/ImageModel.js
@@ -1,21 +1,27 @@
 import Immutable from 'immutable';
 
 export const Image = Immutable.Record({
-  type: ':square',
-  height: 408,
-  width: 408,
+  type: null,
+  height: null,
+  width: null,
   url: null,
 });
 
 export const ImageRefs = Immutable.Record({
-  square: new Image(),
-  square2x: new Image(),
+  square: null,
+  square2x: null,
+  medium: null,
+  small: null,
+  thumb: null,
 });
 
 export const parse = (data) => {
   const knownStyles = {
     ':square': 'square',
     ':square_2x': 'square2x',
+    ':medium': 'medium',
+    ':small': 'small',
+    ':thumb': 'thumb',
   };
   const images = data.map(([type, height, width, url]) =>
     new Image({ type, height, width, url }));
--- a/client/app/models/ImageModel.js
+++ b/client/app/models/ImageModel.js
@@ -7,12 +7,25 @@ export const Image = Immutable.Record({
   url: null,
 });
 
-export const ImageRefs = Immutable.Record({
+const ImageRefs = Immutable.Record({
   square: null,
   square2x: null,
   medium: null,
   small: null,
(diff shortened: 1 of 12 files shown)


client/app/models/ProfileModel.js
```diff
@@ -4,14 +4,14 @@ export const Profile = Immutable.Record({
   familyName: 'family name',
   givenName: 'given name',
   description: 'product author',
-  avatarURL: null,
+  avatarImage: null,
```
Gnito: This could be set to correct value ```new Image()``` - makes it easier to understand what this record can hold.

client/app/models/ImageModel.js
```diff
@@ -1,21 +1,27 @@
...
+  type: null,
+  height: null,
+  width: null,
   url: null,
 });
 
 export const ImageRefs = Immutable.Record({
-  square: new Image(),
-  square2x: new Image(),
+  square: null,
```
Gnito: Also here, I'd prefer: ```new Image()```

client/app/models/ImageModel.js
```diff
@@ -1,21 +1,32 @@
...
-  square2x: new Image(),
+const ImageRefs = Immutable.Record({
+  square: null,
+  square2x: null,
+  medium: null,
+  small: null,
+  thumb: null,
+});
+
+export const ResponsiveImage = Immutable.Record({
```
Gnito: I thought that ImageRefs would be ResponsiveImage - i.e. object that holds all the different image formats for one original image. I'm not sure what's the benefit of having this separate model.

We probably should rename these styles in DiscoveryAPI since they clash otherwise - ping @sktoiva

Listing images:
```ruby
        :small_3x2 => "240x160#",
        :medium => "360x270#",
        :thumb => "120x120#",
        :original => "#{APP_CONFIG.original_image_width}x#{APP_CONFIG.original_image_height}>",
        :big => Proc.new { |instance| instance.crop_big },
        :email => "150x100#",
        :square => "408x408#",
        :square_2x => "816x816#"}
```
person avatar:
```ruby
         :medium => "288x288#",
         :small => "108x108#",
         :thumb => "48x48#",
         :original => "600x800>"}
```

So, after some thinking I kind of suggest that we'll make something like this:
```javascript
export const ResponsiveImage = Immutable.Record({
  'square_48x48': null,
  'square_108x108': null,
  'square_288x288': null,
  'square_408x408': null,
  'square_816x816': null,
  'portrait_200x300': null, // future format
  'portrait_400x600': null, // future format
  'original': null, // future format
  'origina_small': null, // future format
  'original_medium': null, // future format
});
```

Gnito: Alternatively, ResponsiveAvatarImage & ResponsiveListingImage could be separated since they have different crop sizes in the data source (i.e. R&A).

Gnito:
Tests fail, but otherwise looks good to me 👍

PR #2627, opened 2016-10-10
Transit uuid not available yet

Gnito:
Test data gives us UUIDs and (in some point in the future) we should get those from real data too.
Currently DiscoveryAPI will give us strings instead.

Related:
Why key is needed: http://revelry.co/dynamic-child-components-with-react/
Problem: https://github.com/uber/r-dom/issues/19 (i.e. we just need to remember this)

### DIFF
--- a/client/app/components/sections/SearchPage/SearchPage.js
+++ b/client/app/components/sections/SearchPage/SearchPage.js
@@ -24,8 +24,9 @@ class SearchPage extends Component {
   }
 
   listingProps(listing, color) {
+    const listingKey = listing.id.toString();
     return {
-      key: `card_${listing.id}`,
+      key: `card_${listingKey}`,
       color,
       listing,
     };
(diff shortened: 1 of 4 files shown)


client/app/components/sections/SearchPage/SearchPage.js
```diff
@@ -24,8 +25,11 @@ class SearchPage extends Component {
   }
 
   listingProps(listing, color) {
+
+    // DiscoveryAPI doesn't return UUID yet but a string
+    const listingKey = listing.id instanceof UUID ? listing.id.value : listing.id;
```
mporkola: String is a perfectly valid UUID representation. I think this should be handled in `transitImmutableConverter`, and have a shared way of turning UUIDs to strings.

Gnito: If we use Immutable.Record to represent our UUIDs then this info about type (i.e. UUID model) needs to be available.

PR #2628, opened 2016-10-10
Order type used as default label if no price is found

Gnito:
screen shot 2016-10-10 at 17 42 33
screen shot 2016-10-10 at 17 43 17

DiscoveryAPI's test data doesn't cover these yet.

### DIFF
--- a/client/app/components/composites/ListingCard/ListingCard.css
+++ b/client/app/components/composites/ListingCard/ListingCard.css
@@ -85,15 +85,15 @@
   align-items: flex-end;
 }
 
-.priceWrapper {
-  flex-basis: 68%;
+.orderTypeWrapper {
+  flex-basis: 67%;
+  font-size: var(--ListingCard_fontSizePrice);
+  font-weight: var(--ListingCard_fontWeightMedium);
+  line-height: 1em;
 }
 
 .price {
   float: left;
-  font-size: var(--ListingCard_fontSizePrice);
-  font-weight: var(--ListingCard_fontWeightMedium);
-  line-height: 1em;
   margin-right: var(--ListingCard_textSpacing);
 }
 
@@ -106,7 +106,7 @@
 }
 
 .distance {
-  flex-basis: 32%;
+  flex-basis: 33%;
   display: flex;
   flex-direction: row;
   justify-content: flex-end;
--- a/client/app/components/composites/ListingCard/ListingCard.story.js
+++ b/client/app/components/composites/ListingCard/ListingCard.story.js
@@ -13,60 +13,57 @@ import css from './ListingCard.story.css';
 const { storiesOf, specs, expect } = storybookFacade;
 const containerStyle = { style: { background: 'white' } };
 
+
+const basicListingData = {
+  id: 'lkjg84573874yjdf',
+  title: 'Title',
+  images: new Immutable.List([new ListingImage({
+    square: new Image({
+      url: 'https://placehold.it/408x408',
+    }),
+    square2x: new Image({
+      type: 'square2x',
+      width: 816,
+      height: 816,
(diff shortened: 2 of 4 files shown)


client/app/components/composites/ListingCard/ListingCard.css
```diff
@@ -105,6 +105,13 @@
   padding-top: 5px;
 }
 
+.orderTypeLabel {
+  flex-basis: 68%;
+  font-size: var(--ListingCard_fontSizePrice);
+  font-weight: var(--ListingCard_fontWeightMedium);
+  line-height: 1em;
+}
+
```
mporkola: The styling appears to be ~same as in priceWrapper + price, so maybe these could be combined somehow? As they fill the same function, I think they could both be inside the `.priceWrapper` div to make the structure more obvious

client/app/components/composites/ListingCard/ListingCard.story.js
```diff
@@ -78,6 +77,13 @@ const ListingCardNoImage =
           value: 0.67, // eslint-disable-line no-magic-numbers
           unit: ':km',
         }),
+        author: {
+          familyName: 'family name',
+          givenName: 'given name',
+          description: 'product author',
+          avatarURL: 'https://placehold.it/40x40',
+          profileURL: `#profile${Math.random(10)}`, // eslint-disable-line no-magic-numbers
+        },
```
mporkola: Yet another place that's clashing with the avatar profile pic changes, we need to have a central place for generating these :/

client/app/components/composites/ListingCard/ListingCard.story.js
```diff
@@ -113,6 +118,50 @@ const ListingCardImageError =
...
+          value: 12972, // eslint-disable-line no-magic-numbers
+          unit: ':miles',
+        }),
+        author: {
+          familyName: 'family name',
+          givenName: 'given name',
+          description: 'product author',
+          avatarURL: 'https://placehold.it/40x40',
+          profileURL: `#profile${Math.random(10)}`, // eslint-disable-line no-magic-numbers
+        },
```
mporkola: Also this: couldn't this class have a base version of listing card data, and alter it to match the cases?

client/app/components/composites/ListingCard/ListingCard.story.js
```diff
@@ -176,5 +230,21 @@ storiesOf('Search results')
...
+    const mountedCard = mount(card);
+
+    specs(() => describe('ListingCard - basic', () => {
+      it('Should display order type "Giving away"', () => {
+        expect(mountedCard.text()).to.include(card.props.listing.orderType.get('en'));
+        expect(mountedCard.find('.ListingCard_orderType')).to.have.length(1);
+        expect(mountedCard.find('.ListingCard_price')).to.have.length(0);
+      });
+      testDistance(card, mountedCard);
+    }));
```
mporkola: 👍

client/app/components/composites/ListingCard/ListingCard.story.js
Gnito: Probably 👍 
I'll do that after your image branch is merged (i.e. this branch will wait a bit).

client/app/components/composites/ListingCard/ListingCard.css
Gnito: You have a sharp eye :)
I wanted to keep them separated since those styles are not yet final - design review needed.
If we end up merging, I'll suggest a renaming (priceWrapper -> orderTypeWrapper) since price is just one special case of order type.

mporkola: Let's wait until design review then 👍

mporkola:
👍 remember to squash

PR #2629, opened 2016-10-11
Bring notifications to React pages

mporkola:


app/views/layouts/react_page.haml
```diff
@@ -5,6 +5,9 @@
   -# Render railsContext here before any react component gets rendered
   = prepend_render_rails_context("")
 
+  - if notifications
+    = react_component("Notifications", props: notifications, prerender: true)
```
Gnito: I'm making visual components now in separate branch. As we discussed earlier, I'm assuming that the notifications are passed to Topbar instead of some separate NotificationsApp.

What comes to naming: I'm currently using FlashMessage (since notification word is already used in the front end and it's reserved for notifications about new unread Inbox messages (which is to say that notifications should be something that happens between users instead of between user and platform).

PR #2630, opened 2016-10-11
Moved some translation keys back to main locale file

thomasmalbaux:
When moving some translation keys related to the welcome email to their own yml file in order to remove them from WTI (see [this PR](https://github.com/sharetribe/sharetribe/pull/2611/files)), I actually moved too many of them: some translation keys related to the user welcome emails, and not only the new admin welcome email, were moved too.

This of course shouldn't have happened.

This PR moves the correct keys to the main yml file and removes them from the dedicated welcome_email yml files.

And unless there's an easier and faster way (for me ;-)), I'll simply add the translation (fr, es-ES, ...) back to WTI directly.


thomasmalbaux:
And unless there's an easier and faster way (for me ;-)), I'll simply add the translation (fr, es-ES, ...) back to WTI directly.

rap1ds:
LGTM

PR #2632, opened 2016-10-12
Retry only on error and if status is 5xx

rap1ds:
Change the Retry middleware that it retries only is status is `5xx` or an exception happens.

### DIFF
--- a/app/utils/service_client/middleware/retry.rb
+++ b/app/utils/service_client/middleware/retry.rb
@@ -41,8 +41,10 @@ module ServiceClient
         ctx.fetch(:enter_queue).dup + [self]
       end
 
+      # Retries if all attempts are not used and
+      # status is 5xx
       def leave_needs_retry?(ctx)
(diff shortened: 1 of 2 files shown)


app/utils/service_client/middleware/retry.rb
```diff
@@ -41,8 +41,10 @@ def build_retry_queue(ctx)
         ctx.fetch(:enter_queue).dup + [self]
       end
 
+      # Retries if all attempts are not used and
+      # status is 5xx
       def leave_needs_retry?(ctx)
-        !max_attempts?(ctx) && !ctx.fetch(:res).fetch(:success)
+        !max_attempts?(ctx) && (500..599).include?(ctx.fetch(:res).fetch(:status))
```
ovan: Will this result in a sequential comparison against all numbers from 500-> one at a time? >= 500 and < 600 would be faster and simpler.

rap1ds: Range is more cleaver than that :)

PR #2638, opened 2016-10-13
FlashMessage component

Gnito:
This adds FlashNotifications to the Redux store as well as a component which is capable of showing those messages.

Naming: I chose FlashNotifications to distinguish these platform notifications from normal user-to-user notifications (like inbox message notifications which we use or similar notifications in Facebook / Twitter). 

How to use FlashNotifications:

- use action creator (FlashNotificationActions)
```javascript
this.props.actions.addFlashNotification('error', 'Dynamic error message!!!');
```


client/app/startup/SearchPageApp.js
```diff
@@ -56,7 +73,15 @@ export default (props) => {
     currentPage: rawListings.map((l) => l.get(':id')),
   });
 
-  const combinedProps = Object.assign({}, { marketplace: props.marketplace }, { searchPage, routes, listings, profiles });
+  const flashNotifications = systemNotificationsToList(props.marketplace.notifications);
```
mporkola: Now we have the notifications in two different parts of redux state: `flashNotifications` and `marketplace.notifications`. Not sure how this should be fixed, but it's not very clean now.

PR #2639, opened 2016-10-13
Harmony API JWT authentication

kpuputti:
This PR adds JWT authentication to all HarmonyClient API calls if the authentication configuration is enabled. The token needs an authentication context hash with `:marketplace_id` and `:actor_id` keys with their corresponding UUIDs as values.

### DIFF
--- a/app/models/transaction.rb
+++ b/app/models/transaction.rb
@@ -87,6 +87,14 @@ class Transaction < ActiveRecord::Base
     self.booking_uuid = UUIDUtils.raw(uuid)
   end
 
+  def community_uuid_object
+    if self[:community_uuid].nil?
+      nil
+    else
+      UUIDUtils.parse_raw(self[:community_uuid])
+    end
+  end
+
   def status
     current_state
   end
--- a/app/services/marketplace_service/transaction.rb
+++ b/app/services/marketplace_service/transaction.rb
@@ -10,6 +10,7 @@ module MarketplaceService
       Transaction = EntityUtils.define_entity(
         :id,
         :community_id,
+        :community_uuid,
         :last_transition,
         :last_transition_at,
         :listing,
@@ -104,6 +105,7 @@ module MarketplaceService
           booking: transaction_model.booking,
           availability: transaction_model.availability.to_sym,
           booking_uuid: transaction_model.booking_uuid_object,
+          community_uuid: transaction_model.community_uuid_object,
           __model: transaction_model
         })]
       end
@@ -370,6 +372,11 @@ module MarketplaceService
       def paid(transaction)
         return unless transaction[:availability].to_sym == :booking
 
+        auth_context = {
+          marketplace_id: transaction[:community_uuid],
+          actor_id: UUIDUtils.base64_to_uuid(transaction[:listing_author_id])
+        }
+
         HarmonyClient.post(
           :accept_booking,
           params: {
@@ -380,7 +387,8 @@ module MarketplaceService
             reason: :provider_accepted
           },
           opts: {
-            max_attempts: 3
+            max_attempts: 3,
+            auth_context: auth_context
           }).on_error { |error_msg, data|
           log_and_notify_harmony_error!("Failed to accept booking",
                                         :failed_accept_booking,
@@ -390,6 +398,11 @@ module MarketplaceService
 
       def rejected(transaction)
         return unless transaction[:availability].to_sym == :booking
+
+        auth_context = {
+          marketplace_id: transaction[:community_uuid],
+          actor_id: UUIDUtils.base64_to_uuid(transaction[:listing_author_id])
+        }
 
         HarmonyClient.post(
           :reject_booking,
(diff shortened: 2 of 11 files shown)


app/services/marketplace_service/transaction.rb
```diff
@@ -29,6 +29,7 @@ module Entity
         :created_at,
         :availability,
         :booking_uuid,
+        :auth_data,
```
rap1ds: The `Transaction` entity is used to convert Model objects to Hashes. See line: https://github.com/sharetribe/sharetribe/pull/2639/files#diff-e2097c0e3cb9d60a2686ec38668efb43L92

```
Transaction[EntityUtils.model_to_hash(transaction_model)]
```

It's good to keep in mind that the `Transaction` entity here is a little bit different that the other Entities you may find from the code base. As you can see, the entity is builded by using method `.define_entity`. Which is actually a bit older and simpler version of what we use today, `.define_builder`.

In the line above, we have the `transaction_model`, which is an ActiveRecord model. EntityUtils.model_to_hash takes the model `attributes` hash and symbolizes the keys. Then that hash is passed to `Transaction` entity. Entities are usually called with `.call`, but since the `.define_entity` method returns a lambda, and lambda aliases `.call` to `[]`, `Transaction[some_hash]` equals `Transaction.call(some_hash)`. This might be a bit misleading.

So the reason why `auth_data` is empty is that the `auth_data` can't be found from the model. I guess you should build the `auth_data` in the `paid` method where you are calling the `HarmonyClient`.

app/utils/service_client/middleware/jwt_authenticator.rb
```diff
@@ -0,0 +1,25 @@
+module ServiceClient
+  module Middleware
+
+    TOKEN_DATA_KEYS = Set.new([:marketplaceId, :actorId, :role])
+
+    class JwtAuthenticator < MiddlewareBase
+      def enter(ctx)
+        unless APP_CONFIG.harmony_api_disable_authentiation
```
rap1ds: Instead of accessing the `APP_CONFIG` global, I'd recommend to add a `.initialize` method and pass the value as a constructor parameter. Makes testing much easier.

app/utils/service_client/middleware/jwt_authenticator.rb
```diff
@@ -0,0 +1,25 @@
...
+          token = create_token(ctx[:opts][:auth_data])
+          ctx[:req][:headers]["Authorization"] = "Token #{token}"
+        end
+        ctx
+      end
+
+      private
+
+      def create_token(auth_data)
+        if !auth_data || auth_data.keys.to_set != TOKEN_DATA_KEYS
```
rap1ds: Would it make sense to make a new TokenData entity? Then you could add types (uuid for marketplace and actor and enum for role) to the entity.

app/utils/service_client/middleware/jwt_authenticator.rb
```diff
@@ -0,0 +1,25 @@
...
+          raise ArgumentError.new("Invalid authentiction data")
+        end
+        JWTUtils.encode(auth_data, APP_CONFIG.harmony_api_token_secret, exp: 5.minutes.from_now)
```
rap1ds: Same comment here about `APP_CONFIG`

app/models/person.rb
```diff
@@ -180,6 +180,16 @@ def uuid_object
     UUIDTools::UUID.parse_raw(Base64.urlsafe_decode64(self.id))
   end
 
+  def role
+    if self.is_admin?
+      :superAdmin
```
rap1ds: Convention in Ruby is to use :snake_case

config/config.defaults.yml
```diff
@@ -286,6 +286,8 @@ default: &default_settings
 
   # Harmony service API connection
   harmony_api_url:
+  harmony_api_disable_authentiation: false
```
rap1ds: s/authentiation/authentication

app/utils/service_client/middleware/jwt_authenticator.rb
```diff
@@ -0,0 +1,27 @@
+module ServiceClient
+  module Middleware
+
+    TOKEN_DATA_KEYS = Set.new([:marketplace_id, :actor_id])
+
+    class JwtAuthenticator < MiddlewareBase
+      def enter(ctx)
+        unless APP_CONFIG.harmony_api_disable_authentiation
```
rap1ds: s/authentiation/authentication

app/models/transaction.rb
```diff
@@ -87,6 +87,18 @@ def booking_uuid_object=(uuid)
...
 
+  def community_uuid_object
+    if self[:community_uuid].nil?
+      nil
+    else
+      UUIDUtils.parse_raw(self[:community_uuid])
+    end
+  end
+
+  def community_uuid_object=(uuid)
```
rap1ds: I think we could also live without the setter. The UUID shouldn't be changed after it's set the first time.

PR #2640, opened 2016-10-13
Improve error messages on booking error

rap1ds:
This PR adds better error messages if booking fails. It also differentiates two cases: double booking and other booking error, i.e. connection error.

This PR also changes the "process token" logic a bit. The `finalize_create` method fetches now an existing `process_token` and returns that if it exist. This way the view can show the error message that is stored to the process_token result.

### DIFF
--- a/app/services/transaction_service/transaction.rb
+++ b/app/services/transaction_service/transaction.rb
@@ -4,8 +4,10 @@ module TransactionService::Transaction
   end
 
   DataTypes = TransactionService::DataTypes::Transaction
+  ProcessStatus = TransactionService::DataTypes::ProcessStatus
 
   TxStore = TransactionService::Store::Transaction
+  ProcessTokenStore = TransactionService::Store::ProcessToken
 
   DEPRECATED_GATEWAYS = [:braintree, :checkout]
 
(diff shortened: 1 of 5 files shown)


config/locales/unreleased_features/en.yml
```diff
@@ -12,7 +12,8 @@ en:
       error_in_checking_availability: "Could not check availability for the selected dates"
   error_messages:
     booking:
-      booking_failed_payment_voided: "Failed to create a booking. Your payment will be refunded."
+      booking_failed_payment_voided: "Booking failed due to an unexpected error. Please try again later. You haven't been charged."
+      double_booking_payment_voided: "Unfortunately the dates you chose are no longer available. Please choose another dates. You haven’t been charged."
```
ovan: s/another/other. Has @Kusti checked these?

app/services/transaction_service/transaction.rb
```diff
@@ -100,30 +102,48 @@ def create(opts, force_sync: true)
...
+        tx = m_tx.get
+
+        tx_process = tx_process(tx[:payment_process])
+        gw = gateway_adapter(tx[:payment_gateway])
+
+        tx_process.finalize_create(
+          tx: tx,
+          gateway_adapter: gw,
+          force_sync: force_sync)
+      end
```
ovan: This logic could be simplified. :) There's nested ifs and m.get / m.is_none? used which smells like there's room for improvement.

PR #2641, opened 2016-10-14
Move Topbar inside SearchPage (other pages still use TopbarApp)

Gnito:
What happens after this PR
- TopbarApp is used on basic layout (application.haml)
- However, when react_page.haml layout is used Topbar's props are passed to SearchPageApp and Topbar is initialized inside SearchPage.js


Gnito:
I wasn't sure what should be done with maps library, since that would raise a question about where other script tags should be added - in layout or inside component?

Gnito:
MikkoP checked this

PR #2643, opened 2016-10-14
Fix preview link target for CLP users

Kusti:
One line fix to make sure preview link in admin panel has correct target (landing page of the marketplace) also when custom landing page is in use.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -334,7 +334,7 @@ module ApplicationHelper
       :topic => :general,
       :text => t("admin.left_hand_navigation.preview"),
       :icon_class => icon_class("eye"),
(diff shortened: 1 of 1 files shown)


app/helpers/application_helper.rb
```diff
@@ -334,7 +334,7 @@ def admin_links_for(community)
       :topic => :general,
       :text => t("admin.left_hand_navigation.preview"),
       :icon_class => icon_class("eye"),
-      :path => search_path(big_cover_photo: true),
+      :path => homepage_without_locale_path(big_cover_photo: true),
```
rap1ds: In other places where this path helper is used, we also set `locale: nil`. Which makes me think, should we set it also here?

PR #2653, opened 2016-10-18
Add UUID to People

rap1ds:
- [x] Add `uuid` column to `people`
- [x] Save `uuid`s when a new Person is created
- [x] Populate `people.uuid`
- [x] Add uniq and not null index to `people.uuid`
- [ ] Deploy [0/3]
  - [ ] Stop indexer
  - [ ] Run migration that adds `people.uuid` column
  - [ ] Wait until new code is taken into use
  - [ ] Run populate and add index migrations
  - [ ] Restart indexer

### DIFF
--- a/db/migrate/20161018090517_populate_people_uuid.rb
+++ b/db/migrate/20161018090517_populate_people_uuid.rb
@@ -1,0 +1,49 @@
+class PopulatePeopleUuid < ActiveRecord::Migration
+
+  class Person < ActiveRecord::Base
+  end
+
+  def up
+    Person.order(:created_at).pluck(:id).each_slice(1000) { |ids|
+      selects = ids.map { |id|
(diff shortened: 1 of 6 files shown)


db/migrate/20161018090517_populate_people_uuid.rb
```diff
@@ -0,0 +1,49 @@
...
+                 #{selects.join("\nUNION ALL\n")}
+               ) joined
+               ON orig.id = joined.id
+               SET orig.uuid = joined.uuid"
+
+      exec_update(sql, "Update people UUIDs", [])
+    }
+  end
+
+  ## UUIDUtils
```
kpuputti: Is the point of inlining this code to avoid using code elsewhere that might change and make this migration possibly incorrect later?

rap1ds: Yep, exactly.

PR #2660, opened 2016-10-19
Add URL to listing page to listing data

Gnito:
This adds listingURL to listing's data model (alternative is to pass routes instance everywhere).

### DIFF
--- a/client/app/models/ListingModel.js
+++ b/client/app/models/ListingModel.js
@@ -37,7 +37,7 @@ const parseListingImages = (images) => new ListingImage({
   square2x: images.square2x,
 });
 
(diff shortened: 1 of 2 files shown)


client/app/models/ListingModel.js
```diff
@@ -37,14 +37,15 @@ const parseListingImages = (images) => new ListingImage({
   square2x: images.square2x,
 });
 
-export const parse = (l) => {
+export const parse = (l, formListingPath) => {
```
mporkola: the naming is a bit obscure, maybe `listingPathRoute` or `getListingPath` or something?

PR #2663, opened 2016-10-20
Pagination for new browse view

mporkola:
Search page pagination, without any visual styling yet

### DIFF
--- a/client/app/components/sections/SearchPage/SearchPage.js
+++ b/client/app/components/sections/SearchPage/SearchPage.js
@@ -1,8 +1,9 @@
 import { Component, PropTypes } from 'react';
-import r, { div } from 'r-dom';
+import r, { div, a } from 'r-dom';
 import Immutable from 'immutable';
 import styleVariables from '../../../assets/styles/variables';
 import { routes as routesProp } from '../../../utils/PropTypes';
+import { upsertSearchQueryParam } from '../../../utils/url';
 
 import Topbar from '../../sections/Topbar/Topbar';
 import ListingCard from '../../composites/ListingCard/ListingCard';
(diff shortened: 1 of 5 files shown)


client/app/components/sections/SearchPage/SearchPage.js
```diff
@@ -22,6 +23,35 @@ class SearchPage extends Component {
...
+
+  prevPage() {
+    return this.props.searchPage.state.get('page') - 1;
+  }
+
+  setPage(num) {
+    return (e) => {
+      e.preventDefault();
+      const next = this.pageUrl(num);
+      window.location = next;
```
Gnito: ```window.location = this.pageUrl(num);```
(const *next* is probably just confusing due to another meaning in this context.)

client/app/components/sections/SearchPage/SearchPage.js
```diff
@@ -41,6 +71,11 @@ class SearchPage extends Component {
         this.listings.map((listing) =>
           r(ListingCard, this.listingProps(listing, marketplaceColor1))
       )),
+      div({}, [
+        `${this.props.searchPage.state.get('total')} listings, page ${this.props.searchPage.state.get('page')}/${this.totalPages}`,
```
Gnito: I guess this is test code? (Otherwise we need to localize those texts.)

client/app/utils/url.js
```diff
@@ -73,9 +75,18 @@ const currySearchParams = function currySearchParams(restrict_to_params) {
   };
 };
 
+const upsertSearchQueryParam = function upsertSearchQueryParam(location, param, value) {
+  const originalParams = parseSearchQueryParams(location);
+  const newParams = { ...originalParams, [param]: value };
```
Gnito: *value* is probably now mutable since it's not cloned with Object.assign or spread operator. Although, this returns  a string map, so it doesn't matter.

PR #2665, opened 2016-10-20
Update Node.js version

kpuputti:
This PR updates Node.js version to the recently released LTS (long term support) version 6.9.

There is also an extra script run before each NPM script to ensure that the Node.js in use is the correct one.

## TODO

 - [x] Update NPM packages
 - [x] Update Node.js version to the latest LTS 6.9
 - [x] Add a version check to scripts to enforce correct Node.js version
 - [x] Generate a new shrinkwrap file
 - [x] Ensure everything works in Heroku


rap1ds:
Looks good to me but maybe a good idea if someone how knows Node better than me checks this also. Ping @Gnito @mporkola

mporkola:
🚢

PR #2673, opened 2016-10-24
Fix person view

lyyder:
This PR aims to fix the person view which breaks when using the Discovery API as search backend.


lyyder:
Solved in #2675, closing this.

PR #2675, opened 2016-10-24
Discovery client

lyyder:
Adds a discovery client that is used to call the discovery API. Also removes the discovery adapter and makes the app use the client to search for listings.


app/controllers/homepage_controller.rb
```diff
@@ -181,26 +181,73 @@ def find_listings(params, current_page, listings_per_page, filter_params, includ
...
-      ).and_then { |res|
-      Result::Success.new(
-        if FeatureFlagHelper.search_engine == :discovery
-          res
-        else
+    if FeatureFlagHelper.feature_enabled?(:searchpage_v1)
+      DiscoveryClient.get(:query_listings,
+                          params: format_params(search.merge(marketplace_id: @current_community.id)))
+      .rescue {
+        binding.pry
```
kpuputti: Remove breakpoint.

app/controllers/homepage_controller.rb
```diff
@@ -181,26 +181,72 @@ def find_listings(params, current_page, listings_per_page, filter_params, includ
...
+     :'page[number]' => original[:page],
+     :'page[size]' => original[:per_page],
+     :'filter[price_min]' => Maybe(original[:price_cents]).map{ |p| p.min }.or_else(nil),
+     :'filter[price_max]' => Maybe(original[:price_cents]).map{ |p| p.max }.or_else(nil),
+     :'filter[omit_closed]' => !original[:include_closed],
+     :'filter[listing_shape_ids]' => Maybe(original[:listing_shape_ids]).join(",").or_else(nil),
+     :'filter[category_ids]' => Maybe(original[:categories]).join(",").or_else(nil),
+     :'search[locale]' => original[:locale],
+     :sort => original[:sort]
+    }.merge(location_params).merge(custom_fields).compact
```
mporkola: I'm not sure if `HomepageController` should be further bloated by Discovery-specific parameter formatting

lyyder: True that, moved to a utils module.

app/utils/service_client/middleware/body_encoder.rb
```diff
@@ -97,7 +98,12 @@ def leave(ctx)
         encoder = encoder_by_content_type(headers["Content-Type"]) || @_request_encoder
 
         begin
-          ctx[:res][:body] = encoder[:encoder].decode(body)
+          ctx[:res][:body] =
+            if @_decode_response
+              encoder[:encoder].decode(body)
+            else
+              body
+            end
```
ovan: This functionality could be implemented by setting the encoder's decode method to identity function at configuration time in case the @_decode_response is false. But it comes down to stylistic choice, one more if per request is unlikely to make a difference.

app/utils/discovery_utils.rb
```diff
@@ -0,0 +1,43 @@
+module DiscoveryUtils
+
+  module_function
+
+  def format_params(original)
```
ovan: I'm thinking about the naming here. To me it's not obvious what DiscoveryUtils contains and less so what DiscoveryUtils::format_params does. Format what params? How? Is this generic or something very specific to this one search query? Do we need this functionality across controllers?

PR #2676, opened 2016-10-24
downcase username before checking availability on blacklist

gnomet:
Before this, the username availability check in the UI didn't complain for "ADmin" although "admin" is in the blacklist.
Now compares always downcase version to the list where all is downcases.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -218,7 +218,7 @@ class Person < ActiveRecord::Base
   end
 
   def self.username_available?(username, community_id)
-    !username.in?(USERNAME_BLACKLIST) &&
+    !username.downcase.in?(USERNAME_BLACKLIST) &&
       !Person
         .where("username = :username AND (is_admin = '1' OR community_id = :cid)", username: username, cid: community_id)
         .present?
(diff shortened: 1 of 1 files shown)


app/models/person.rb
```diff
@@ -218,7 +218,7 @@ def self.username_blacklist
   end
 
   def self.username_available?(username, community_id)
-    !username.in?(USERNAME_BLACKLIST) &&
+    !username.downcase.in?(USERNAME_BLACKLIST) &&
```
sktoiva: Maybe turn this the other way around:

```ruby
USERNAME_BLACKLIST.include?(username.downcase)
```

The `in?` method actually does excatly that an is a Rails addition to Object. Personally, I'd prefer the use of core Ruby methods.

Maybe even consider adding to the end of line 158 `to_set`. There's no need for duplicate entries in the blacklist and `include?` is faster on sets (though here I doubt it matters much).

```ruby
 USERNAME_BLACKLIST = YAML.load_file("#{Rails.root}/config/username_blacklist.yml").to_set
```

gnomet: Ok, good tips! Making those changes now!

gnomet: although adding to_set seemed to break a validation where that list was used, so left that out for now, but did the other change to use include? rather than in?

PR #2678, opened 2016-10-24
Make allowed email check case-insensitive

gnomet:


app/services/email_service.rb
```diff
@@ -82,13 +82,14 @@ def self.is_only_notification_email?(all_emails, email)
...
     return true unless allowed_emails.present?
+    return false if address.nil?
 
     allowed = false
     allowed_array = allowed_emails.split(",")
     allowed_array.each do |allowed_domain_or_address|
       allowed_domain_or_address.strip!
       allowed_domain_or_address.gsub!('.', '\.') #change . to be \. to only match a dot, not any char
-      if address =~ /#{allowed_domain_or_address}$/
+      if address.downcase =~ /#{allowed_domain_or_address.downcase}$/
```
bladealslayer: There's better way to match case-insensitively - using the `i` modifier for the regex:

```
if address =~ /#{allowed_domain_or_address}$/i
```

PR #2682, opened 2016-10-25
Remove broken marker labels

mporkola:
For some reason, we tried to set an empty label to all map markers. The code was broken, so it was displayed as `[Object object]`. Not setting the label at all works better.


Gnito:
Such Code, Much Confusion, Very WTF

PR #2697, opened 2016-10-31
Allow admin to access admin panel features without confirmed email

gnomet:
The idea is that instead of requiring email confirmation as very first step after marketplace creation, we direct new admin to Onboarding Wizard in the Admin Dashboard and allow them to play around in the admin side, and only require email confirmation before accessing the user side of the marketplace.

This PR allows admin to use admin panel without confirmed email, changes the URL that is returned to marketing site after new trial site has been created and changes the instruction texts that admin sees on the confirmation_pending page.

There are few specific questions for the reviewer, marked separately as line comments.

### DIFF
--- a/app/controllers/admin/categories_controller.rb
+++ b/app/controllers/admin/categories_controller.rb
@@ -1,9 +1,4 @@
-class Admin::CategoriesController < ApplicationController
-
-  before_filter :ensure_is_admin
-
-  #Allow admin to access admin panel before email confirmation
-  skip_filter :cannot_access_without_confirmation
+class Admin::CategoriesController < Admin::AdminAreaController
 
   def index
     @selected_left_navi_link = "listing_categories"
--- a/app/controllers/admin/categories_controller.rb
+++ b/app/controllers/admin/categories_controller.rb
@@ -2,6 +2,9 @@ class Admin::CategoriesController < ApplicationController
 
   before_filter :ensure_is_admin
 
+  #Allow admin to access admin panel before email confirmation
+  skip_filter :cannot_access_without_confirmation
+
   def index
     @selected_left_navi_link = "listing_categories"
     @categories = @current_community.top_level_categories.includes(:translations, children: :translations)
--- a/app/controllers/admin/categories_controller.rb
+++ b/app/controllers/admin/categories_controller.rb
@@ -1,4 +1,4 @@
-class Admin::CategoriesController < Admin::AdminAreaController
+class Admin::CategoriesController < Admin::AdminBaseController
 
   def index
     @selected_left_navi_link = "listing_categories"
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -205,7 +205,7 @@ class ApplicationController < ActionController::Base
   def ensure_user_belongs_to_community
     return unless @current_user
 
-    if !@current_user.is_admin? && @current_user.accepted_community != @current_community
+    if !@current_user.has_admin_rights? && @current_user.accepted_community != @current_community
 
       logger.info(
         "Automatically logged out user that doesn't belong to community",
--- a/app/controllers/int_api/marketplaces_controller.rb
+++ b/app/controllers/int_api/marketplaces_controller.rb
@@ -46,8 +46,12 @@ class IntApi::MarketplacesController < ApplicationController
         locale: params[:marketplace_language]},
         marketplace[:id]).data
 
+    # make the marketplace creator land to the admin panel
+    url = "#{marketplace[:url]}/admin"
+
+    # make the marketplace creator be logged in via Auth Token
     auth_token = UserService::API::AuthTokens.create_login_token(user[:id])
-    url = URLUtils.append_query_param(marketplace[:url], "auth", auth_token[:token])
+    url = URLUtils.append_query_param(url, "auth", auth_token[:token])
 
     # Enable specific features for all new trials
     FeatureFlagService::API::Api.features.enable(community_id: marketplace[:id], person_id: user[:id], features: [:topbar_v1])
--- a/app/controllers/int_api/marketplaces_controller.rb
+++ b/app/controllers/int_api/marketplaces_controller.rb
@@ -46,15 +46,8 @@ class IntApi::MarketplacesController < ApplicationController
         locale: params[:marketplace_language]},
         marketplace[:id]).data
 
-    # make the marketplace creator land to the admin panel
-    url_params = {host: marketplace[:url]}
-
-    #include port separately to url_params if it was present
-    if marketplace[:url] && marketplace[:url].split(':')[2]
-      url_params[:port] = marketplace[:url].split(':')[2]
-    end
-
-    url = admin_getting_started_guide_url(url_params)
+    base_url = URI.new(marketplace[:url])
+    url = admin_getting_started_guide_url(host: base_url.host, port: base_url.port)
 
     # make the marketplace creator be logged in via Auth Token
     auth_token = UserService::API::AuthTokens.create_login_token(user[:id])
--- a/app/controllers/int_api/marketplaces_controller.rb
+++ b/app/controllers/int_api/marketplaces_controller.rb
@@ -51,7 +51,7 @@ class IntApi::MarketplacesController < ApplicationController
 
     #include port separately to url_params if it was present
     if marketplace[:url] && marketplace[:url].split(':')[2]
-      url_params.merge!(port: marketplace[:url].split(':')[2])
+      url_params[:port] = marketplace[:url].split(':')[2]
     end
 
     url = admin_getting_started_guide_url(url_params)
--- a/app/controllers/int_api/marketplaces_controller.rb
+++ b/app/controllers/int_api/marketplaces_controller.rb
@@ -47,7 +47,14 @@ class IntApi::MarketplacesController < ApplicationController
         marketplace[:id]).data
 
     # make the marketplace creator land to the admin panel
-    url = "#{marketplace[:url]}/admin"
+    url_params = {host: marketplace[:url]}
+
+    #include port separately to url_params if it was present
+    if marketplace[:url] && marketplace[:url].split(':')[2]
+      url_params.merge!(port: marketplace[:url].split(':')[2])
+    end
+
+    url = admin_getting_started_guide_url(url_params)
 
     # make the marketplace creator be logged in via Auth Token
     auth_token = UserService::API::AuthTokens.create_login_token(user[:id])
--- a/app/controllers/int_api/marketplaces_controller.rb
+++ b/app/controllers/int_api/marketplaces_controller.rb
@@ -46,7 +46,7 @@ class IntApi::MarketplacesController < ApplicationController
         locale: params[:marketplace_language]},
         marketplace[:id]).data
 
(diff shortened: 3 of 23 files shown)


app/views/community_memberships/_confirmation_form.haml
```diff
@@ -2,11 +2,14 @@
   initialize_confirmation_pending_form("#{I18n.locale}","#{email_not_accepted_message}");
 
 - email_to_confirm = @current_user.latest_pending_email_address(@current_community)
+- contact_support_link = link_to t("sessions.confirmation_pending.contact_support_link_text"), "mailto:#{support_email}", "data-uv-trigger" => "contact"
```
gnomet: What is the data-uv-trigger for? I searched through the repo but didn't find full answer, but it seemed to be included in several places where support mailto links are used, so my guess was that it is related to tracking somehow? Anyway seemed better to follow the example and include it here, but of course would be good to know what it does, and not blindly follow examples. So please let me know what's the purpose, and is it right to include it here?

app/controllers/int_api/marketplaces_controller.rb
```diff
@@ -46,8 +46,12 @@ def create
         locale: params[:marketplace_language]},
         marketplace[:id]).data
 
+    # make the marketplace creator land to the admin panel
+    url = "#{marketplace[:url]}/admin"
```
gnomet: I was not very confident when choosing the right place where the /admin should be added (to make new admin land to Onboarding Wizard instead of the Confirmation Pending page).
I chose the controller level, while other possible options would have been in the marketplace service (but if felt better that it returns the normal base url) and modifying the URL at marketing site, but I thought it's better that marketing site doesn't need to know about URL changes in R&A.
So controller feels best, but I hope the reviewer to give second opinion on this choice.

app/controllers/admin/categories_controller.rb
```diff
@@ -2,6 +2,9 @@ class Admin::CategoriesController < ApplicationController
 
   before_filter :ensure_is_admin
 
+  #Allow admin to access admin panel before email confirmation
+  skip_filter :cannot_access_without_confirmation
```
ovan: Is the idea with this change to say that all admin only pages are from now on accessible without email confirmation but all non-admin pages require it? This caught my eye because it appears this skip_filter now goes together with ensure_is_admin filter. skip_filter is a bit of an anti-pattern because it undoes something the parent controller did. Having to do it in a non-exceptional situation (~ in multiple places, as a pattern) indicates that the controller inheritance hierarchy has serious problems.

I have two alternatives in mind here:

* a) introduce a separate admin controller that inherits from application controller and use is as a base class for all admin-only controllers. Similarly introduce a user controller and use it has a base class for non-admin controllers.
* b) remove the filter from application controller and introduce it to all non-admin controllers with before_filter. The option a is preferred because option b makes things only a little bit clearer (explicitly adding something instead of undoing things).

This need for refactoring to admin and user controllers has reared it's head before too. One symptom is these type of skip_filters and other is the sheer size of application controller. We've mitigated the last problem already by moving stuff to helpers and rack middleware. This refactoring would help a lot. The refactoring also doesn't have to be all-in; you can only move the admin and confirmation checks from the application controller to admin controller and user controller and leave rest of the unrelated stuff still to application controller even if it's not truly shared between the two parent controllers. This makes it easier for upcoming changes to keep improving on this by pulling more stuff to these new specialized base controllers.

What do you think?

app/views/community_memberships/_confirmation_form.haml
ovan: Not 100% sure but if I would have to guess it's for the UV widget.

app/controllers/int_api/marketplaces_controller.rb
ovan: More than the place the problem with this is that we're building URLs with string concatenation. This means that there's now an implicit dependency between this code and the code that builds the original url; that is the base url is something you can just append /admin to and end up with a sane result. This assumption is broken if something related to these paths is changed in routes.rb or in the code that builds the url being appended to. Rails has helpers for reverse routing (i.e. building a url) to avoid these problems and those helpers should be used whenever building URLs. I suggest making the change wherever the original url is built instead.

app/controllers/application_controller.rb
```diff
@@ -207,7 +207,7 @@ def ensure_consent_given
   def ensure_user_belongs_to_community
     return unless @current_user
 
-    if !@current_user.is_admin? && @current_user.accepted_community != @current_community
+    if !@current_user.has_admin_rights? && @current_user.accepted_community != @current_community
```
rap1ds: Can you explain why this change is needed?

app/controllers/int_api/marketplaces_controller.rb
rap1ds: To me it feels that this is the right layer 👍 

However, we should use the path helpers instead of hard-coding the URLs (i.e. the hard-coded `/admin` URL here). AFAIK path helpers can take an additional `host` parameter.

app/views/community_memberships/_confirmation_form.haml
rap1ds: To my understanding, it will trigger the UserVoice widget, see the screenshot

screen shot 2016-10-31 at 15 54 01

gnomet: Oh, yea that makes sense. It didn't do it just by adding the attribute, so I guess some JS needs to be prepared too.
On the other hand in this particular case a direct mailto might even be better, as if there are issues with mail confirmation, the knowledge base can't help much more than the "check spam filter" instruction already on the page.. So I might now just leave the attribute out for simplicity, and to encourage trials starters who are in trouble at this stage to speak to a human. :)

app/controllers/application_controller.rb
gnomet: @rap1ds there seems to be and extra check after Hydra, called `ensure_user_belongs_to_community` that logs the user out in case they don't have an approved membership to the community. Now my attempt is to allow admin to use admin panel before having "a fully approved memembership = confirmed email" and this check always threw the admin out. So I extended this exception to cover normal admins in addition to super admins.

I thought that this shouldn't cause risk to the original purpose.  But actually good if you check as you know Hydra bit better, if this could cause any trouble? Could some very old session now be able to hang around in a marketplace they don't belong? (anyway there is a check that should confirm the person has membership with admin status in this community)

Also the comment few lines above suggests that the whole check should be removed at some point. But how can we be sure there are no more old sessions around..?

app/controllers/admin/categories_controller.rb
gnomet: @ovan I decided to first do even less all the way approach and only introduce AdminAreaController. (I chose to add "Area" to the name as AdminController/UserController might sound more like something that handles those directly (compared to PeopleController) but if you think AdminController would be better name, it might be. It was actually the UserController name that I didn't feel fully comfortable with for this purpose. 

Anyway I decided now to leave it out, as I also consider this as a test of if it has any effect to conversion to free admins a bit from the email confirmation. Based on results, we might end up changing this quite soon, so that the limit is no more in the admin panel only (but allow admins to also view the marketplace without logging in. And if that would be done, then there's nothing special in the UserAreaController. So I decided to now take only the first step as the benefit from that feels very concrete and clear.

All admin controllers now inherit AdminAreaContorller, so no need to repeat the filter that ensures Admin rights or the exception of not requiring email confirmation. And I do think here it fits to logic quite well to handle it as exception for the Admins rather than a requirement only for the user area.

Although it is true, that skipping filters can always be a bit of a surprise hidden in one layer of controllers.. :/

So feel free to comment. Anyway I think this is better than repeating the exception everywhere. And this takes a first (smaller) step in this refactoring direction and UserAreaController could be added later if it seems useful. Now it would be just one row of content there.

app/controllers/int_api/marketplaces_controller.rb
```diff
@@ -47,7 +47,14 @@ def create
...
     # make the marketplace creator land to the admin panel
-    url = "#{marketplace[:url]}/admin"
+    url_params = {host: marketplace[:url]}
+
+    #include port separately to url_params if it was present
+    if marketplace[:url] && marketplace[:url].split(':')[2]
+      url_params.merge!(port: marketplace[:url].split(':')[2])
+    end
+
+    url = admin_getting_started_guide_url(url_params)
```
ovan: This is better now with no string catenating and using path helper. Two small suggestions: no need to use mutable url_params here, the code feels overly complex with conditional mutation and Ruby's URI class is better first choice than string splitting whenever thinking about URI/URL parsing:

```ruby
base_url = URI.new(marketplace[:url])
url = admin_getting_started_guide_url(host: base_url.host, port: base_url.port)
```

app/controllers/admin/categories_controller.rb
ovan: Understood. Ok to take it one step at a time. Good point about AdminController and UserController names having a possibility for confusion. How about AdminPageController ?

ovan: Or perhaps AdminBaseController

app/controllers/int_api/marketplaces_controller.rb
```diff
@@ -51,7 +51,7 @@ def create
 
     #include port separately to url_params if it was present
     if marketplace[:url] && marketplace[:url].split(':')[2]
-      url_params.merge!(port: marketplace[:url].split(':')[2])
+      url_params[:port] = marketplace[:url].split(':')[2]
```
gnomet: @ovan thanks for tip! I also felt bit overly complicated at that point, but somehow didn't think that some library could give the port nicely even thought the URL was just a text variable at that point.

Somehow I notice still this tendency to like passing stuff as objects that can nicely tell about their contents, and when all is passed as simple data types I feel like my tool box is much emptier. But I should learn better the tools that do nice things with simple datatypes, as that seems to be the direction were heading.. :)

ovan: @gnomet I need to clarify this a bit so you don't end up with wrong mental model. There's no preference or need to use only language's primitive types (strings, numbers, etc). The marketplace_url in this case could as well be passed as URI object from service. We just haven't had the need for anything other than string before. URI is a canonical way to represent urls in Ruby and it's a perfectly fine value type (not immutable though but neither are strings). Another example is monetary values represented as Money objects. Those are also value types and there's no problem whatsoever using them where it makes sense. In fact, the discovery api even puts (abstract versions of) URI and Money onto wire with Transit. Those are then parsed to language native presentations at receive time (e.g. JS in browser parses them as different wrapped objects than what Ruby does). What we don't want to do is to use objects for composite data types, i.e. to use a special purpose object to wrap a collection related properties. Hash (map) does much better job there.

gnomet: @ovan Thanks! This was a good clarification! 
I didn't have the difference between value type objects and composite data type objects so clear in mind, so indeed there was some misunderstanding that I'm more aware with now. 👍

app/controllers/admin/categories_controller.rb
gnomet: @ovan yea, perhaps the AdminBaseController could be good one. Not pointing so directly to only pages or area, as the base controller could handle responsibilities also not directly related to certain pages or areas, but more just admin actions etc.. I'll switch to that.

PR #2699, opened 2016-10-31
Change marketplace currency from admin panel

kpuputti:
This PR enables marketplace admins to change the payment system currency to one of the currencies that PayPal supports. The currency is still stored in the `available_currencies` column of the `communities` table, but it is also stored to the new `minimum_price_currency` and `minimum_transaction_fee_currency` columns of the `payment_settings` table for the `minimum_price_cents` and `minimum_transaction_fee_cents` values.

Note that the minimum commission for a currency validation is removed from the frontend since it relied on converting money units into floating point numbers for JavaScript, which is not reliable anyways. The proper validation is still done in the backend.

## Screenshots

screen shot 2016-11-02 at 10 49 25

screen shot 2016-11-02 at 10 50 01

## TODO:

- [x] add dropdown to select currency in payment system preferences
- [x] populate dropdown with available currencies
- [x] remove instructions to contact support for changing currency
- [x] validate currency in POST params
- [x] change currency in model/DB
- [x] migration to add currency to payment_settings and copy the value from community
- [x] change labels in frontend when currency changed
- [x] verify available_currencies contents for communities with a payment_settings row in prod DB
- [x] frontend form validation when currency changes
- [x] copy text and translation changes
- [x] fix min price and min tx fee currency column types
- [x] fix support link
- [x] fix ss-icon usage


app/controllers/admin/paypal_preferences_controller.rb
```diff
@@ -48,7 +52,8 @@ def index
       minimum_commission: minimum_commission,
       commission_from_seller: tx_settings[:commission_from_seller],
       minimum_listing_price: Money.new(tx_settings[:minimum_price_cents], @current_community.default_currency),
-      minimum_transaction_fee: Money.new(tx_settings[:minimum_transaction_fee_cents], @current_community.default_currency)
+      minimum_transaction_fee: Money.new(tx_settings[:minimum_transaction_fee_cents], @current_community.default_currency),
```
rap1ds: Could we now use the `currency` here instead of `@current_community.default_currency`?

kpuputti: @rap1ds Do you mean `marketplace_currency`? Is that available in the scope?

app/controllers/admin/paypal_preferences_controller.rb
```diff
@@ -58,6 +63,9 @@ def index
       admin_getting_started_guide_path,
       Admin::OnboardingWizard.new(@current_community.id).setup_status)
 
+    currencies = MarketplaceService::AvailableCurrencies::CURRENCIES
+    available_currencies = [currency].concat(currencies.select { |c| c != currency })
```
rap1ds: In this case, we can already rely on the fact that the user is using PayPal, right? So, could we also rely that the current currency is in the list of available currencies, without adding it to the list?

app/services/marketplace_service/available_currencies.rb
```diff
@@ -67,4 +67,6 @@ module MarketplaceService::AvailableCurrencies
       "VA" => "EUR"
   }
 
+  CURRENCIES = ["USD"].concat(COUNTRY_CURRENCIES.values).uniq.sort
```
rap1ds: Feel that [SortedSet](https://ruby-doc.org/stdlib-1.9.3/libdoc/set/rdoc/SortedSet.html) could be the right data structure for this.

app/controllers/admin/paypal_preferences_controller.rb
kpuputti: That was originally there to put the current currency as the first value in the list, no matter what the sorting was. But I guess it doesn't make that much sense, especially since I now figured out how to define the selected value... :P

app/controllers/admin/paypal_preferences_controller.rb
rap1ds: Never mind. I just realized that `currency == @current_community.default_currency`.

db/migrate/20161101133602_add_currency_column_to_payment_settings.rb
```diff
@@ -0,0 +1,16 @@
+class AddCurrencyColumnToPaymentSettings < ActiveRecord::Migration
+  def up
+    add_column :payment_settings, :minimum_price_currency, :text, after: :minimum_price_cents
+    add_column :payment_settings, :minimum_transaction_fee_currency, :text, after: :minimum_transaction_fee_cents
+    sql = "UPDATE communities c, payment_settings s
+           SET s.minimum_price_currency = c.available_currencies,
+               s.minimum_transaction_fee_currency = c.available_currencies
```
kpuputti: Currently there seems to be one marketplace in the production DB where `available_currencies` holds a list of multiple currencies that would be migrated to the `payment_settings` table in this migration. Is that a problem?

app/assets/stylesheets/views/_account-settings.css.scss
```diff
@@ -245,3 +245,14 @@
 .paypal-last-steps-list {
   margin-bottom: lines(1.5);
 }
+
+.paypal-currency-information {
+  padding-left: 30px;
+  position: relative;
+
+  .ss-info {
```
rap1ds: This will not work if someone uses font-awesome. Maybe the icons could be wrapped in a `span` or maybe you could add an extra class and use that as a selector?

db/migrate/20161101133602_add_currency_column_to_payment_settings.rb
```diff
@@ -0,0 +1,16 @@
+class AddCurrencyColumnToPaymentSettings < ActiveRecord::Migration
+  def up
+    add_column :payment_settings, :minimum_price_currency, :text, after: :minimum_price_cents
```
rap1ds: The column type is now `text`, which is way too much for a single currency (65535 characters). You could use varchar here.

db/migrate/20161101133602_add_currency_column_to_payment_settings.rb
```diff
@@ -0,0 +1,16 @@
...
+    add_column :payment_settings, :minimum_transaction_fee_currency, :text, after: :minimum_transaction_fee_cents
```
rap1ds: And here.

app/views/admin/paypal_preferences/index.haml
```diff
@@ -60,13 +54,27 @@
...
                     =icon_tag("check", ["icon-fix", "paypal-success-mark"])
-                  =t("admin.paypal_accounts.set_minimum_price_and_fee")
+                  =t("admin.paypal_accounts.edit_payment_settings")
+            .row
+              .col-12
+                %p.paypal-currency-information
+                  =icon_tag("information")
+                  =t("admin.paypal_accounts.supported_currencies_information_text",
+                     contact_support_link: link_to(t("admin.paypal_accounts.contact_support_link_text"),
+                                                   "mailto:support@sharetribe.com")).html_safe
```
rap1ds: We shouldn't hard-code the support email here. How have we done these suppot links in other parts of the app? In addition to that, should we use the `data-uv-trigger` tag to use the UserVoice widget?

db/migrate/20161101133602_add_currency_column_to_payment_settings.rb
rap1ds: Yeah I think it is a problem. We should pick only one currency from that list.

app/assets/stylesheets/views/_account-settings.css.scss
kpuputti: @rap1ds There will be an empty space if font-awesome is in use. I don't see this as a problem?

screen shot 2016-11-08 at 14 57 47

kpuputti: Fixed now.

PR #2706, opened 2016-11-01
Save exclusive booking end date

rap1ds:
This PR adds a new column `bookings.end_on_exclusive`. When a new booking transaction is created, the `end_on_exclusive` information is saved to the database.

This PR touches the code very little! The main purpose of this PR is to save data in the new form where the end date is always exclusive. There will be follow-up Pull Requests, which will do more code changes, i.e. start reading the end value from `end_on_exclusive`.

### DIFF
--- a/app/models/booking.rb
+++ b/app/models/booking.rb
@@ -2,12 +2,13 @@
 #
 # Table name: bookings
 #
-#  id             :integer          not null, primary key
-#  transaction_id :integer
-#  start_on       :date
-#  end_on         :date
-#  created_at     :datetime         not null
-#  updated_at     :datetime         not null
+#  id               :integer          not null, primary key
+#  transaction_id   :integer
+#  start_on         :date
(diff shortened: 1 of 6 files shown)


app/models/booking.rb
```diff
@@ -18,8 +19,6 @@ class Booking < ActiveRecord::Base
 
   belongs_to :tx, class_name: "Transaction", foreign_key: "transaction_id"
 
-  attr_accessible :transaction_id, :end_on, :start_on
```
rap1ds: This is removed because `attr_accessible` is deprecated in favor of `strong_params`

ovan:
The code changes look ok, nothing to comment there. About the naming, is end_on_exclusive a temporary name  and the idea is to refactor that eventually back to being just end_on? I feel that would be nicer because end_on_exclusive doesn't match with any other names we use.

PR #2707, opened 2016-11-01
Design review changes and fixes

Gnito:
These are related to Trello task: *Style review - bugs and enhancements*

### DIFF
--- a/client/app/components/composites/ListingCardPanel/ListingCardPanel.css
+++ b/client/app/components/composites/ListingCardPanel/ListingCardPanel.css
@@ -1,7 +1,9 @@
 .container {
   display: flex;
   flex-direction: column;
-  padding: calc(var(--ListingCardPanel_gutterSpacing) / 2);
+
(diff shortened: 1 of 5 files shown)


client/app/components/composites/ListingCardPanel/ListingCardPanel.css
```diff
@@ -2,7 +2,7 @@
   display: flex;
   flex-wrap: wrap;
   justify-content: space-around;
-  padding: calc(var(--ListingCardPanel_gutterSpacing) / 2);
+  padding: calc(var(--ListingCardPanel_gutterSpacing) / 4);
```
sktoiva: Use either variable or directly 6px, maybe with comment to avoid magic numbers.

Gnito: Changed this to explicit padding (18px) - managed to do that by removing margins from listing cards in single column layout (mobile)

PR #2711, opened 2016-11-02
Always use exclusive end on

rap1ds:


app/mailers/transaction_mailer.rb
```diff
@@ -30,7 +30,7 @@ def transaction_preauthorized(transaction)
       payment_type = MarketplaceService::Community::Query.payment_type(@community.id)
       gateway_expires = MarketplaceService::Transaction::Entity.authorization_expiration_period(payment_type)
 
-      expires = Maybe(transaction).booking.end_on.map { |booking_end|
+      expires = Maybe(transaction).booking.end_on_exclusive.map { |booking_end|
```
ovan: I'm I correct that this code has actually been broken for day pickers until now? There's no check of unit type and the assumption seems to have already been end_on is exclusive.

app/services/transaction_service/process/preauthorize.rb
```diff
@@ -173,7 +170,7 @@ def initiate_booking(tx:)
           customerId: tx[:starter_uuid],
           initialStatus: :paid,
           start: tx[:booking][:start_on],
-          end: end_adjusted
+          end: tx[:booking][:end_on]
```
ovan: Should this be :end_on_exclusive?

app/services/transaction_service/store/transaction.rb
```diff
@@ -192,8 +192,10 @@ def add_opt_shipping_address(hash, m)
   def add_opt_booking(hash, m)
     if m.booking
       booking_data = EntityUtils.model_to_hash(m.booking)
-      hash.merge(booking: Booking.call(
-                  booking_data.merge(duration: m.listing_quantity)))
+      hash.merge(booking: Booking.call({
```
ovan: Indentation! :)

db/migrate/20161102101418_populate_exclusive_end_where_unit_type_is_null.rb
```diff
@@ -0,0 +1,11 @@
+class PopulateExclusiveEndWhereUnitTypeIsNull < ActiveRecord::Migration
+  def up
+    name = "Populate end_on_exclusive for bookings where unit type is NULL"
+    exec_update([
+                  "UPDATE bookings",
+                  "INNER JOIN transactions ON transactions.id = bookings.transaction_id",
+                  "SET bookings.end_on_exclusive = ADDDATE(bookings.end_on, 1)",
```
ovan: ADDDATE, TIL

app/mailers/transaction_mailer.rb
rap1ds: That's a tricky question. This will definitely change the behaviour: For day pickers, the expires date is now +1 day, if the `(booking end + 2 days)` < `gateway expiration`. It's quite tricky logic...

I'm not sure if I can say that it has been "broken" before. Why did you conclude that exclusive end_on is assumed here? It feels that this is not exact science due to the +2 days that is added by the `preauth_expires_at` method.

app/services/transaction_service/process/preauthorize.rb
rap1ds: Noup, the `:end_on_exclusive` is Transaction Store internal. The store returns a Booking entity with `:start_on`, `:end_on` and `:duration`

PR #2719, opened 2016-11-04
Remove availability flag

rap1ds:
This PR:

- Removes availability flag (i.e. enables it to everyone)
- Adds `harmony_api_in_use` environment variable
- Adds a flash notification when listing is edited and availability is enabled/disabled
- Adds a link to knowledge base article

### DIFF
--- a/config/config.defaults.yml
+++ b/config/config.defaults.yml
@@ -285,6 +285,7 @@ default: &default_settings
   external_search_offset_min: 0
 
   # Harmony service API connection
+  harmony_api_in_use: false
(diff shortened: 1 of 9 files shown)


config/config.defaults.yml
```diff
@@ -285,6 +285,7 @@ default: &default_settings
   external_search_offset_min: 0
 
   # Harmony service API connection
+  harmony_api_in_use: false
```
ovan: Why did you feel we need this boolean separately from simply checking if the api_url is set?

PR #2720, opened 2016-11-04
Add admin controls for toggling new searchpage

sktoiva:
Adds admin controls for searhpage. New searchpage requires topbar and
it is enabled automatically if searchpage is enabled. JS is used to
disable and enable topbar controls when searchpage is checked and
unchecked.

Adds also a rule for private marketplaces as these cannot have
searchpage if custom landing page isn't enabled.

## Screenshots

image

image

image

### DIFF
--- a/app/assets/javascripts/new_layout.js
+++ b/app/assets/javascripts/new_layout.js
@@ -4,10 +4,10 @@ window.ST = window.ST || {};
 
   function check_handler(element) {
    return function(){
-      if(!element.prop("checked")){
+      if (!element.prop("checked")){
         element.prop("checked", this.checked);
       }
-      element.attr("disabled", this.checked)
+     element.attr("disabled", this.checked);
     }
   };
 
--- a/app/assets/javascripts/new_layout.js
+++ b/app/assets/javascripts/new_layout.js
@@ -1,0 +1,23 @@
+window.ST = window.ST || {};
+
+(function(module) {
+
+  function check_handler(element) {
+   return function(){
+      if(!element.prop("checked")){
+        element.prop("checked", this.checked);
+      }
+      element.attr("disabled", this.checked)
+    }
+  };
+
+  module.initializeNewLayoutManager = function(){
+    var $topbar_user = $("#enabled_for_user_topbar_v1");
+    var $topbar_community = $("#enabled_for_community_topbar_v1");
+
+    var $user_searchpage = $("#enabled_for_user_searchpage_v1")
+        .click(check_handler($topbar_user));
+    var $community_searchpage = $("#enabled_for_community_searchpage_v1")
+        .click(check_handler($topbar_community));
+  };
+})(window.ST);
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/new_layout.js
```diff
@@ -0,0 +1,23 @@
...
+      }
+      element.attr("disabled", this.checked)
+    }
+  };
+
+  module.initializeNewLayoutManager = function(){
+    var $topbar_user = $("#enabled_for_user_topbar_v1");
+    var $topbar_community = $("#enabled_for_community_topbar_v1");
+
+    var $user_searchpage = $("#enabled_for_user_searchpage_v1")
```
lyyder: This var can be removed as it is not used.

app/assets/javascripts/new_layout.js
```diff
@@ -0,0 +1,23 @@
...
+        .click(check_handler($topbar_user));
+    var $community_searchpage = $("#enabled_for_community_searchpage_v1")
```
lyyder: Same here.

app/view_utils/new_layout_view_utils.rb
```diff
@@ -38,18 +57,46 @@ def features(community_id, person_id)
...
+  end
+
+
+  def all_features(person_flags, community_flags, private_community, clp_enabled)
+    all_flags = (person_flags | community_flags).to_a
+
+    # Additional rules for flags
+    all_flags.delete(:searchpage_v1) unless can_manage_searchpage?(all_flags, private_community, clp_enabled)
+
+    (FEATURES + EXPERIMENTAL_FEATURES.values_at(*all_flags)).compact
```
Gnito: Does this work? I assume that values_at should be applied to end-result of *FEATURES + EXPERIMENTAL_FEATURES* ? Now, I read this so that it's added to experimental_features only

Gnito: Never mind

app/assets/javascripts/new_layout.js
```diff
@@ -0,0 +1,27 @@
...
+        element.prop("checked", this.checked);
+      }
+      element.attr("disabled", this.checked)
+    }
+  };
+
+  // Disables and enables required checkboxes for parent flags
+  module.initializeNewLayoutManager = function(feature_rels){
+    Object.keys(feature_rels).forEach(function(key,index) {
+      var $parent_for_user = $("#enabled_for_user_" + key);
```
Gnito: I guess, this is OKish. However, I have a habit of adding className of parent container to params, so that these could be restricted to certain part of the page (you never know if there is more components with same name added in the future).

That way you could write the content in a vanilla js manner:
```javascript
var parent_for_user = document.querySelector(parentClass + " #enabled_for_user_" + key);
var required_for_user =  document.querySelector(parentClass + " #enabled_for_user_" + feature_rels[key]);
parent_for_user.addEventListener("click", check_handler(required_for_user));
```

PR #2721, opened 2016-11-04
Add error handling for discovery api errors

sktoiva:
Handles null bootsrapped data and renders empty page with proper flash
notification.

screen shot 2016-11-04 at 17 12 53

### DIFF
--- a/client/app/startup/SearchPageApp.js
+++ b/client/app/startup/SearchPageApp.js
@@ -79,13 +79,12 @@ export default (props) => {
     'sign_up',
   ], { locale });
 
-  const bootstrappedData = TransitImmutableConverter.fromJSON(props.searchPage.data);
+  const bootstrappedData = TransitImmutableConverter.fromJSON(_.get(props, 'searchPage.data', null));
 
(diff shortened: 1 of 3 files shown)


client/app/startup/SearchPageApp.js
```diff
@@ -79,17 +79,16 @@ export default (props) => {
     'sign_up',
   ], { locale });
 
-  const bootstrappedData = TransitImmutableConverter.fromJSON(props.searchPage.data);
+  const bootstrappedData = TransitImmutableConverter.fromJSON(_.get(props, 'searchPage.data', null));
 
-  const rawListings = bootstrappedData
-    .get(':data');
+  const rawListings = _.get(bootstrappedData, ':data', []);
```
Gnito: These are already converted to Immutable.js format and it would be better to handle gets etc. through immutable's own functions (those immutable objects are quite far away from normal JS objects and therefore lodash doesn't fully know what it's doing with those maps/lists etc.)

There is a default for Immutable.get too
https://facebook.github.io/immutable-js/docs/#/Map/get

PR #2724, opened 2016-11-07
Remove multicurrency support PHASE 2

kpuputti:
This PR is the second phase in removing the support for multiple currencies and builds on top of #2722 

The old currency value in `communities.available_currencies` is copied to the new `communities.currency` column and the value is read from the new column from now on.

### DIFF
--- a/app/models/community.rb
+++ b/app/models/community.rb
@@ -569,8 +569,8 @@ class Community < ActiveRecord::Base
   end
 
   def default_currency
-    if available_currencies
-      available_currencies.gsub(" ","").split(",").first
+    if currency
+      currency
     else
       MoneyRails.default_currency
     end
--- a/app/models/community.rb
+++ b/app/models/community.rb
@@ -49,6 +49,7 @@
 #  stylesheet_needs_recompile                 :boolean          default(FALSE)
 #  service_logo_style                         :string(255)      default("full-logo")
 #  available_currencies                       :text(65535)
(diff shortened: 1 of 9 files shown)


app/models/community.rb
```diff
@@ -568,8 +569,8 @@ def payments_in_use?
   end
 
   def default_currency
-    if available_currencies
-      available_currencies.gsub(" ","").split(",").first
+    if currency
```
rap1ds: When you are done with the migrations, I'd set the currency to NOT NULL in database and at the same time remove this `if` block

kpuputti: Done in phase 3 in #2726

app/views/listings/form/_currency_selector.haml
```diff
@@ -1,10 +1,6 @@
...
-    - if currencies.size > 1
-      = form.select :currency, currencies, { :selected => (@listing.currency || currencies.first) }
-    - else
-      = Money.new(1, currencies[0]).symbol
-      = form.hidden_field :currency, :value => currencies[0]
+  - if @current_community.currency
+    = Money.new(1, @current_community.currency).symbol
+    = form.hidden_field :currency, :value => @current_community.currency
   - else
     = form.select :currency, major_currencies(Money::Currency.table), { :selected => (@listing.currency || "EUR") }
```
kpuputti: Done in phase 3 in #2726

PR #2725, opened 2016-11-07
Banned user check for profile view

lyyder:
Adds a check that user is not banned when loading a user profile.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -22,6 +22,7 @@ class PeopleController < Devise::RegistrationsController
 
   def show
     @person = Person.find_by!(username: params[:username], community_id: @current_community.id)
+    binding.pry
     raise PersonDeleted if @person.deleted?
 
     redirect_to landing_page_path and return if @current_community.private? && !@current_user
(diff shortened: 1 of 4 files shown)


app/controllers/people_controller.rb
```diff
@@ -6,6 +6,7 @@ class PersonDeleted < StandardError; end
 
   before_filter EnsureCanAccessPerson.new(
     :id, error_message_key: "layouts.notifications.you_are_not_authorized_to_view_this_content"), only: [:update, :destroy]
+  #before_filter :ensure_person_not_banned, :only => [:show, :update]
```
rap1ds: This seems to be commented out. By accident?

rap1ds:
There seems to be already a quite handy way implemented for checking if the person is deleted. Probably, adding a new PersonBanned exception would've been the easiest way to implement this. Maybe you should consider doing it that way?

app/controllers/people_controller.rb
lyyder: Oh, nice catch!

lyyder:
@rap1ds Sure, I'll have a look.

rap1ds:
LGTM

PR #2726, opened 2016-11-07
Remove multicurrency support PHASE 3

kpuputti:
In the third and final phase of removing support for multiple currencies, we clean up the currency data in the DB and change the code to rely on a currency value for a community.

Builds on top of #2722 and #2724

### DIFF
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -715,7 +715,7 @@ class ListingsController < ApplicationController
   def commission(community, process)
     payment_type = MarketplaceService::Community::Query.payment_type(community.id)
     payment_settings = TransactionService::API::Api.settings.get_active(community_id: community.id).maybe
-    currency = community.default_currency
+    currency = community.currency
 
     case [payment_type, process]
     when matches([__, :none])
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -464,17 +464,20 @@ class ListingsController < ApplicationController
     update_successful = @listing.update_fields(listing_params)
 
     upsert_field_values!(@listing, params[:custom_fields])
+    finalise_update(@listing, shape, @current_community, update_successful, old_availability)
+  end
 
+  def finalise_update(listing, shape, community, update_successful, old_availability)
     if update_successful
-      @listing.location.update_attributes(params[:location]) if @listing.location
+      listing.location.update_attributes(params[:location]) if listing.location
       flash[:notice] = update_flash(old_availability: old_availability, new_availability: shape[:availability])
-      Delayed::Job.enqueue(ListingUpdatedJob.new(@listing.id, @current_community.id))
-      reprocess_missing_image_styles(@listing) if listing_reopened
-      redirect_to @listing
+      Delayed::Job.enqueue(ListingUpdatedJob.new(listing.id, community.id))
+      reprocess_missing_image_styles(listing) if listing.closed?
+      redirect_to listing
     else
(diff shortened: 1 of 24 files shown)


db/migrate/20161107112025_remove_available_currencies_column_from_communities.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveAvailableCurrenciesColumnFromCommunities < ActiveRecord::Migration
+  def up
+    remove_column :communities, :available_currencies
+  end
```
rap1ds: Rollback migration is missing.

But instead of adding `down` migration, you can just rename the `up` to `change` and add some information to `remove_column`:

```
def up
  remove_column :communities, :available_currencies, :string, limit: 3, after: ...
end
```

app/views/listings/form/_currency_selector.haml
```diff
@@ -1,10 +1,3 @@
...
-    - currencies = @current_community.available_currencies.gsub(" ","").split(",")
-    - if currencies.size > 1
-      = form.select :currency, currencies, { :selected => (@listing.currency || currencies.first) }
-    - else
-      = Money.new(1, currencies[0]).symbol
-      = form.hidden_field :currency, :value => currencies[0]
-  - else
-    = form.select :currency, major_currencies(Money::Currency.table), { :selected => (@listing.currency || "EUR") }
+  = Money.new(1, @current_community.currency).symbol
+  = form.hidden_field :currency, :value => @current_community.currency
```
rap1ds: Now that a marketplace has only one currency, we don't need to circulate that information through view layer. This hidden field could be removed, and the `listings_controller` could be changed so that it would read `@current_community.currency` instead of `params[:currency]`.

db/migrate/20161107113656_clean_up_community_currencies.rb
```diff
@@ -0,0 +1,11 @@
+class CleanUpCommunityCurrencies < ActiveRecord::Migration
+  def up
+    sql = 'UPDATE communities c
+           SET c.currency = "EUR"
+           WHERE c.currency IS NULL'
+    exec_update(sql, "Change NULL currencies to EUR", [])
+
+    change_column_default :communities, :currency, "EUR"
+    change_column_null :communities, :currency, false
```
rap1ds: These could be splitted to another migration, because these can be rolled back easily. In addition to that, maybe we shouldn't add the default value? Isn't that a programming error if we add a new community without currency? Maybe in that case it's just better to crash.

db/migrate/20161107112025_remove_available_currencies_column_from_communities.rb
kpuputti: I intentionally left the rollback migration out since the data is lost anyways. Does it still make sense to add the column back if it loses the original data?

rap1ds: You can (and probably should) add a down migration to `CopyCommunityCurrencyToCurrencyColumn`, where `SET c.available_currency = c.currency`

app/controllers/listings_controller.rb
```diff
@@ -873,7 +875,7 @@ def numeric_field_ids(custom_fields)
   end
 
   def normalize_price_params(listing_params)
-    currency = listing_params[:currency]
+    currency = @current_community.currency
```
rap1ds: You could make this method a bit more pure by reading the currency from `listing_params[:currency]`

rap1ds:
There's one annoying Rubocop issue :(

PR #2733, opened 2016-11-08
SessionContextStore

rap1ds:
This PR adds a `SessionContextStore`, which is a global object that holds information about the current user and the current marketplace. `SessionContextStore` can be used for example to pass authentication information to other services, such as Harmony.

List of changes:

- [X] Change JwtAutheticator ServiceClient middleware so that it takes a lambda and calls it if authentication information is not passed to it per request.
- [X] Add `SessionContextStore`. It holds the information about current user and current marketplace. Under the surface, the information is stored to `RequestStore`
- [X] Add `SessionContextMiddleware` Rack middleware, which will store the the information to `SessionContextStore` before each request
- [X] Add a module `SessionContextSerializer`. This module can be included to any Delayed Job, and it will automatically serialize/deserialize the `SessionContextStore` context.
- [X] Include `SessionContextSerializer` to every delayed job (where adding it makes sense)
- [X] Add a Delayed Job plugin, which cleans up `RequestStore` after each job run. This will also clear the `SessionContextStore` because it uses `RequestStore`
- [X] In `PayPalEvents` module, set the content of `SessionContextStore` manually. We need to do this, because the event may be triggered by IPN message or Scheduled task.
- [X] When retrying payment tokens, clear the `SessionContextStore` after each retry
- [X] Remove manually passed `auth_context` from some of the Harmony calls and let the `SessionContextStore` mechanism handle that

Todo:

- [ ] In the future, we can also remove the manually passed `auth_context` from all Harmony calls. However, we can't do it yet, since there are some jobs waiting to be executed, without the `SessionContext` being serialized.


app/jobs/send_welcome_email.rb
```diff
@@ -1,5 +1,6 @@
 class SendWelcomeEmail < Struct.new(:person_id, :community_id)
 
+  include SessionContextSerializer
```
ovan: Why did you want to include the mixin even for jobs that don't make external API calls?

ovan: Another question about this. Will deserializing a job work even if the job was originally serialized without session context when the code doing the deserialization has it?

app/services/harmony_client.rb
```diff
@@ -21,7 +21,16 @@
     ServiceClient::Middleware::Timing.new,
     ServiceClient::Middleware::BodyEncoder.new(:transit_msgpack),
     ServiceClient::Middleware::ParamEncoder.new,
-    ServiceClient::Middleware::JwtAuthenticator.new(APP_CONFIG.harmony_api_disable_authentication,
-                                                    APP_CONFIG.harmony_api_token_secret),
+    ServiceClient::Middleware::JwtAuthenticator.new(disable: APP_CONFIG.harmony_api_disable_authentication,
```
ovan: How about an enter before `disable: ...`, would keep the line length in control.

app/utils/service_client/middleware/jwt_authenticator.rb
```diff
@@ -1,21 +1,24 @@
...
+      def initialize(disable:, secret:, default_auth_context: IDENTITY)
+        @_disabled = disable
+        @_secret = secret
+        @_default_auth_context = default_auth_context
       end
 
       def enter(ctx)
         unless @_disabled
-          token = create_token(ctx[:opts][:auth_context])
+          token = create_token(ctx[:opts][:auth_context] || @_default_auth_context.call)
```
ovan: I think we should have at least log a warning in case authentication is disabled but neither the parameters nor the auth context builder give us one. This can happen if session context is relied on but is not in place due to e.g. forgetting to include serialization mixin in a job.

app/services/transaction_service/paypal_events.rb
```diff
@@ -50,34 +62,66 @@ def update_transaction_details(flow, details)
...
+        {
+          user_id: tx[:listing_author_id],
+          user_uuid: tx[:listing_author_uuid]
+        }
+      when :unknown
+        # Unknown user
+        {}
+      else
+        raise ArgumentError.new("Unknown transition actor: #{actor}")
+      end
```
ovan: How about adding a helper method to SessionContextStore similar to set_from_models that would build a proper session context from given parameters? That way the details of which keys are used would remain internal to store.

app/services/transaction_service/paypal_events.rb
```diff
@@ -50,34 +62,66 @@ def update_transaction_details(flow, details)
...
-    [:preauthorized_to_expired,     [:preauthorized, :expired]],
-    [:pending_ext_to_denied,        [:pending_ext, :denied]]
+    {name: :initiated_to_preauthorized,   actor: :starter, state: [:initiated, :pending, :authorization]},
+    {name: :initiated_to_voided,          actor: :starter, state: [:initiated, :voided]},
+    {name: :preauthorized_to_paid,        actor: :author,  state: [:preauthorized, :completed]},
+    {name: :preauthorized_to_pending_ext, actor: :author,  state: [:preauthorized, :pending]},
+    {name: :pending_ext_to_paid,          actor: :author,  state: [:pending_ext, :completed]},
+    {name: :preauthorized_to_voided,      actor: :author,  state: [:preauthorized, :voided]},
+    {name: :preauthorized_to_expired,     actor: :author,  state: [:preauthorized, :expired]},
+    {name: :pending_ext_to_denied,        actor: :author,  state: [:pending_ext, :denied]}
```
ovan: I didn't realize setting the context here would be so involved. Anyway, this looks like a good way to do it.

app/jobs/send_welcome_email.rb
rap1ds: First Q: My thinking was to include it to every job, so that if someone writes some code that depends on the SessionContextStore content (e.g. add a new API call), there's no need to think whether the SessionContextStore is initialized or not. Does this make sense?

Second Q: Yes it should. But let me test that once more.

app/utils/service_client/middleware/jwt_authenticator.rb
rap1ds: I guess you mean "in case authentication is **enabled**", right?

If that happens, the Entity call on [line 30](https://github.com/sharetribe/sharetribe/pull/2733/files/cacc9bcc594dd584de87964c73c720f93567b521#diff-37cd77e957ff88d796c4f822409c28e8R30) will throw an error because mandatory values are missing.

app/jobs/send_welcome_email.rb
ovan: Yeah, that's what I figured. I think I would have left it away by default simply to be explicit about where it is required (and have client middleware log a warning / raise when there's no auth context). The small risk with this approach is that it gets included without understanding what it is or does by those who come later. Either way works of course and don't feel particularly strongly about this.

ovan:
Yeah, this covers all the cases and in general looks very well thought out.

PR #2734, opened 2016-11-08
Change constants to functions

sktoiva:
Don't use function calls in constants

Fixes a bug  where new layout admin ui doesn't show correct translations as constants are shared across requests.

### DIFF
--- a/app/view_utils/new_layout_view_utils.rb
+++ b/app/view_utils/new_layout_view_utils.rb
@@ -16,21 +16,30 @@ module NewLayoutViewUtils
     searchpage_v1: :topbar_v1
   }
 
-  FEATURES = [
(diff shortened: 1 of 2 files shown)


app/view_utils/new_layout_view_utils.rb
```diff
@@ -16,21 +16,28 @@ module NewLayoutViewUtils
...
-      name:  :searchpage_v1,
-    },
-  }
+   ]
+  end
 
-  module_function
+  # Returns a hash describing experimental features Expermental
+  # features maps flag that is required for toggling the feature in
+  # admin ui to the feature
```
lyyder: Can you check this function documentation again.

PR #2736, opened 2016-11-08
Fix listing images parsing - they were records not arrays

Gnito:


Gnito:
After merge Hannu checked this

PR #2737, opened 2016-11-08
Icons update

Luis-RG:
This PR corrects some icons names, removes icons with incorrect name to standardise all the codes and names to the current version of IconJar


bladealslayer:
LGTM

PR #2738, opened 2016-11-09
ListingCard enable cmd + click to new tab

Gnito:
All different meta keys + click open the listing to a new tab.
Default behaviour is prevented (since some part of the card already contain ```a``` link wrapping)


Gnito:
Sampo checked

PR #2739, opened 2016-11-09
Delete _map_bubble_spacer.haml

1c7:
File first submit at 2014
and never changed after that.

and no other file refer this file.

I think this is a useless file.
should be remove


rap1ds:
Thanks!

PR #2741, opened 2016-11-09
Show warning when currency is about to change

kpuputti:
Changing the marketplace currency requires potential actions and might not be apparent to the admins. This PR adds a warning to the PayPal preferences page when the currency dropdown value changes.

screen shot 2016-11-09 at 15 57 33


rap1ds:
Didn't test this but the code LGTM.

PR #2742, opened 2016-11-09
Listingcard column ie11fix

Gnito:


Gnito:
Hannu checked this before leaving

PR #2745, opened 2016-11-10
Generic analytics events + send to amplitude

ovan:


app/views/layouts/_amplitude.erb
```diff
@@ -0,0 +1,42 @@
...
+   function v(e){function t(t){e[t]=function(){e._q.push([t].concat(Array.prototype.slice.call(arguments,0)));
+   }}for(var n=0;n<d.length;n++){t(d[n])}}v(n);n.getInstance=function(e){e=(!e||e.length===0?"$default_instance":e).toLowerCase();
+     if(!n._iq.hasOwnProperty(e)){n._iq[e]={_q:[]};v(n._iq[e])}return n._iq[e]};e.amplitude=n;
+ })(window,document);
+
+ var ampClient = amplitude.getInstance();
+ ampClient.init('<%= APP_CONFIG.amplitude_api_key %>');
+
+ <% if user %>
+   ampClient.setUserId('<%= user.uuid_object %>');
```
kpuputti: Should this be `user.uuid_object.to_s` to be more explicit?

app/views/layouts/_amplitude.erb
```diff
@@ -0,0 +1,42 @@
...
+   var userInfo = new amplitude.Identify()
+                               .set('community_id', <%= community.id %>)
+                               .set('marketplace_uuid', '<%= community.uuid_object %>')
```
kpuputti: As above, stringifying the UUID explicitly.

app/utils/analytics.rb
```diff
@@ -0,0 +1,34 @@
...
+#
+module Analytics
+
+  EVENT_KEY = :_analytics_event
+  LOGOUT_KEY = :_analytics_logout
+
+  module_function
+
+  def record_event(flash_or_now, event_name, props)
+    flash_or_now[EVENT_KEY] = {
```
rap1ds: Should we change this to array? What if we want to record multiple events?

app/views/layouts/_amplitude.erb
```diff
@@ -0,0 +1,42 @@
...
+   i.parentNode.insertBefore(r,i);function s(e,t){e.prototype[t]=function(){this._q.push([t].concat(Array.prototype.slice.call(arguments,0)));
+     return this}}var o=function(){this._q=[];return this};var a=["add","append","clearAll","prepend","set","setOnce","unset"];
+   for(var u=0;u<a.length;u++){s(o,a[u])}n.Identify=o;var c=function(){this._q=[];return this;
+   };var p=["setProductId","setQuantity","setPrice","setRevenueType","setEventProperties"];
+   for(var l=0;l<p.length;l++){s(c,p[l])}n.Revenue=c;var d=["init","logEvent","logRevenue","setUserId","setUserProperties","setOptOut","setVersionName","setDomain","setDeviceId","setGlobalUserProperties","identify","clearUserProperties","setGroup","logRevenueV2","regenerateDeviceId"];
+   function v(e){function t(t){e[t]=function(){e._q.push([t].concat(Array.prototype.slice.call(arguments,0)));
+   }}for(var n=0;n<d.length;n++){t(d[n])}}v(n);n.getInstance=function(e){e=(!e||e.length===0?"$default_instance":e).toLowerCase();
+     if(!n._iq.hasOwnProperty(e)){n._iq[e]={_q:[]};v(n._iq[e])}return n._iq[e]};e.amplitude=n;
+ })(window,document);
```
kpuputti: Is this the whole Amplitude script inlined here? How do we know which version it is if there's updates?

ovan: It's not the whole script, it creates a script tag. This is official way documented by amplitude so I think it's best to stick with that.

app/assets/javascripts/transaction.js
```diff
@@ -78,13 +78,24 @@ window.ST.transaction = window.ST.transaction || {};
       .flatMapLatest(function (data) { return Bacon.$.ajaxPost(formAction, data); })
       .flatMapLatest(toOpResult);
 
+    var analyticsEventSent = formSubmitWithData
+      .flatMapLatest(function() {
+        var timeout = Bacon.later(3000, "timeout");
+        var response = Bacon.fromCallback(function(callback) {
+          amplitude.getInstance().logEvent(analyticsEvent[0], analyticsEvent[1], callback);
```
ovan: If amplitude is not loaded (i.e. api key is not set) this will lead to null pointer. Maybe a JS abstraction for analytics that only calls through if amplitude is defined?

app/controllers/transactions_controller.rb
```diff
@@ -4,6 +4,21 @@ class TransactionsController < ApplicationController
     controller.ensure_logged_in t("layouts.notifications.you_must_log_in_to_view_your_inbox")
   end
 
+  before_filter only: [:new] do |controller|
+    unless controller.ensure_logged_in t("layouts.notifications.you_must_log_in_to_do_a_transaction")
+      fetch_data(params[:listing_id]).on_success do |listing_id, listing_model|
+        Analytics.record_event(
+          flash,
+          "Unlogged user tried to initiate a transaction",
```
ovan: The event names could be a little more like variable name like

PR #2753, opened 2016-11-14
Side push overlay

kpuputti:
This PR adds a generic side menu that pushes the page contents behind an overlay.

## Example

Latest UI example [In Styleguide](http://sharetribe.github.io/sharetribe/side-push-overlay/?selectedKind=General&selectedStory=SideWinder&full=0&down=1&left=1&panelRight=0&downPanel=storybook-addon-specifications%2Fspecifications-panel)

side-menu-no-outline

## TODO

- [x] Generic Side menu component in Styleguide
- [x] Close button + event
- [x] Fade overlay
- [x] Close menu when overlay clicked
- [x] Close menu with ESC
- [x] Clean up native DOM API usage
- [x] X icon SVG
- [x] Tune details based on the Sketch design
- [x] Prevent bg content scrolling
- [x] Extract vars from CSS
- [x] Add documentation
- [x] Git cleanup

## Review/bug fixes

- [x] Style the close button outline
- [x] Fix visible scrollbars when animating (requires a mouse attached)
- [x] Open side menu at the current scroll position instead of scrolling the page to the top
- [x] Fix scrollTop in FF
- [x] Adjust overlay height when window height changes

### DIFF
--- a/client/app/components/composites/SideWinder/SideWinder.css
+++ b/client/app/components/composites/SideWinder/SideWinder.css
@@ -1,0 +1,82 @@
+.wrapper {
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  right: 0;
+  transition: right 0.5s;
+}
+
+.winderOpen {
+  /* Prevent scrolling and hide scrollbars */
+  overflow: hidden;
+}
+
+.root {
+  display: none;
+  position: absolute;
+  top: 0;
+  right: 0;
+  height: 100%;
+
+  @nest .winderOpen & {
+    display: block;
+  }
+}
+
+.overlay {
+  position: absolute;
+  top: 0;
+  right: 0;
+  height: 0;
+  width: 100%;
+  background-color: rgb(44, 47, 50);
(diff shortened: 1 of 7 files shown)


client/app/components/composites/SideWinder/SideWinder.css
```diff
@@ -0,0 +1,78 @@
+.wrapper {
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  right: 0;
+  transition: right 0.5s;
+}
+
+.winderOpen {
+  overflow-y: hidden;
```
kpuputti: @rap1ds No, that is `overflow-y` on purpose. It prevents being able to scroll the page in the background vertically when the overlay is open.

rap1ds:
As discussed elsewhere, I see two issues there:

- [ ] Scrollbars appearing during the animation
- [ ] (If I understood correctly) page is always scrolled to the top. I think this is quite bad UX. Is there anything we can do about it?

PR #2759, opened 2016-11-20
Lightbox paragraphs had different font-size than list items

Gnito:
Hopefully this helps with that strange list item font-size bug.
(Who ever is able to review this - feel free to merge too.)


bladealslayer:
This seems to work. I'll merge.

PR #2765, opened 2016-11-24
ManageAvailabilityHeader component

kpuputti:
This PR adds the `ManageAvailabilityHeader` component that has the listing title and blurred background image (behind the marketplace color). The component will be used in the side menu of editing the listing availability.

new-header

### DIFF
--- a/client/app/components/composites/ManageAvailabilityHeader/ManageAvailabilityHeader.css
+++ b/client/app/components/composites/ManageAvailabilityHeader/ManageAvailabilityHeader.css
@@ -1,0 +1,34 @@
+.root {
+  position: relative;
+  overflow: hidden;
+}
+
+.backgroundLayer {
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  background-repeat: no-repeat;
+  background-size: cover;
+  background-position: center;
+  filter: blur(5px);
+}
+
+.listingHeader {
+  position: absolute;
+  bottom: 34px;
+  left: 34px;
+  right: 34px;
+  color: #fff;
+  font-size: 24px;
+  line-height: 1.4;
+}
+
+.availabilityHeader {
+  position: absolute;
+  top: -30px;
+  font-size: 10px;
+  text-transform: uppercase;
+  letter-spacing: 2.5px;
+}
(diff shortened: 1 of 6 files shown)


client/app/components/composites/ManageAvailabilityHeader/ManageAvailabilityHeader.css
```diff
@@ -0,0 +1,35 @@
...
+}
+
+.colorLayer {
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  opacity: 0.9;
+  filter: blur(5px);
```
rap1ds: I think the blur filter might be in the wrong layer. I guess it should blur the image, but now the image is sharp and clear but the color layer has some blurred white borders. See screenshot:

screen shot 2016-11-25 at 16 05 22

client/app/components/composites/ManageAvailabilityHeader/ManageAvailabilityHeader.css
```diff
@@ -0,0 +1,35 @@
+.root {
+  position: relative;
+  background-repeat: no-repeat;
+  background-size: cover;
+  background-position: center;
+  overflow: hidden;
+}
+
+.colorLayer {
```
rap1ds: FYI @jannekoivistoinen used a box shadow inset cool trick to add a color overlay without extra divs. We are using it already on the [success stories page hover](https://www.sharetribe.com/stories.html):

screen shot 2016-11-25 at 16 03 37

Would that we useful here?

kpuputti: Usually a pseudo element is a simpler and more expandable solution for these kinds of problems, but here I couldn't use one since the custom color has to be inlined from the JS side.

But the box shadow might work if I'm able to put the blur filter somewhere there. Thanks, I'll try it out.

client/app/components/composites/ManageAvailabilityHeader/ManageAvailabilityHeader.css
kpuputti: Yeah, I need to test this more.

kpuputti:
@rap1ds I had to use a separate div to apply the blur filter to the background image but not the text, but the box-shadow trick helped with the color. Now everything is updated and looks correct.

PR #2766, opened 2016-11-24
Implement Harmony Proxy

rap1ds:
This PR implements a `HarmonyProxyController`, which has two main responsibilities:

* Forward requests to Harmony service
* Authenticate and authorize the request

The path for the controller is `/harmony_proxy`. Request to `/harmony_proxy/path/to/somewhere` are forwarded to Harmony with path `/path/to/somewhere`. Query and body parameters are forwarded without changing them. Only whitelisted paths will be forwarded.

The proxy endpoint tries to mimic the real Harmony service as close as possible, but there are two exceptions:

* JWT token / authentication context is not needed
* Only JSON format is supported

Also, one change is made to the `BodyEncoder` middleware: The middleware takes now `encoding` options, which allows to override the default encoding. HarmonyProxy uses JSON, where as Transit is the default encoding.

### DIFF
--- a/app/controllers/harmony_proxy_controller.rb
+++ b/app/controllers/harmony_proxy_controller.rb
@@ -23,7 +23,7 @@ class HarmonyProxyController < ApplicationController
 
   EndpointDefinition = EntityUtils.define_builder(
     [:name, :symbol, :mandatory],
-    [:required_role, :mandatory, one_of: [:none, :user, :admin]],
+    [:login_needed, :bool, :mandatory],
     [:authorization, :callable, :mandatory])
 
   # Define here all the endpoints that you want to forward to Harmony
@@ -32,9 +32,7 @@ class HarmonyProxyController < ApplicationController
   #
   # - name: This is the endpoint name as a symbol. The name MUST match to one
   #         of the endpoints in the Harmony Client endpoint map.
-  # - required_role: The minimum role that is required to the action.
-  #                  Available roles are :none, :user and :admin, where
-  #                  :none < :user < :admin
+  # - login_needed: if `false`, this endpoint can be accessed by unlogged user
   # - authorization: A callable, which is called with two params, req and
   #                  auth_context. The function should return `true` if the user
   #                  is allowed to perform the given action, otherwise `false`
@@ -42,7 +40,7 @@ class HarmonyProxyController < ApplicationController
   ENDPOINTS = [
     {
       name: :show_bookable,
-      required_role: :user,
+      login_needed: true,
       authorization: AuthorizeShowBookable
     }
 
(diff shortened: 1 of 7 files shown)


app/controllers/harmony_proxy_controller.rb
```diff
@@ -0,0 +1,190 @@
...
+
+      raw_uuid = UUIDUtils.raw(UUIDTools::UUID.parse(q[:refId]))
+      listing = Listing.find_by(uuid: raw_uuid)
+
+      return false if listing.nil?
+
+      author_uuid = listing.author.uuid_object.to_s
+
+      auth_context[:marketplaceId] == q[:marketplaceId]
+      auth_context[:actorId] == author_uuid
```
kpuputti: Is there an `&&` missing to take both of these comparisons into account?

app/controllers/harmony_proxy_controller.rb
```diff
@@ -0,0 +1,190 @@
...
+  # The purpose of the controller is to forward all calls to it to
+  # Harmony service and authrize the requests before forwarding them.
+  #
+  # Errors (from the proxy):
+  #
+  # - 404: Endpoint not found
+  # - 401: Authentication needed, but user is not logged in
+  # - 403: Authorization failed
+  #
+  # If there are now errors from the proxy (auth passed and call forwarded),
```
kpuputti: s/now/no/

app/controllers/harmony_proxy_controller.rb
```diff
@@ -0,0 +1,190 @@
...
+  end
+
+  # Authenticate the user:
+  #
+  # - If login is needed, but user is not logged in, return 401
+  # - Else, create auth_context from @current_user
+  #
+  def authenticate(ctx)
+    if ctx[:endpoint][:login_needed] == true && @current_user.nil?
+      Result::Error.new("Unauhtorized", ctx.merge(error: { status: 401 }))
```
kpuputti: s/Unauhtorized/Unauthorized/

config/routes.rb
```diff
@@ -60,6 +60,10 @@
     post "/prospect_emails" => "marketplaces#create_prospect_email"
   end
 
+  # Harmony Proxy
+  # This endpoint proxies the requests to Harmony and does authorization
+  match '/harmony_proxy/*path' => 'harmony_proxy#proxy', via: :all
```
kpuputti: Rename path to harmony_path?

app/controllers/harmony_proxy_controller.rb
```diff
@@ -0,0 +1,190 @@
...
+  # The endpoint definition has contains following values:
+  #
+  # - name: This is the endpoint name as a symbol. The name MUST match to one
+  #         of the endpoints in the Harmony Client endpoint map.
+  # - login_needed: if `false`, this endpoint can be accessed by unlogged user
+  # - authorization: A callable, which is called with two params, req and
+  #                  auth_context. The function should return `true` if the user
+  #                  is allowed to perform the given action, otherwise `false`
+  #
+  ENDPOINTS = [
```
kpuputti: Rename this to convey that it's local endpoint that can be used?

PR #2768, opened 2016-11-28
Manage listing availability

kpuputti:


rap1ds:
Code looks good! 🎉

Two things that I noted while playing with the Storybook component:

1. Should there be hide animation for the sidebar when the close icon (the X) is clicked? Now the sidebar is hidden immediately.
2. This was a bit unexpected, steps:

  1. Open sidebar
  2. Click one day to block it
  3. Click the same day to unblock it

  Expected: Sidebar can be closed and there is no warning about unsaved changes. Save button is disabled/hidden

  Actual: Closing the sidebar shows warning about unsaved changed. Save button is visible.

kpuputti:
@rap1ds To the first point, you're right, is should be smoother and animated. I'm not sure if there's a regression or something, I'll check.

To the second point, the Styleguide wrapper component is just a bit dump and only marks that there are changes if an action is done. I don't see a point in making it more clever since the code is just for the Styleguide. The actual Redux code should work as you expect.

kpuputti:
@rap1ds Since the lack of close animation is about the SideWinder component and might require a somewhat large refactoring, I'll do it in another PR.

PR #2770, opened 2016-11-29
Refactor SideWinder component

kpuputti:
This PR refactors/rewrites the SideWinder component to use more generic utilities. The problem with the old component was the hacky native DOM manipulation and not being able to apply styles to all transition states. This resulted at least in not having closing animation working. The following approaches are taken in this PR:

 - Add Portal component to abstract rendering a node outside the current React render tree
 - Use [ReactTransitionGroup](https://facebook.github.io/react/docs/animation.html#reacttransitiongroup) for handle state changes in animations
 - Minimise/abstract native DOM handling

sidewinder-refactored

See interactive examples for a Portal and a SideWinder in the [Styleguide](http://sharetribe.github.io/sharetribe/sidewinder-refactor/?selectedKind=General&selectedStory=SideWinder&full=0&down=1&left=1&panelRight=0&downPanel=storybook-addon-specifications%2Fspecifications-panel).

### DIFF
--- a/client/app/components/composites/Portal/Portal.js
+++ b/client/app/components/composites/Portal/Portal.js
@@ -4,10 +4,8 @@
   Portal is a React component that is rendered outside the current
   render tree. It takes a parent element and children as props and
   creates a container element with the given children. This container
-  element is then rendered to the given parent element.
-
-  Only one Portal can be created for a single parent element at a
-  time. The children should be a single element.
+  element is then rendered to the given parent element. The children
+  should be a single element.
 
   ## Example:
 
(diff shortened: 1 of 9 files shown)


client/app/components/composites/Portal/Portal.js
```diff
@@ -0,0 +1,73 @@
...
+    if (!this.element) {
+      return;
+    }
+    ReactDOM.render(
+      this.props.children,
+      this.element
+    );
+  }
+  componentWillUnmount() {
+    if (!this.element) {
```
Gnito: Is this possible?

client/app/components/composites/Portal/Portal.js
```diff
@@ -0,0 +1,73 @@
+/*
+  # Portal
+
+  Portal is a React component that is rendered outside the current
+  render tree. It takes a parent element and children as props and
+  creates a container element with the given children. This container
+  element is then rendered to the given parent element.
+
+  Only one Portal can be created for a single parent element at a
```
rap1ds: Only one Portal per parent feels quite strict limitation

client/app/components/composites/Portal/Portal.js
kpuputti: I'll remove the restriction of allowing only a single Portal in a parent. That was only there for the parent CSS class, and that isn't used anywhere.

client/app/components/composites/Portal/Portal.js
kpuputti: Limitation removed.

client/app/components/composites/SideWinder/SideWinderTransition.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+  componentWillUnmount() {
+    document.body.classList.remove(css.transitionVisible);
+  }
+  componentWillAppear(callback) {
+    callback();
+  }
+  componentWillEnter(callback) {
+    this.setState({ entering: true, open: true });
+    window.setTimeout(callback, this.props.enterTimeout);
```
Gnito: As discussed, React life cycle could cause problems with uncleared timeouts (at least, I had lots of problems with FlashMessages related to these.)

PR #2778, opened 2016-12-01
Topbar mouseover hover

Gnito:
Due to [order of sent events](https://www.html5rocks.com/en/mobile/touchandmouse/#toc-together) hovering state is not useful knowledge - so, it is removed.

Current behaviour:
Scenario 1 - Desktop:
mouseOver -> open menu
mouseLeave - close menu

Scenario 2 - Desktop:
mouseOver -> open menu
click -> close menu
click -> open menu
click -> close menu
mouseLeave - menu is already closed

Scenario 3 - Touch device:
click -> open menu
click -> close menu

### DIFF
--- a/client/app/components/composites/AvatarDropdown/AvatarDropdown.js
+++ b/client/app/components/composites/AvatarDropdown/AvatarDropdown.js
@@ -3,27 +3,54 @@ import r, { div } from 'r-dom';
 import classNames from 'classnames';
 
 import { className } from '../../../utils/PropTypes';
-import { hasTouchEvents } from '../../../utils/featureDetection';
 import ProfileDropdown from './ProfileDropdown';
 import Avatar from '../../elements/Avatar/Avatar';
 import NotificationBadge from '../../elements/NotificationBadge/NotificationBadge';
 
 import css from './AvatarDropdown.css';
 
+const HOVER_TIMEOUT = 250;
+
 class AvatarDropdown extends Component {
   constructor(props, context) {
     super(props, context);
 
+    this.handleMouseOver = this.handleMouseOver.bind(this);
+    this.handleMouseOut = this.handleMouseOut.bind(this);
     this.handleClick = this.handleClick.bind(this);
     this.handleBlur = this.handleBlur.bind(this);
 
     this.state = {
       isOpen: false,
+      isHovering: false,
+      isMounted: false,
     };
+
+    this.mouseOverTimeout = null;
+    this.mouseOutTimeout = null;
+  }
+
+  componentDidMount() {
+    this.setState({ isMounted: true });// eslint-disable-line react/no-set-state
+  }
+
+  handleMouseOver() {
+    clearTimeout(this.mouseOutTimeout);
+    clearTimeout(this.mouseOverTimeout);
+    this.mouseOverTimeout = setTimeout(() => (
+      this.setState({ isHovering: true, isOpen: true }) // eslint-disable-line react/no-set-state
+      ), HOVER_TIMEOUT);
+  }
+
+  handleMouseOut() {
+    clearTimeout(this.mouseOverTimeout);
+    this.mouseOutTimeout = setTimeout(() => (
+      this.setState({ isHovering: false, isOpen: false }) // eslint-disable-line react/no-set-state
+      ), HOVER_TIMEOUT);
   }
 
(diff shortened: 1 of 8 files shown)


client/app/components/composites/AvatarDropdown/AvatarDropdown.js
```diff
@@ -3,27 +3,54 @@ import r, { div } from 'r-dom';
...
+    this.mouseOutTimeout = null;
+  }
+
+  componentDidMount() {
+    this.setState({ isMounted: true });// eslint-disable-line react/no-set-state
+  }
+
+  handleMouseOver() {
+    clearTimeout(this.mouseOutTimeout);
+    clearTimeout(this.mouseOverTimeout);
```
kpuputti: Let's prefix all these native global functions with `window`, ok? We should probably turn on the ESLint rule to complain about those as well.

client/app/components/composites/AvatarDropdown/AvatarDropdown.js
```diff
@@ -37,13 +64,16 @@ class AvatarDropdown extends Component {
...
+    const openOnHoverClass = this.state.isMounted ? '' : css.openOnHover;
+    const transitionDelayClass = this.state.isMounted ? '' : css.transitionDelay;
     const openClass = this.state.isOpen ? css.openDropdown : '';
     const notificationsClass = this.props.notificationCount > 0 ? css.hasNotifications : null;
     const notificationBadgeInArray = this.props.notificationCount > 0 ?
       [r(NotificationBadge, { className: css.notificationBadge }, this.props.notificationCount)] :
       [];
     return div({
+      onMouseOver: this.handleMouseOver,
+      onMouseLeave: this.handleMouseOut,
```
kpuputti: Maybe the method could be renamed to `handleMouseLeave` to indicate the mouse event it attaches to?

client/app/components/composites/AvatarDropdown/AvatarDropdown.js
```diff
@@ -37,13 +64,16 @@ class AvatarDropdown extends Component {
   }
```
kpuputti: The timeouts should be cleared also in `componentWillUnmount` to avoid changing the state after the component is unmounted.

client/app/components/composites/Menu/Menu.js
```diff
@@ -40,9 +47,24 @@ class Menu extends Component {
...
+      ), HOVER_TIMEOUT);
+  }
+
+  handleMouseout() {
+    clearTimeout(this.mouseOverTimout);
+    this.mouseOutTimout = setTimeout(() => (
+      this.setState({ isHovering: false, isOpen: false }) // eslint-disable-line react/no-set-state
+      ), HOVER_TIMEOUT);
+  }
+
```
kpuputti: Maybe no need to change at this point, but this uses the same hovering/timeout code that the other component above and could be abstracted into a HOC that wraps these components and provides a `hover` and `open` flags etc.

Gnito:
Needs still testing.

client/app/components/composites/Menu/Menu.js
```diff
@@ -40,9 +47,24 @@ class Menu extends Component {
     });
   }
 
+  handleMouseover() {
+    clearTimeout(this.mouseOutTimout);
+    clearTimeout(this.mouseOverTimout);
```
kpuputti: Prefix `clearTimeout` with `window` as above.

client/app/components/composites/Menu/Menu.js
```diff
@@ -21,17 +21,24 @@ class Menu extends Component {
...
     super(props, context);
 
+    this.handleMouseover = this.handleMouseover.bind(this);
+    this.handleMouseout = this.handleMouseout.bind(this);
     this.handleClick = this.handleClick.bind(this);
     this.handleBlur = this.handleBlur.bind(this);
     this.calculateDropdownPosition = this.calculateDropdownPosition.bind(this);
     this.state = {
       isOpen: false,
+      isHovering: false,
```
kpuputti: The `isHovering` handling should probably be removed from this component as well?

Gnito: Good point, I was just testing profile thing, since that's only testable thing on iOS world (our mobile layout is so wide).

PR #2782, opened 2016-12-05
Tech Stack Section(first section)

1c7:


1c7:
I think first section would be an appropriate place, what you think?

rap1ds:
Thanks!

1c7:
@rap1ds  You are welcome!

PR #2784, opened 2016-12-05
Listing availability management Redux store and actions

kpuputti:
This PR adds the Redux setup for the listing availability management UI.

The store structure copied from the code:

```js
const initialState = new Map({
  isOpen: true,
  visibleMonth: moment().startOf('month'),

  // List of days that buyers have reserved. These cannot be blocked.
  reservedDays: new List(),

  // List of days that are blocked and save to the API.
  blockedDays: new List(),

  // List of changes with `action` (String) and `day` (moment
  // instance) keys. Whenever the user blocks/allows a day, a change
  // is added to this list. The actual changes for the UI and the API
  // are calculated by going through this list to see the final values
  // for each day and comparing those to the blocked days above.
  changes: new List(),
});
```

Each change the user makes to the blocked days is collected to the `changes` list in the store. The derived state of the store or the changes to be sent to the API are then calculated from this change list. This also enables implementing undo functionality quite easily in the future, but that is not a requirement here.

This PR also adds the URL hash handling to toggle the store `isOpen` flag.

The rest of the actions are implemented in later PRs.


Gnito:
To continue a bit about what you were discussing with Mikko:
one way to keep record of single Day record is to see it as an undoable state.

Which is to say there's history of states
http://redux.js.org/docs/recipes/ImplementingUndoHistory.html#first-attempt-writing-a-reducer
There's also lots of ready made solutions (e.g. https://github.com/omnidan/redux-undo )

client/app/reducers/ManageAvailabilityReducer.js
```diff
@@ -1,5 +1,67 @@
...
+      state.get('daysToAllow').size > 0 ||
+      state.get('daysToBlock').size > 0;
+
+const sameDayAs = (day1) => (day2) => isSameDay(day1, day2);
+
+const withDay = (days, day) => days.push(day);
+
+const withoutDay = (days, day) => days.filterNot(sameDayAs(day));
+
+const allowDay = (state, day) =>
```
rap1ds: One idea would be to have one list for all block and allow actions, and then use a `reduce` function (`reduce` as in functional programming `reduce`, not necessarily as Redux `reducer`).

Example (in pseudo-code that resembles JavaScript but probably doesn't work):

```
events = [];
events.push({action: "block", value: 1}) // => [{action: "block", value: 1}]
events.push({action: "block", value: 2}) // => [{action: "block", value: 1}, {action: "block", value: 2}]
events.push({action: "allow", value: 1}) // => [{action: "block", value: 1}, {action: "block", value: 2}, {action: "allow", value: 1}]
```

Now, to calculate the changes, you could do something like:

```
changes = events.reduce(function(all_changes, event) {
  // Implementation omitted
  //
  // One wacky idea how to implement this: 
  // Map "blocks" to `-1`, "allows" to "1", then `sum` them and filter all zeros :)
}, [])

console.log(changes) // => [{action: "block", value: 2}]
```

To calculate the current view:

```
newReservedDays = blockedDays + changes
```

client/app/reducers/ManageAvailabilityReducer.js
```diff
@@ -1,5 +1,135 @@
...
+export const hasChanges = (state) =>
+  compressedChanges(state).size > 0;
+
+// Calculate currently blocked days from the fetched ones and the
+// unsaved changes.
+export const blockedDays = (state) => {
+  const changes = compressedChanges(state);
+
+  // Collect lists of day instances for allows and blocks
+  const splitChanges = changes.reduce((result, c) => {
```
rap1ds: This looks like `groupBy`. https://facebook.github.io/immutable-js/docs/#/List/groupBy

No need to change though.

client/app/reducers/ManageAvailabilityReducer.js
```diff
@@ -1,5 +1,135 @@
...
-  return state;
-}
+const initialState = new Map({
+  isOpen: true,
+  visibleMonth: moment().startOf('month'),
+
+  // List of days that buyers have reserved. These cannot be blocked.
+  reservedDays: new List(),
+
+  // List of days that are blocked and save to the API.
```
rap1ds: s/save/saved ?

client/app/reducers/ManageAvailabilityReducer.js
kpuputti: Ah, now I remember why I didn't use `groupBy`: you'd need another iteration to "pluck" only the days from the change instances.

PR #2796, opened 2016-12-09
ManageAvailability: render link for styleguide

Gnito:


client/app/components/sections/ManageAvailability/ManageAvailability.js
```diff
@@ -45,13 +45,17 @@ class ManageAvailability extends Component {
       this.setState({ renderCalendar: true }); // eslint-disable-line react/no-set-state
     }, CALENDAR_RENDERING_TIMEOUT);
 
-    document.getElementById(this.props.availability_link_id)
-      .addEventListener('click', this.clickHandler);
+    if (this.props.availability_link_id) {
+      document.getElementById(this.props.availability_link_id)
+        .addEventListener('click', this.clickHandler);
+    }
```
kpuputti: Could the DOM element be passed instead of the id?

That way the ManageAvailabilityApp component could do the DOM querying and this component could only attach the event listeners if the given element is truthy.

Gnito: I'll change

PR #2799, opened 2016-12-09
Adding FlashNotifications for ManageAvailability component

Gnito:
After this PR, you can send user alerts like
```
this.props.actions.addFlashNotification('error', 'Something bad happened!')
this.props.actions.addFlashNotification('warning', 'Something bad happened, but don't worry!')
this.props.actions.addFlashNotification('notice', 'New blocked dates have been saved!')
```
(In ManageAvailability.js)


lyyder:
LGTM!

PR #2800, opened 2016-12-09
Use Transit to communication to Harmony

rap1ds:
The original plan was to send JSON to Harmony and back. However, since we have to send `Date`s and `UUID`s, it makes a lot of sense to use Transit for that. This PR makes the required changes to use Transit.

- [X] Generalize Transit converter: There are some Listing specific types and handlers in the `utils/transitImmutableConverter`. Thus, it's needed to:
  - [X] Move generic types (`Money` and `Distance`) to `types/types.js`
  - [X] Move listing specific types and handlers away from `utils/transitImmutableConverter`
  - [X] Don't create a singleton converter. Instead, expose a constructor which lets others to instantiate new converters with a little bit different set of handlers.
- [X] Change Harmony Proxy so that it doesn't necessarily encode/decode request/response
- [X] Use `UUID` type for props
- [X] Use Immutable JS when ever it makes sense

TODO

- [x] Test new Search page layout (I don't have local setup)

### DIFF
--- a/client/app/reducers/ManageAvailabilityReducer.js
+++ b/client/app/reducers/ManageAvailabilityReducer.js
@@ -1,9 +1,8 @@
 /* eslint-disable no-alert */
-import { Map, List } from 'immutable';
+import { Map, List, Range } from 'immutable';
 import moment from 'moment';
 import { isSameDay } from 'react-dates';
 import * as actionTypes from '../constants/ManageAvailabilityConstants';
-import _ from 'lodash';
 
 const initialState = new Map({
   isOpen: true,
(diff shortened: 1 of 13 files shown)


client/app/reducers/ManageAvailabilityReducer.js
```diff
@@ -45,25 +44,26 @@ const withChange = (state, action, day) => {
...
 
 const mergeNovelty = (state, novelty) => {
-  const blocks = splitRanges(ranges(novelty.blocks));
-  const bookings = splitRanges(ranges(novelty.bookings));
+  const blocks = splitRanges(ranges(novelty.get('blocks')));
+  const bookings = splitRanges(ranges(novelty.get('bookings')));
 
   return state.set('reservedDays', state.get('reservedDays').concat(bookings))
-       .set('blockedDays', state.get('blockedDays').concat(blocks));
+              .set('blockedDays', state.get('blockedDays').concat(blocks));
```
kpuputti: Maybe this PR is the  wrong place to fix this, but just commenting here that only adding the new bookings and blocks doesn't result in correct data. It should also remove existing bookings and blocks that are not relevant anymore.

PR #2802, opened 2016-12-12
Add -worker suffix to the worker NEW_RELIC_APP_NAME

bladealslayer:


script/startup.sh
```diff
@@ -25,6 +25,8 @@ case "$APP_MODE" in
             # Do nothing
             exec sleep 86400
         else
+            # add -worker suffix to newrelic app name to separate
```
lyyder: Separate what from what?

lyyder:
Didn't understand the comment, otherwise 👍

script/startup.sh
bladealslayer: Thanks, will fix!

bladealslayer:
I think this is not really needed. Closing.

PR #2804, opened 2016-12-12
Improve data loading

rap1ds:
This PR improves the data fetching. When the user opens the sidebar, we'll load the data for the initial data for the sidebar + +/-2 months for additional preloaded data. When user changes a month, we load more data so that we keep the +/-2 months of data prefetched. In addition, a book keeping is implemented to remember which months have been loaded already. A month that has been loaded will never be reloaded.

TODO:

- [X] Load initial data when sidebar is opened.
- [X] Include `start` and `end` parameters to the fetch request
- [X] Preload +/-2 months
- [X] Keep track of loaded months and do not reload already loaded months.
- [X] Load data only when the edit listing sidebar is opened.

### DIFF
--- a/client/app/actions/ManageAvailabilityActions.js
+++ b/client/app/actions/ManageAvailabilityActions.js
@@ -1,6 +1,6 @@
 import * as actionTypes from '../constants/ManageAvailabilityConstants';
 import * as harmony from '../services/harmony';
-import { Map, List } from 'immutable';
+import Immutable from 'immutable';
 import { expandRange } from '../utils/moment';
 
 export const allowDay = (day) => ({
@@ -78,9 +78,9 @@ export const changeMonth = (day) =>
       }).then((response) => {
         const groups = response.get(':included').groupBy((v) => v.get(':type'));
 
-        const slots = new Map({
-          blocks: groups.get(':block', new List()),
-          bookings: groups.get(':booking', new List()),
+        const slots = Immutable.Map({
+          blocks: groups.get(':block', Immutable.List()),
+          bookings: groups.get(':booking', Immutable.List()),
         });
 
         dispatch(dataLoaded(slots, expandRange(start, end, 'months').toSet()));
--- a/client/app/actions/ManageAvailabilityActions.js
+++ b/client/app/actions/ManageAvailabilityActions.js
@@ -50,11 +50,21 @@ const loadRange = (date, preloadMonths) => {
 
 /**
    Given a list of `months` and a Set of already `loadedMonths`,
-   return an object { start, end } where already loaded months is removed.
+   return an object { start, end } where months that are already
+   loaded are removed from the beginning and the end of the range.
+
+   Example:
+
+   start: 4
+   end: 11
+   loadedMonths: [2, 3, 4, 5, 6, 9, 11, 12]
+
+   result: { start: 7, end: 10 }
+
  */
 const removeLoadedMonths = (months, loadedMonths) => ({
   start: months.find((s) => !loadedMonths.includes(s)),
-  end: months.reverse().find((e) => !loadedMonths.includes(e)),
+  end: months.findLast((e) => !loadedMonths.includes(e)),
 });
 
 const monthsToLoad = (day, loadedMonths, preloadMonths) =>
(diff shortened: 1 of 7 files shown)


client/app/actions/ManageAvailabilityActions.js
```diff
@@ -17,31 +18,74 @@ const changeVisibleMonth = (day) => ({
...
+  return expandRange(start, end, 'months');
+};
+
+/**
+   Given a list of `months` and a Set of already `loadedMonths`,
+   return an object { start, end } where already loaded months is removed.
+ */
+const removeLoadedMonths = (months, loadedMonths) => ({
+  start: months.find((s) => !loadedMonths.includes(s)),
+  end: months.reverse().find((e) => !loadedMonths.includes(e)),
```
kpuputti: No need to `reverse()`, you can use `findLast()`:

https://facebook.github.io/immutable-js/docs/#/List/findLast

client/app/actions/ManageAvailabilityActions.js
```diff
@@ -17,31 +18,74 @@ const changeVisibleMonth = (day) => ({
...
+  const start = startOfMonth.clone().subtract(preloadMonths, 'months');
+  const end = endOfMonthExclusive.clone().add(preloadMonths, 'months');
+
+  return expandRange(start, end, 'months');
+};
+
+/**
+   Given a list of `months` and a Set of already `loadedMonths`,
+   return an object { start, end } where already loaded months is removed.
+ */
```
kpuputti: This comment could be cleared. The code removes already loaded months from the start and end of the range, but there can be some loaded months in the middle.

client/app/actions/ManageAvailabilityActions.js
```diff
@@ -17,31 +18,74 @@ const changeVisibleMonth = (day) => ({
...
+        end: end.toJSON(),
+      }).then((response) => {
+        const groups = response.get(':included').groupBy((v) => v.get(':type'));
 
-      dispatch(dataLoaded(slots));
-    });
+        const slots = Immutable.Map({
+          blocks: groups.get(':block', Immutable.List()),
+          bookings: groups.get(':booking', Immutable.List()),
+        });
```
kpuputti: Should we write down a convention of always prefixing Immutable data structures with the namespace **and** not using `new` as you do here?

client/app/components/sections/ManageAvailability/ManageAvailability.js
```diff
@@ -25,6 +25,32 @@ SaveButton.propTypes = {
...
+*/
+const shouldLoad = (isOpen, prevIsOpen) => {
+  if (!isOpen) {
+    return false;
+  } else if (prevIsOpen && isOpen) {
+    return false;
+  } else {
+    return true;
+  }
+};
```
kpuputti: Could this whole function be simplified to `return isOpen && !prevIsOpen;`?

client/app/utils/moment.js
```diff
@@ -0,0 +1,29 @@
...
+
+   Examples:
+
+   expandRange(moment("2016-12-12"), moment("2016-12-14"), "days")
+    -> [moment("2016-12-12"), moment("2016-12-13")]
+
+   expandRange(moment("2016-12-01"), moment("2017-02-01"), "days")
+    -> [moment("2016-12-01"), moment("2017-01-01")]
+
+   expandRange(moment("2016-12-01"), moment("2017-02-15"), "days")
```
kpuputti: `s/days/months/` ?

client/app/utils/moment.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+   Takes a time range (`start` and `end` (exclusive)) and returns an Immutable List
+   containing the range expanded by the given duration.
+
+   Examples:
+
+   expandRange(moment("2016-12-12"), moment("2016-12-14"), "days")
+    -> [moment("2016-12-12"), moment("2016-12-13")]
+
+   expandRange(moment("2016-12-01"), moment("2017-02-01"), "days")
```
kpuputti: `s/days/months/` ?

client/app/actions/ManageAvailabilityActions.js
rap1ds: That's what was discussed in the chat. We also seem to have an exception set for `Immutable.List` etc. that they don't need to have the `new`: https://github.com/sharetribe/sharetribe/blob/master/client/.eslintrc.js#L161

PR #2815, opened 2016-12-14
Update keyserver

thii:


bladealslayer:
@thii did you have issues with the high-availability pool? This code is taken from the official Docker image for node.

thii:
Can you build the container with high-availability pool? I got a `keyserver receive failed` error.

[Node.js's main repo](https://github.com/nodejs/node) instructs using this keyserver.

bladealslayer:
Yes, the ha.pool works for me right now.

The node-related section in our Dockerfile is based on this: https://github.com/nodejs/docker-node/blob/6948057bbd9cc1469ca0e5e64d3bd5f000d4dc97/6.9/Dockerfile

By the way, the Docker support is in alpha stage now. For example, it doesn't support sphinx search, so keep that in mind if you are planning to use images built with our Dockerfile.

PR #2817, opened 2016-12-14
Fix location.hash bug by using pushState

Gnito:
And move it to component (inside life cycle process)

### DIFF
--- a/client/app/components/sections/ManageAvailability/ManageAvailability.js
+++ b/client/app/components/sections/ManageAvailability/ManageAvailability.js
@@ -122,7 +122,7 @@ class ManageAvailability extends Component {
     const maybeRenderDefaultLink = this.props.availability_link ? null : defaultLink;
 
     const winder = {
-      wrapper: document.querySelector('#sidewinder-wrapper'),
+      wrapper: this.props.sideWinderWrapper,
       maxWidth: cssVariables['--ManageAvailability_maxWidth'],
       minWidth: cssVariables['--ManageAvailability_minWidth'],
       isOpen: this.props.isOpen,
@@ -133,6 +133,9 @@ class ManageAvailability extends Component {
 
         if (!this.props.hasChanges || window.confirm(text)) { // eslint-disable-line no-alert
           this.props.actions.closeEditView();
+          if (typeof this.props.onCloseCallback === 'function') {
+            this.props.onCloseCallback();
+          }
         }
       },
     };
@@ -173,9 +176,11 @@ ManageAvailability.propTypes = {
   saveInProgress: PropTypes.bool.isRequired,
   onOpen: PropTypes.func.isRequired,
   onSave: PropTypes.func.isRequired,
+  onCloseCallback: PropTypes.func,
   isOpen: PropTypes.bool.isRequired,
   header: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
   calendar: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
+  sideWinderWrapper: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
 };
 
 export default ManageAvailability;
--- a/client/app/components/sections/ManageAvailability/ManageAvailability.js
+++ b/client/app/components/sections/ManageAvailability/ManageAvailability.js
@@ -3,10 +3,12 @@ import r, { button, div, a } from 'r-dom';
 import Immutable from 'immutable';
 import classNames from 'classnames';
 import { t } from '../../../utils/i18n';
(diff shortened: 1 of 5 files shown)


client/app/components/sections/ManageAvailability/ManageAvailability.js
```diff
@@ -70,6 +81,15 @@ class ManageAvailability extends Component {
     loadInitialDataIfNeeded(this.props);
   }
 
+  componentWillUpdate(nextProps) {
+    // manage location hash
+    if (nextProps.isOpen && window.location.hash !== '#manage-availability') {
+      setPushState(null, 'Availability calendar is open', '#manage-availability');
+    } else if (!nextProps.isOpen && window.location.hash === '#manage-availability') {
+      setPushState(null, 'Availability calendar is closed', '#');
```
rap1ds: Can we remove the anchor altogether?

client/app/components/sections/ManageAvailability/ManageAvailability.js
```diff
@@ -70,6 +81,15 @@ class ManageAvailability extends Component {
     loadInitialDataIfNeeded(this.props);
   }
 
+  componentWillUpdate(nextProps) {
+    // manage location hash
+    if (nextProps.isOpen && window.location.hash !== '#manage-availability') {
```
rap1ds: Will the `componentWillUpdate` function be run in the server rendering mode? If yes, then I think the `window.location.hash` will throw a null pointer. Should this be wrapped in `canUseDom`?

client/app/components/sections/ManageAvailability/ManageAvailability.js
```diff
@@ -37,14 +39,23 @@ const shouldLoad = (isOpen, prevIsOpen) => isOpen && !prevIsOpen;
...
-  const prevIsOpen = prevProps && prevProps.winder.isOpen;
+  const isOpen = props.isOpen;
+  const prevIsOpen = prevProps && prevProps.isOpen;
 
   if (shouldLoad(isOpen, prevIsOpen)) {
     props.calendar.onMonthChanged(props.calendar.initialMonth);
   }
 };
 
+const setPushState = (state, title, path) => {
```
rap1ds: As mentioned in the long Flowdock discussion, I think there might be some better places to do `window.location` manipulation, e.g.

1. A routing library
1. A non-React module that subscribes to the Store
1. A separate React component that doesn't render anything but only manages the `window.location`.

The problem with `window.location` manipulation in a component such as this is that the `window.location` is a global thing. And when component is manipulating a global thing, it's messing around with stuff that is not inside the component boundaries. What happens if two (or three) components are messing around with the `window.location`? Clashes are very possible. In addition, what happens when this component in rendered in some other context, e.g. in story book? Will it mess the story book URL?

So I'm a bit concerned about doing the hash manipulation in this component. It might be an okeyish solution as "well this time only and when there will be next time, then well refactor", but the story book part is still concerning.

client/app/components/sections/ManageAvailability/ManageAvailability.js
kpuputti: I don't think the `componentWillUpdate` lifecycle method will be run in the server. That method is called when the component is rendered with new props, and the server only renders the component once.

FYI, we are not server rendering this app at all, but that shouldn't affect how we do the components within this app.

client/app/components/sections/ManageAvailability/ManageAvailability.js
kpuputti: No hash vs. empty hash (# in the URL) would be nice, if possible.

client/app/components/sections/ManageAvailability/ManageAvailability.js
kpuputti: If the Storybook works, it should be fine. The component is rendered in an iframe, so the url it manipulates would be the inner frame url and doesn't mess with the URL of the Storybook itself.

kpuputti: My comment above was only an answer to the concern about Storybook.

client/app/components/sections/ManageAvailability/ManageAvailability.js
Gnito: It can't be empty, but it definitely can be more explicit.

client/app/components/sections/ManageAvailability/ManageAvailability.js
Gnito: I'd go with React-Router or similar HoC solution if we need more advanced route manipulation in the future.
Store subscribe is a valid alternative as a short term solution.

PR #2818, opened 2016-12-15
Manage availaiblity hover states

rap1ds:
Manage availability hover states.

- [X] Add "Close" help text next to the cross icon
- [X] Add day hovers
- [X] Add save button hover

Screenshots:

### DIFF
--- a/client/app/components/sections/ManageAvailability/ManageAvailability.css
+++ b/client/app/components/sections/ManageAvailability/ManageAvailability.css
@@ -21,10 +21,6 @@
 
   & button {
     margin: 0;
-
-    &:hover {
-      background-color: var(--colorReservedAvailability);
-    }
   }
 
   /* --- OVERRIDE DEFAULT STYLES END --- */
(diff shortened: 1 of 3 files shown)


client/app/components/composites/SideWinder/SideWinder.js
```diff
@@ -155,6 +156,7 @@ class SideWinder extends Component {
                 className: css.closeButton,
                 dangerouslySetInnerHTML: { __html: closeIcon },
               }),
+              span({ className: css.closeHelpText }, t('web.sidewinder.close')),
```
kpuputti: We had an explicit conversation with @jannekoivistoinen earlier about not doing this. Sorry if you didn't know about this.

Did you test this on Mobile Safari? If you make a visual change when hovering, you need two taps for the click action to go through.

client/app/components/sections/ManageAvailability/ManageAvailability.css
```diff
@@ -54,6 +50,13 @@
   &[disabled]:hover {
     background-color: #eee;
   }
+
+  will-change: background-color;
```
kpuputti: Again, only add the `will-change` if you really need it and notice that it makes a difference.

PR #2834, opened 2016-12-19
Don't clear changes before sidewinder is closed.

Gnito:
If this would be a SPA, save action should update *blocks* list after ajax call. 
(But now that improvement is postponed.)

### DIFF
--- a/client/app/reducers/ManageAvailabilityReducer.js
+++ b/client/app/reducers/ManageAvailabilityReducer.js
@@ -198,7 +198,7 @@ const manageAvailabilityReducer = (state = initialState, action) => {
     case actionTypes.START_SAVING:
       return state.set('saveInProgress', true);
     case actionTypes.CHANGES_SAVED:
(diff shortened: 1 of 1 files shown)


rap1ds:
Looks good to me, but maybe @kpuputti could have a quick look and say if this affects somehow the calendar data handling or something.

client/app/reducers/ManageAvailabilityReducer.js
```diff
@@ -200,7 +200,7 @@ const manageAvailabilityReducer = (state = initialState, action) => {
     case actionTypes.START_SAVING:
       return state.set('saveInProgress', true);
     case actionTypes.CHANGES_SAVED:
-      return state.set('saveInProgress', false).set('changes', Immutable.List());
+      return state.set('saveInProgress', false);
```
kpuputti: This is ok now that we close (and thus clear the store) the view after the save is finished. If we change the saving not to close the view at some point, we need to be careful here.

PR #2835, opened 2016-12-19
Don't handle pointer events in invisible month

Gnito:
I made bug report to React-Dates


rap1ds:
Looks good to me. Could you add a link to the Github Issue to the comment, so that it's easy to track when the issue has been solved in upstream and when this change can be removed from our source code.

rap1ds:
https://github.com/airbnb/react-dates/issues/224

PR #2838, opened 2016-12-20
Calendar timezone fixes

rap1ds:
This PR fixes the timezone issues in calendar. The main two issues were:

1. Wrong day is shown as the current day, if the system timezone is such that the day of month is different than in UTC timezone
1. Wrong day gets blocked, if the system timezone is such that the day of month is different than in UTC timezone

Harmony server is in UTC and it expects days represented as midnight date in UTC. User's browser is in users local timezone and the calendar expects days in local timezone UTC.

Because of this, two main helper functions were added.

`fromMidnightUTCDate(<Tue, 20 Dec 2016 00:00:00 GMT>) => <Tue Dec 20 2016 00:00:00 GMT+0200>`

`toMidnightUTCDate(<Tue, 20 Dec 2016 00:00:00 GMT+0200>) => <Tue Dec 20 2016 00:00:00 GMT>`

All the dates coming from server, we run through `fromMidnightUTCDate`. All the dates we are about to send to server, we run through `toMidnightUTCDate`

In addition addition to that, two changes have been made to the `harmony` data layer:

- Added `showBookable` method and removed the generic `get` method.
- Changed the data layer so that it expects native JS Date objects, instead of `moment` instances.

**Steps to reproduce the bugs:** See the Trello cards.

### DIFF
--- a/client/app/startup/ManageAvailabilityApp.js
+++ b/client/app/startup/ManageAvailabilityApp.js
@@ -24,7 +24,6 @@ export default (props) => {
     manageAvailability: Immutable.Map({
       isOpen: window.location.hash.replace(/^#/, '') === EDIT_VIEW_OPEN_HASH,
       visibleMonth: moment()
-        .utc()
         .startOf('month'),
       bookings: Immutable.List(),
       blocks: Immutable.List(),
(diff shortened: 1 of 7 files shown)


client/app/startup/ManageAvailabilityApp.js
```diff
@@ -24,7 +24,6 @@ export default (props) => {
     manageAvailability: Immutable.Map({
       isOpen: window.location.hash.replace(/^#/, '') === EDIT_VIEW_OPEN_HASH,
       visibleMonth: moment()
-        .utc()
```
kpuputti: If we shouldn't use `.utc()`, that should probably be removed from the two places it's used in `ManageAvailabilityReducer.js` as well?

PR #2839, opened 2016-12-20
Back to 100vh

Gnito:
Browser bars are a problem with flexbox layout. They might hide save button. This PR changes sideWinder height so that it will initially take 100vh and then (when JS kicks in) window.innerHeight.

### DIFF
--- a/client/app/components/sections/ManageAvailability/ManageAvailability.js
+++ b/client/app/components/sections/ManageAvailability/ManageAvailability.js
@@ -79,9 +79,13 @@ const setPushState = (state, title, path) => {
 class ManageAvailability extends Component {
   constructor(props) {
     super(props);
-    this.state = { renderCalendar: false };
+    this.state = {
+      renderCalendar: false,
+      viewportHeight: null,
+    };
 
     this.clickHandler = this.clickHandler.bind(this);
+    this.resizeHandler = this.resizeHandler.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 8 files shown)


Gnito:
I'm not sure if this is good enough, but if not - we need to figure out some reliable way to detect those browser bars.

kpuputti:
@Gnito I don't think there's a reliable way to detect the browser bars. However, `window.innerWidth` should be correct and Mobile Safari sends a `resize` event to `window` when the bars become visible. So we could dynamically adjust the height with inline styles.

This is exactly what I had done earlier, but the problem was passing the new height in the component children so that the child components would be rendered again.

client/app/components/sections/ManageAvailability/ManageAvailability.js
```diff
@@ -90,7 +94,7 @@ class ManageAvailability extends Component {
     // asynchronously allows the calendar to calculate the height
     // properly.
     // See: https://github.com/airbnb/react-dates/issues/46
-    window.setTimeout(() => {
+    this.calendarTimeout = window.setTimeout(() => {
```
kpuputti: Good catch to clear this timeout!

client/app/components/sections/ManageAvailability/ManageAvailability.js
```diff
@@ -99,6 +103,9 @@ class ManageAvailability extends Component {
     }
 
     loadInitialDataIfNeeded(this.props);
+
+    this.setState({ viewportHeight: window.innerHeight }); // eslint-disable-line react/no-set-state
+    window.addEventListener('resize', this.resizeHandler, false);
```
kpuputti: No need to add the capture flag, it defaults to `false`.

PR #2840, opened 2016-12-22
Show loading spinner while availability save is in progress

kpuputti:
This PR adds a loading SVG animation to the save button while the saving is in progress.

loading-done-checkmark


client/app/actions/ManageAvailabilityActions.js
```diff
@@ -6,6 +6,9 @@ import { expandRange, fromMidnightUTCDate, toMidnightUTCDate } from '../utils/mo
 import { addFlashNotification } from './FlashNotificationActions';
 import { blockChanges, unblockChanges } from '../reducers/ManageAvailabilityReducer';
 
+// Delay to show the save button checkmark before closing the winder.
+const SAVE_FINISHED_DELAY = 2000;
```
Gnito: Did you test other delay times?

client/app/components/sections/ManageAvailability/ManageAvailability.css
```diff
@@ -1,3 +1,26 @@
+@keyframes stroke {
```
Gnito: Are these in global namespace (or how these work within this file)?
(If they are in global namespace, maybe a bit more complex names would be better)

client/app/components/sections/ManageAvailability/ManageAvailabilityContainer.js
```diff
@@ -16,13 +16,15 @@ const ManageAvailabilityContainer = ({
...
   saveInProgress,
+  saveFinished,
   reservedDays,
   blockedDays,
   sideWinderWrapper,
 }) =>
       r(ManageAvailability, {
         hasChanges,
         saveInProgress,
+        saveFinished,
```
Gnito: Btw. we do this quite much. What is actually removed from container props?
(I.e. could we do this passing of partial props with "...restProps" approach)
Anyway, just a thought.

client/app/actions/ManageAvailabilityActions.js
kpuputti: The checkmark animation takes 0.9 seconds and the checkmark was supposed to be visible for a second, thus 2s. I don't see a need to test any other delays?

client/app/components/sections/ManageAvailability/ManageAvailability.css
kpuputti: Initially I though so as well, and even prefixed them. But CSS modules is clever enough to also change those keyframe names, so no need to prefix.

client/app/components/sections/ManageAvailability/ManageAvailabilityContainer.js
kpuputti: Yeah, maybe that should be done. Some props are restructured, so I wouldn't start refactoring those here.

PR #2841, opened 2016-12-22
Partial fix: orientation change is fixed to top of sidewinder

Gnito:
(That will keep original scrolling position on the actual page.)


client/app/components/composites/SideWinder/SideWinder.js
```diff
@@ -135,7 +146,7 @@ class SideWinder extends Component {
     }, [
       r(ReactTransitionGroup, [
         div({
-          className: css.overlay,
+          className: classNames('SideWinder_orientationHook', css.overlay),
```
kpuputti: The class name could come from CSS modules.

Gnito:
I also tested this on iPhone7.

PR #2842, opened 2016-12-25
Fix missing currency symbol

Gnito:


app/views/listings/form/_currency_selector.haml
```diff
@@ -1,3 +1,2 @@
 .currency-selector
   = Money.new(1, @current_community.currency).symbol
-  = form.hidden_field :currency, :value => @current_community.currency
```
lyyder: This seems to break the tests, what's the rationale for the hidden fields?

kpuputti: Originally the currency was passed through the form as a hidden field. The commit that this PR reverts changed that to read the currency straight from the community model.

lyyder: @kpuputti @Gnito So with this revert the hidden fields would still be required?

Gnito: @lyyder @kpuputti I added one more commit that reintroduced currency handling in controller. https://github.com/sharetribe/sharetribe/pull/2842/commits/b533b65174a3028dbf81bae404823c4937bb152b

app/views/listings/form/_currency_selector.haml
```diff
@@ -0,0 +1,2 @@
+.currency-selector
+  = Money.new(1, @current_community.currency).symbol
```
kpuputti: This used to be a dropdown since in theory marketplaces supported multiple currencies. That feature was removed, so calling this `currency-selector` seems misleading.

Gnito: I'll rename this to currency-symbol

PR #1664, opened 2016-01-21
Capistrano 3

frontandstart:
Integration Capistrano 3 with capistrano-passenger, capistrano-rvm, capistrano-bundler

PR #2849, opened 2017-01-13
Remove Capfile (Capistrano). stated on #1664 Capistrano is not supported

bekicot:
Remove Capfile (Capistrano). stated on #1664 Capistrano is not supported


rap1ds:
Thanks a lot!

PR #2857, opened 2017-01-26
changes to star icon

Luis-RG:
changes star icon


bladealslayer:
@Luis-RG squash your 3 commits into 1, since the latter are two reverts. Then it's ok.

PR #2862, opened 2017-02-03
Icon updates

Schwerbelastung:


bladealslayer:
@Schwerbelastung you'd need to rebase, since meanwhile some other icon changes were merged. Otherwise, LGTM.

One suggestion from now on - since there are already lots of icons, I'd sort them by name in the file.

PR #2867, opened 2017-02-14
Add notification about SES sending confirmation email

sktoiva:
Adds notification on how email verification works. It will appear when a valid email address is entered to the fields.

screen shot 2017-02-14 at 15 42 32

### DIFF
--- a/app/views/admin/communities/edit_welcome_email.haml
+++ b/app/views/admin/communities/edit_welcome_email.haml
@@ -130,11 +130,20 @@
           %label.sender-address-label= t("admin.communities.outgoing_email.sender_email_label")
           %input.js-sender-email-input{name: "email", type: :text, placeholder: t("admin.communities.outgoing_email.sender_email_placeholder"), disabled: disabled_attr, class: disabled_class}
 
-      .row.js-sender-address-preview-container.sender-address-preview-container.hidden
-        .col-12
-          .sender-address-preview
-            = t("admin.communities.outgoing_email.this_is_how_it_will_look")
-            %span.js-sender-address-preview-values
+      .js-sender-address-preview-container.sender-address-preview-container.hidden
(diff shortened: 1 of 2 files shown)


app/views/admin/communities/edit_welcome_email.haml
```diff
@@ -130,11 +130,20 @@
...
-          .sender-address-preview
-            = t("admin.communities.outgoing_email.this_is_how_it_will_look")
-            %span.js-sender-address-preview-values
+      .js-sender-address-preview-container.sender-address-preview-container.hidden
+        .row
+          .col-12
+            .sender-address-preview
+              = t("admin.communities.outgoing_email.this_is_how_it_will_look")
+              %span.js-sender-address-preview-values
+        .row.js-sender-address-preview-container.sender-address-preview-container.hidden
```
rap1ds: Seems that there's a small inconsistency between lines 133 and 139: 

```
.js-sender-address-preview-container.sender-address-preview-container.hidden
  .row
```

vs.

```
.row.js-sender-address-preview-container.sender-address-preview-container.hidden
```

I don't know what the classes `.js-sender-address-preview-container.sender-address-preview-container` contain so I don't know if it matters that they are a bit different. Probably it doesn't matter.

sktoiva:
The hyphen is related to the fact that the text is inside `p` tags and we have `hyphens: auto;` for `p`. Not gonna change that now.

app/views/admin/communities/edit_welcome_email.haml
sktoiva: They originally control the automatic showing of controlled the automatic showing of the `This is how it will look: ` field. I want the notification to be inside the same logic. Thus, there's now one extra wrapping div for two rows.

PR #2869, opened 2017-02-16
Add possibility to use storefront name

sktoiva:
- [x] Adds a possibility to add a display name to user profile that'll replace other names in all views.
- [x] Admin sees both whole name and display name (also added to the csv):
  
image

- [x] Unifies how info texts are shown in profile edit page:
image

### DIFF
--- a/app/views/settings/show.haml
+++ b/app/views/settings/show.haml
@@ -27,14 +27,25 @@
           = t("settings.profile.#{@current_community.name_display_type}")
     = form.text_field :family_name, :class => "text_field", :maxlength => "30"
 
+    -# Display name
+    .inline-label-container
+      = form.label :display_name, t("settings.profile.display_name"), class: "inline"
+    %span.alert-box-icon
+      = icon_tag("alert", ["icon-fix"])
+    %small
+      = t("settings.profile.display_name_description")
+    = form.text_field :display_name, class: "text_field", :maxlength => "30"
+
     -# Location
     .inline-label-container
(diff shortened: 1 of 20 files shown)


app/views/settings/show.haml
```diff
@@ -27,14 +27,25 @@
           = t("settings.profile.#{@current_community.name_display_type}")
     = form.text_field :family_name, :class => "text_field", :maxlength => "30"
 
+    -# Display name
+    .inline-label-container
+      = form.label :display_name, t("settings.profile.display_name"), class: "inline"
+    %span.alert-box-icon
+      = icon_tag("alert", ["icon-fix"])
```
rap1ds: I'd rather use info-icon instead of an alert-icon. Or what do you think?

spec/controllers/admin/csv_export_spec.rb
```diff
@@ -25,12 +25,14 @@
 
       expect(user["first_name"]).to eq(@person.given_name)
       expect(user["last_name"]).to eq(@person.family_name)
+      expect(user["display_name"]).to eq(@person.display_name || "")
```
rap1ds: I'm not completely sure why you need to use OR `||` here. Is there some unknown configuration coming outside which affects to the output, or why it's not known what will be the end result?

sktoiva: My thinking went so that if you are to add to FactoryGirl a display_name it'll still work as intended. Currently, display_name is nil and then it's supposed to be empty string.

PR #2879, opened 2017-02-23
Replace removed locale with default locale for marketplace users

sktoiva:
Replaces removed locale with default locale for marketplace users and fixes bugs where marketplace has customization for language that's not anymore available.

## To reproduce:

1. Have a marketplace with two languages: English and Swedish
2. Set marketplace names accordingly: English: "Eng" & Swedish: "Swe"
3. Set users locale to Swedish
4. Remove Swedish from available languages
5. Set new membership notifications on
6. Create a new membership

## Expected 

* Email has marketplace name "Eng"

## Actual

* Email has marketplace name "Swe"

There are probably other places also where we rely on locale that's not available.

### DIFF
--- a/app/services/marketplace_service/api/marketplaces.rb
+++ b/app/services/marketplace_service/api/marketplaces.rb
@@ -116,6 +116,14 @@ module MarketplaceService::API
       community_name = community.name(default_locale)
       locales.each { |locale| Helper.first_or_create_community_customization!(community, community_name, locale) }
 
+      # Replace removed locale with default for users of marketplace
+      removed_locales = community.locales - locales
+      if removed_locales.present?
+        UserService::API::Users.replace_with_default_locale(community_id: community.id,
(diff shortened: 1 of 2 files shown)


app/services/marketplace_service/api/marketplaces.rb
```diff
@@ -116,6 +116,13 @@ def set_locales(community, locales)
       community_name = community.name(default_locale)
       locales.each { |locale| Helper.first_or_create_community_customization!(community, community_name, locale) }
 
+      # Replace removed locale with default for users of marketplace
+      removed_locales = community.locales - locales
+      to_locale = removed_locales.include?(default_locale) ? locales.first : default_locale
```
rap1ds: If I read this piece of code right, the `default_locale` here is the old default locale before the update. The new default locale will be the `locales.first`. So would it make sense to set the `to_locale` to `locales.first` in every case?

PR #2886, opened 2017-03-02
Save new user info in transaction. Catch FB pic upload errors

rap1ds:
This commit fixes an issue which happens if Amazon S3 is down. The user creation will partially fail, if the profile picture upload to S3 fails. This issue is addressed by:

- Saving person information in transaction
- Catching the possible exception caused by the S3 upload failure


app/controllers/sessions_controller.rb
```diff
@@ -47,7 +47,7 @@ def create
       @current_user.update_attribute(:facebook_id, session["devise.facebook_data"]["id"])
       # FIXME: Currently this doesn't work for very unknown reason. Paper clip seems to be processing, but no pic
       if @current_user.image_file_size.nil?
-        @current_user.store_picture_from_facebook
+        @current_user.store_picture_from_facebook!
```
rap1ds: I didn't add exception catching here, because I didn't quite understand when this piece of code is executed (if ever? 😬)

PR #2889, opened 2017-03-07
adding Hebrew to unsupported languages

Luis-RG:
Badabing-badabang!


config/available_locales.rb
```diff
@@ -58,6 +58,7 @@ module Sharetribe
     {ident: "sk-SK", name: "Slovenčina", language: "sk-SK", region: "SK", fallback: nil}, # Slovak
     {ident: "ti", name: "Tigrinya", language: "ti", region: "ER", fallback: nil}, # Tigrinya
     {ident: "ar", name: "Arabic", language: "ar", region: "SA", fallback: nil} # Arabic (Saudi Arabia)
```
rap1ds: You need to add comma after Arabic:

```
{ident: "ar", name: "Arabic", language: "ar", region: "SA", fallback: nil}, # Arabic (Saudi Arabia)
``

PR #2894, opened 2017-03-09
Turn expiration off by default

rap1ds:
This PR turn listing expiration off by default for new marketplaces. It also moves the expiration setting from Admin > Settings to Admin > Listing fields & Filters

Screenshots: 

Old:

screen shot 2017-03-09 at 17 01 25

New:

screen shot 2017-03-09 at 16 58 37

screen shot 2017-03-09 at 16 58 18

### DIFF
--- a/app/controllers/admin/custom_fields_controller.rb
+++ b/app/controllers/admin/custom_fields_controller.rb
@@ -200,6 +200,14 @@ class Admin::CustomFieldsController < Admin::AdminBaseController
     @community = @current_community
   end
 
+  def edit_expiration
+    @selected_tribe_navi_tab = "admin"
+    @selected_left_navi_link = "listing_fields"
+    @community = @current_community
+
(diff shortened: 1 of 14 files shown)


app/controllers/admin/custom_fields_controller.rb
```diff
@@ -234,6 +242,26 @@ def update_location
...
+  def update_expiration
+    listing_expiration_enabled = params[:listing_expiration_enabled] == "enabled"
+
+    success = @current_community.update_attributes(
+      { hide_expiration_date: !listing_expiration_enabled })
+
+    if success
+      redirect_to admin_custom_fields_path
+    else
+      flash[:error] = "Expiration field editing failed"
```
sktoiva: Should there be a translation for this?

rap1ds: I was lazy and decided to not add translation, since this shouldn't happen in normal situations.

PR #2906, opened 2017-03-17
Remove default twitter handle

sktoiva:
If marketplace admin hasn't set up a twitter handle we add `via @Sharetribe` at the end of tweets. This doesn't provide value and causes trouble in some cases. This PR removes it.

### DIFF
--- a/app/views/listings/show.haml
+++ b/app/views/listings/show.haml
@@ -105,7 +105,8 @@
             .listing-fb-like-button
               = facebook_like(current_user?(@listing.author))
             .listing-tweet-button
-              = link_to("", "https://twitter.com/share", :class => "twitter-share-button", "data" => {count: "horizontal", via: (@current_community.twitter_handle || "Sharetribe"), text: @listing.title })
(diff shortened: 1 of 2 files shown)


app/views/listings/show.haml
```diff
@@ -105,7 +105,8 @@
             .listing-fb-like-button
               = facebook_like(current_user?(@listing.author))
             .listing-tweet-button
-              = link_to("", "https://twitter.com/share", :class => "twitter-share-button", "data" => {count: "horizontal", via: (@current_community.twitter_handle || "Sharetribe"), text: @listing.title })
+              - twitter_params =  {count: "horizontal", via: (@current_community.twitter_handle), text: @listing.title }.compact
```
gnomet: @sktoiva did you test that this behaves nicely also in case where @current_community.twitter_handle is not set? (and probably nil)

If yes, then all the code looks good to me. 👍

sktoiva: Compact removes all hash keys with nil values.

PR #2914, opened 2017-03-28
Add Admin Intercom

rap1ds:
This PR adds Intercom widget to Admin UI. 

Intercom has several different products, and the product that we are installing is called **Respond**. Also, the Intercom widget will be only available in **Admin UI**. That's why the configuration variables are names e.g. `admin_intercom_respond_enabled`. 

- [X] Add Intercom Respond widget to Admin UI
- [X] Open Intercom widget when "Support" link in Admin UI is clicked
- [X] Send necessary information to Intercom such as user UUID (main idenfier), email, marketplace UUID, marketplace ID, user ID, etc.
- [X] Add a new Feature Flag
- [X] Add a new environment variable `admin_intercom_respond_enabled`
- [X] Add a new environment variable `admin_intercom_respond_test_group_ratio`. Based on this ratio, add `:admin_intercom_respond` Feature flag to newly created trial marketplace.
- [X] Add a new environment variable `admin_intercom_secure_mode_secret`. This will enable Secure mode. Also, add `IntercomHelper.user_hash` to calculate proper secure hash.

Screenshot:

screen shot 2017-03-28 at 15 41 09

### DIFF
--- a/app/views/layouts/_marketplace_head.haml
+++ b/app/views/layouts/_marketplace_head.haml
@@ -100,5 +100,5 @@
 - unless on_admin?
   = @current_community.custom_head_script.try(:html_safe)
 
-- if on_admin? && FeatureFlagHelper.feature_enabled?(:admin_intercom_respond)
-  = render partial: "layouts/intercom", locals: { email: @current_user.primary_email&.address, user_id: @current_user.id, user_uuid: @current_user.uuid_object.to_s, name: PersonViewUtils.person_display_name(@current_user, @current_community), marketplace_id: @current_community.id, marketplace_uuid: @current_community.uuid_object.to_s, marketplace_url: @current_community.full_url }
+- if on_admin? && IntercomHelper.admin_intercom_respond_enabled?
+  = render partial: "layouts/intercom", locals: { app_id: IntercomHelper.admin_intercom_app_id, email: @current_user.primary_email&.address, user_id: @current_user.id, user_uuid: @current_user.uuid_object.to_s, name: PersonViewUtils.person_display_name(@current_user, @current_community), marketplace_id: @current_community.id, marketplace_uuid: @current_community.uuid_object.to_s, marketplace_url: @current_community.full_url }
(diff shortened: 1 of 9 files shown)


config/config.defaults.yml
```diff
@@ -398,6 +398,35 @@ default: &default_settings
...
+  # Enable/disable Intercom Respond for Admin UI
+  admin_intercom_respond_enabled: false
+
+  # Admin Intercom App Id
+  admin_intercom_app_id: ""
+
+  # Admin Intercom secure mode secret
+  admin_intercom_secure_mode_secret: ""
+
+  # The percentage of marketplaces where Admin Intercom Respond
```
sktoiva: I guess this should be percentage of new marketplaces... not all marketplaces.

app/views/layouts/_marketplace_head.haml
```diff
@@ -99,3 +99,7 @@
 
 - unless on_admin?
   = @current_community.custom_head_script.try(:html_safe)
+
+- if on_admin? && IntercomHelper.admin_intercom_respond_enabled?
+  - user_uuid = @current_user.uuid_object.to_s
+  = render partial: "layouts/intercom", locals: { app_id: IntercomHelper.admin_intercom_app_id, email: @current_user.primary_email&.address, user_id: @current_user.id, user_uuid: user_uuid, name: PersonViewUtils.person_display_name(@current_user, @current_community), marketplace_id: @current_community.id, marketplace_uuid: @current_community.uuid_object.to_s, marketplace_url: @current_community.full_url, user_hash: IntercomHelper.user_hash(user_uuid) }
```
sktoiva: Why are the locals passed from here? Not saying that this is wrong, but I want to just understand this. 

It would make perfect sense if we would pass these from a controller, but now it just seems that the responsiblity to extract locals is transferred to another view.

rap1ds: The reason is that this piece of code is part of the main layout. In normal case I think setting the locals in controller and passing them to the view is the perfect way to go, but since there's no one single controller, it's a bit trickier. One thing that could be done is to set the variables in controller instance variables in ApplicationController's before filter, but I'm not sure that's any better, since the application controller is already quite big and the list of filters is long.

PR #2928, opened 2017-04-11
Don't add redirect_to params to links

rap1ds:
Locale change and login links in the sign up had the `redirect_to`/`return_to` query strings in the links. This is problematic, since it makes the URLs unique in the eyes of a search engine crawler bot.

This PR removes those query strings from those URLs.

**Locale change:** If the user is not logged in, the locale change URL can be changed so that it points the user directly to the old redirect URL. We need to go to the `change_locale` endpoint only if the user is logged in and we really want to change the user preferences.

**Login:** Apparently, the `return_to` session value was already in use in the login URL. The new topbar used `return_to` parameter.

In this PR I decided to remove the `return_to` parameters from the new topbar because a) that was easiest and fastest thing to do b) in this PR the main purpose is to solve the crawler bot issue and this solution solves it c) new topbar and old topbar are now in sync.

In the future we might want to carry the `return_to` URL in the query string and use robots.txt to prevent search bots from crawling those. This should be done if we want to avoid creating sessions for anonymous users.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -557,15 +557,23 @@ class ApplicationController < ActionController::Base
       }
     }.or_else({})
 
+    locale_change_links = available_locales.map { |(title, locale_code)|
+      {
+        url: PathHelpers.change_locale_path(is_logged_in: @current_user.present?,
+                                            locale: locale_code,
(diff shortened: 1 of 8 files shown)


bladealslayer:
LGTM

PR #2929, opened 2017-04-11
Trial expired for all

sktoiva:
Adds trial expiration notice for all users. The message shown is targeted for admins and regular users individually.

- [x] two different expiration notices
- [x] new css for the new expiration notice
- [x] move display notice logic to before filter
- [x] skip display notice before filter in new feedback path
- [x] check if skipping it is required elsewhere
- [x] translations for EN, FI, FR
- [ ] new translations for NL, DE, ES

image

image

image

image


app/controllers/application_controller.rb
```diff
@@ -383,6 +384,13 @@ def after_sign_in_path_for(resourse)
     end
   end
 
+  def set_display_expiration_notice
+    ext_service_active = PlanService::API::Api.plans.active?
+    is_expired = Maybe(@current_plan)[:expired].or_else(false)
```
rap1ds: Did you check that `@current_plan` can be `nil`? If yes, then wrapping in Maybe is necessary. If it's always non-nil, then I think the Maybe wrapping should be removed.

config/locales/expiration_lightbox/en.yml
```diff
@@ -1,8 +1,13 @@
...
+    admin:
+      title: "Your Sharetribe trial has expired."
+      sub_title_new: "Don't worry, your data has not been deleted. You can unlock your marketplace immediately by subscribing to one of our paid plans."
+      link_to_external_service: "Choose a plan"
+      need_more_info: "Need more information?"
+      contact_us: "Contact us."
+    title: "The free trial of this marketplace has expired."
+    unavailable_info: "The marketplace is temporarily unavailable. None of the listings, pictures or other data has been deleted.
"
+    all_available_info: "All will be available again when the marketplace owner subscribes to a paid <a href='https://www.sharetribe.com/?utm_source=expired-trial&utm_medium=modal&utm_campaign=end-user-modal-expired-trial'>Sharetribe</a> plan."
```
rap1ds: I understand that since this is a translation file controlled by us, the markup in the translations isn't that big thing. However, I still think the markup it would be great to remove the markup from the translations. This would also avoid some repetition: the link is now repeated 4 times.

That's my 5 cents, but you can decide how to move forward. I'm also ok with this one.

PR #2939, opened 2017-04-19
Add Scheduled tasks documentation, upgrade changelog and upgrade

rap1ds:
- [X] Add Scheduled tasks documentation
- [X] Update the CHANGELOG.md regarding the session store migration
- [X] Update the UPGRADE.md regarding the session store migration

### DIFF
--- a/docs/scheduled_tasks.md
+++ b/docs/scheduled_tasks.md
@@ -2,11 +2,11 @@
 
 You need to configure scheduled tasks in order to properly run your marketplace in production.
 
-Scheduled tasks are tasks that the system needs to run every now and then to keep the application working as expected. Operations such as sending mass emails to all users or cleaning up old data from the database are done in scheduled tasks.
+Scheduled tasks are tasks that the system needs to run regularly to keep the application working as expected. Operations such as sending mass emails to all users or cleaning up old data from the database are done in scheduled tasks.
 
 ## List of required tasks
 
-Here's a list of scheduled tasks you need to configure:
+Here is a list of scheduled tasks you need to configure:
 
 | Purpose | Command | Recommended run interval | Note |
 |---------|---------|--------------------------|------|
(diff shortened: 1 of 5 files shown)


docs/scheduled_tasks.md
```diff
@@ -0,0 +1,25 @@
+# Scheduled tasks
+
+You need to configure scheduled tasks in order to properly run your marketplace in production.
+
+Scheduled tasks are tasks that the system needs to run every now and then to keep the application working as expected. Operations such as sending mass emails to all users or cleaning up old data from the database are done in scheduled tasks.
```
sktoiva: Maybe regularly instead of every now and then to sound more specific.

docs/scheduled_tasks.md
```diff
@@ -0,0 +1,25 @@
...
+
+## List of required tasks
+
+Here's a list of scheduled tasks you need to configure:
```
sktoiva: In technical documentation and formal writing contractions should be avoided. So, `Here is` instead of `Here's`.

docs/scheduled_tasks.md
```diff
@@ -0,0 +1,25 @@
...
+| Synchronize [Amazon SES](https://aws.amazon.com/ses/) state | `bundle exec rake sharetribe:synchronize_verified_with_ses` | Every 10 minutes | Only if Amazon SES is use |
+| Retry and clean PayPal tokens | `bundle exec rake sharetribe:retry_and_clean_paypal_tokens` | Every 10 minutes | Only if PayPal is in use |
+
+## How to configure scheduled tasks
+
+**Heroku** provides an easy to use and free add-on [Heroku Scheduler](https://devcenter.heroku.com/articles/scheduler) to configure scheduled tasks.
+
+In **Unix servers** `crontab` can be used to configure scheduled tasks
+
+Consult your hosting provider documentation to find out what's the recommended way to configure scheduled tasks.
```
sktoiva: Here again the contraction.

docs/scheduled_tasks.md
rap1ds: Do we want to be "technical and formal"? 😉

I guess we do 😊

PR #2945, opened 2017-04-21
Currency formatting

sktoiva:
Internationalizes currency formatting based on user locale. Adds translation options for currencies where standard Rails currency formatting fails: e.g. guess is Danish krone.

Fixes couple of assumptions in JavaScript code about money precision for currencies where there are no fractional units i.e. JPY. Doesn't take into account in JS some weird currencies that might have three fractional units as we don't support them.

More info about decisions here: https://gist.github.com/sktoiva/79cbc4f72e581b756fc0c770b1966c83

Formatting options are:

```yaml
 en:
   number:
     currency:
       format:
         separator: "."
         delimiter: ","
         format: "%u%n"
```
where:
`separator`: decimal separator
`delimiter`: thousand delimiter
`format`: %u => unit %n => number

## Screenshots

E.g. switches locale `fr`, currency `EUR` from:

image

to:

image

- [x] add general utils for money localization
- [x] replace humanized_money_with_symbol with general money utils
- [x] add formatting options for translations
- [x] add translation formatting
- [x] fix fee calculation in new listing form
- [x] fix additional shipping price calculation in `listing.js`
- [x] deploy to staging for testing
- [x] figure out deploy strategy

### DIFF
--- a/app/assets/javascripts/kassi.js
+++ b/app/assets/javascripts/kassi.js
@@ -318,7 +318,7 @@ function initialize_listing_view(locale) {
   });
 }
 
-function updateSellerGetsValue(priceInputSelector, displayTargetSelector, currencySelector, communityCommissionPercentage, minCommission, showReversed) {
+function updateSellerGetsValue(currency, locale, priceInputSelector, displayTargetSelector, currencySelector, communityCommissionPercentage, minCommission, showReversed) {
   // true == Show the fee instead of what's left after the fee
   showReversed = showReversed || false;
 
@@ -337,9 +337,8 @@ function updateSellerGetsValue(priceInputSelector, displayTargetSelector, curren
     }
 
     $display.text(
-      [ST.paymentMath.displayMoney(Math.max(0, displaySum)),
-       $currency.val()]
-        .join(" "));
+      ST.paymentMath.displayMoney(Math.max(0, displaySum), currency, locale, {inCents: false})
+    );
   }
 
   $input.keyup(updateYouWillGet);
--- a/app/assets/javascripts/kassi.js
+++ b/app/assets/javascripts/kassi.js
@@ -318,13 +318,12 @@ function initialize_listing_view(locale) {
   });
 }
 
-function updateSellerGetsValue(currency, locale, priceInputSelector, displayTargetSelector, currencySelector, communityCommissionPercentage, minCommission, showReversed) {
+function updateSellerGetsValue(currency, locale, priceInputSelector, displayTargetSelector, communityCommissionPercentage, minCommission, showReversed) {
   // true == Show the fee instead of what's left after the fee
   showReversed = showReversed || false;
 
   $display = $(displayTargetSelector);
   $input = $(priceInputSelector);
-  $currency = $(currencySelector);
 
   function updateYouWillGet() {
     var sum = ST.paymentMath.parseFloatFromFieldValue($input.val());
@@ -342,7 +341,6 @@ function updateSellerGetsValue(currency, locale, priceInputSelector, displayTarg
   }
 
   $input.keyup(updateYouWillGet);
-  $currency.change(updateYouWillGet);
 
   // Run once immediately
   updateYouWillGet();
--- a/app/controllers/listings_controller.rb
+++ b/app/controllers/listings_controller.rb
@@ -932,7 +932,7 @@ class ListingsController < ApplicationController
   end
 
   def delivery_config(require_shipping_address, pickup_enabled, shipping_price, shipping_price_additional, currency)
-    shipping = delivery_price_hash(:shipping, shipping_price, shipping_price_additional)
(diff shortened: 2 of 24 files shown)


app/assets/javascripts/kassi.js
```diff
@@ -337,9 +337,8 @@ function updateSellerGetsValue(priceInputSelector, displayTargetSelector, curren
     }
 
     $display.text(
-      [ST.paymentMath.displayMoney(Math.max(0, displaySum)),
-       $currency.val()]
-        .join(" "));
+      ST.paymentMath.displayMoney(Math.max(0, displaySum), currency, locale, false)
```
rap1ds: By reading this like, the reader has no clue what the `false` might be. The reader needs to go to search the definition of the function to see that it means `inCents`.

To improve readability, you could either make a variable

```
var inCents = false;
ST.paymentMath.displayMoney( ..., inCents)
```

...or pass a options object:

```
var inCents = false;
ST.paymentMath.displayMoney( ..., {inCents: false})
```

I don't think this is a big issue so no need to change in this PR.

app/controllers/listings_controller.rb
```diff
@@ -932,9 +932,13 @@ def payment_setup_status(community:, user:, listing:, payment_type:, process:)
...
 
   def delivery_config(require_shipping_address, pickup_enabled, shipping_price, shipping_price_additional, currency)
-    shipping = delivery_price_hash(:shipping, shipping_price, shipping_price_additional)
+    shipping = delivery_price_hash(:shipping, shipping_price, shipping_price_additional) if require_shipping_address
     pickup = delivery_price_hash(:pickup, Money.new(0, currency), Money.new(0, currency))
 
+    delivery_opts(require_shipping_address, pickup_enabled, shipping, pickup)
+  end
+
+  def delivery_opts(require_shipping_address, pickup_enabled, shipping, pickup)
```
rap1ds: What's the reasoning behind this change?

app/view_utils/money_view_utils.rb
```diff
@@ -0,0 +1,15 @@
+module MoneyViewUtils
+  extend ActionView::Helpers::NumberHelper
+
+  module_function
+
+  # Converts instance of Money to properly formatted and localized string for view purposes
+  # Replaces spaces with narrow non breaking spaces, we do not want monetary amounts to be split on separate lines
+  def to_humanized(m, locale)
```
rap1ds: This comment might come a bit too late (since you have already replaced all humanize calls). The `locale` could be also optional and default to `I18n.locale`. That makes it a bit easier to the view code to call this function, since the caller doesn't need to pass the locale. The downside of course is that the function is then dependant of the global `I18n.locale` state.

app/assets/javascripts/kassi.js
```diff
@@ -337,9 +337,8 @@ function updateSellerGetsValue(priceInputSelector, displayTargetSelector, curren
     }
 
     $display.text(
-      [ST.paymentMath.displayMoney(Math.max(0, displaySum)),
-       $currency.val()]
```
sktoiva: It's used on line 345. I need to change the positioning of this nearer there to better reflect where and why it is used. I didn't yet touch it as I was concentrating on other things.

app/controllers/listings_controller.rb
sktoiva: Actually, I think I just went through couple different alternatives before settling on that if clause on line 935 to guard against unnecessary calculation of shipping prices. 

During that process I refactored the case statement into a separate function and it just stuck.

app/view_utils/money_view_utils.rb
sktoiva: My reasoning was that this would take everything it could as parameters. And no global state would be used. 
Translations would be handled like a database. 

But it's not that big of a deal to switch the signature to e.g. 

```ruby
def to_humanized(m, locale = I18n.locale)
```
if it really improves calls to this.

app/assets/javascripts/kassi.js
sktoiva: Actually, I'm going to leave it as is. It makes sense to keep all the jQuery element creations together.

app/view_utils/money_view_utils.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  # Converts instance of Money to properly formatted and localized string for view purposes
+  # Replaces spaces with narrow non breaking spaces, we do not want monetary amounts to be split on separate lines
+  def to_humanized(m, locale = I18n.locale)
+    # TODO: Remove feature flag
+    if FeatureFlagHelper.feature_enabled?(:currency_formatting)
+      separator = I18n.t("number.currency.format.separator", locale: locale)
+      number_to_currency(m.amount, unit: m.symbol, locale: locale, precision: m.currency.exponent.to_i)
+        .tr(" ", "\u202F")
+        .gsub("#{separator}00", "") # remove cents if they are zero
```
rap1ds: Number of zeros may very

PR #2949, opened 2017-04-24
replace the new layout section description 

Luis-RG:
replace the new layout section description text with a more accurate one and changes the translation key to replace the current text in all languages.

### DIFF
--- a/app/views/admin/communities/new_layout.haml
+++ b/app/views/admin/communities/new_layout.haml
@@ -17,8 +17,7 @@
   .row
     .col-12
       %h2= t("admin.communities.new_layout.new_layout")
-      %p= t("admin.communities.new_layout.description_roadmap")
-      %p= t("admin.communities.new_layout.description_process", support_link: support_link).html_safe
+      %p= t("admin.communities.new_layout.description_roadmap_new")
 
   = form_for community, :url => admin_new_layout_path, :method => :patch do |form|
     .row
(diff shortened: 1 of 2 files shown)


Luis-RG:
I am not sure about the .html_safe part though...

config/locales/en.yml
```diff
@@ -246,6 +246,7 @@ en:
         new_layout: "New layout"
         description_roadmap: "Sharetribe is currently in the process of rolling out a new and improved layout. During the transition period, you can select which parts of your marketplace should continue using the old layout using the menu below. New features will only be available on the new layout."
         description_process: "Support for the old layout will be removed in the future. The Sharetribe team will notify you before this happens. Questions? Contact the team via %{support_link}."
```
rap1ds: Now that `description_process` and `description_roadmap` aren't used anymore, they can and should be removed from the `en.yml` file.

app/views/admin/communities/new_layout.haml
```diff
@@ -17,8 +17,7 @@
   .row
     .col-12
       %h2= t("admin.communities.new_layout.new_layout")
-      %p= t("admin.communities.new_layout.description_roadmap")
-      %p= t("admin.communities.new_layout.description_process", support_link: support_link).html_safe
```
rap1ds: `support_link` is not used anymore, so it should be removed also from the line 7.

rap1ds:
@Luis-RG No need to use `.html_safe` in this case because the `support_link` inside the translation is removed. So it's good now as it is 👍

PR #2953, opened 2017-04-28
Change custom_field_values.numeric_value to double

rap1ds:
Steps to reproduce:

Prerequisite: Marketplace with numeric custom fields

1. Create a new listing with numeric custom field
2. Add `1234567890` as a numberic custom field value for the new listing
3. Save the listing

Expected: In listing page, you see value `1234567890`

Actual: In listing page, you see value `123457000`

### DIFF
--- a/db/migrate/20170613153960_change_numberic_value_to_double.rb
+++ b/db/migrate/20170613153960_change_numberic_value_to_double.rb
@@ -1,0 +1,9 @@
+class ChangeNumbericValueToDouble < ActiveRecord::Migration
+  def up
+    change_column :custom_field_values, :numeric_value, :double
+  end
+
+  def down
+    change_column :custom_field_values, :numeric_value, :float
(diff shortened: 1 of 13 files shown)


db/migrate/20170613153960_change_numberic_value_to_double.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeNumbericValueToDouble < ActiveRecord::Migration
+  def up
+    change_column :custom_field_values, :numeric_value, :double
```
bladealslayer: How about using `DECIMAL` instead? Upgrading to `double` feels a bit like just pushing the border a bit, but we still don't handle exact numeric representation. With decimals we'd still need to set limit, but at least representation will be exact.

rap1ds:
Inactive. Closing.

Issue #2161, opened 2016-06-23
Error: npm install error 


chandra10207:
##### Sharetribe Version:  
6a531dd2cadb498765b2c69484e20ee7721b2efe

##### Environment: 
ubuntu 16.04 LTS

##### Server mode: 
development

##### Description:

Hello,
I am new to sharetribe 
Setting up the development environment completed successfully upto step 3.
But it shows error while installing npm.

##### Steps To Reproduce:

1. ###########:~/sharetribe$ npm install

Actual result: 

```
narayan@narayan-Inspiron-5559:~/sharetribe$ npm install
> @ postinstall /home/narayan/sharetribe
> cd client && npm install

npm ERR! Linux 4.4.0-22-generic
npm ERR! argv "/usr/bin/nodejs" "/usr/bin/npm" "install"
npm ERR! node v4.4.5
npm ERR! npm  v2.15.5
npm ERR! code EBADPLATFORM

npm ERR! notsup Unsupported
npm ERR! notsup Not compatible with your operating system or architecture: fsevents@1.0.12
npm ERR! notsup Valid OS:    darwin
npm ERR! notsup Valid Arch:  any
npm ERR! notsup Actual OS:   linux
npm ERR! notsup Actual Arch: x64

npm ERR! Please include the following file with any support request:
npm ERR!     /home/narayan/sharetribe/client/npm-debug.log

npm ERR! Linux 4.4.0-22-generic
npm ERR! argv "/usr/bin/nodejs" "/usr/bin/npm" "install"
npm ERR! node v4.4.5
npm ERR! npm  v2.15.5
npm ERR! code ELIFECYCLE
npm ERR! @ postinstall: `cd client && npm install`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @ postinstall script 'cd client && npm install'.
npm ERR! This is most likely a problem with the  package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     cd client && npm install
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs 
npm ERR! Or if that isn't available, you can get their info via:
npm ERR! 
npm ERR!     npm owner ls 
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /home/narayan/sharetribe/npm-debug.log

```

##### Additional information

```

narayan@narayan-Inspiron-5559:~$ cd sharetribe/
[...]

PR #2960, opened 2017-05-08
Making fsevents an optional dependency to allow building on Linux

dmgarland:
From issue #2161 it seems as some users will experience build errors using node.js depending on their version https://github.com/sharetribe/sharetribe/issues/2161

The solution is to make the package an optional dependency, allowing npm install to work on Linux. On mac platforms, the dependency should be loaded as normal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This file follows the best practices from [keepachangelog.com](http://keepachang
 ### Removed
 
 ### Fixed
+- Fix optional dependency issue on Linux with fsevents [#2161](https://github.com/sharetribe/sharetribe/issues/2161)
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ This file follows the best practices from [keepachangelog.com](http://keepachang
 
 ### Fixed
 
+- Fix optional dependency issue on Linux with fsevents [#2161](https://github.com/sharetribe/sharetribe/issues/2161)
```
rap1ds: 6.3.0 is already released so this line should go under the next "unreleased" version. Line 19 in this file.

kpuputti:
Thanks for the contribution!

We used to have an NPM shrinkwrap file that is notoriously difficult to combine with fsevents. Could someone with a Linux machine confirm that the current release still has this problem that this PR attempts to fix? Our CI and deployment servers are running Linux, so I'm not sure if this is still a problem.

CHANGELOG.md
dmgarland: Done

rap1ds:
Inactive. Closing.

PR #2963, opened 2017-05-12
changes knowledge base links to help center links in onboarding pages…

Luis-RG:
… and some other basic pages. There are more links to be changes but they are under a "knowledge_base_url" object that I don't want to touch.


thomasmalbaux:
I've made those changes (and a couple more) via another branch so closing this PR.

PR #2968, opened 2017-05-26
Fix transaction button styles

valdis:


sktoiva:
The commit messages should follow somewhat the format described here: 
https://chris.beams.io/posts/git-commit/

Not every commit is of course applicable to all of the items on that list but as a general guideline it is a pretty good one. 

So, e.g. in this case, it should be something like 

```
Fix transaction button styles

Styles were broken in IE Edge.
```

app/assets/stylesheets/views/_listings.css.scss
```diff
@@ -229,7 +229,11 @@ $listing-author-avatar-height: 108;
 // Transaction action button
 .book-button .content {
   width: 100%;
-  display: block;
+  height: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  margin: auto;
```
Gnito: This seems to work 👍  
(Personally, I'm pretty annoyed by our SASS `@extend` code, which makes it very hard to find correct code from haml-files. I.e. there is no `.book-button` but `.enabled-book-button`.)

sktoiva:
Review ok, remember to run the sharetribe:cs_extract task.

PR #2969, opened 2017-05-26
fixed: Admin UI language change has wrong paths

valdis:
prefix locale path after locale change only if route target does not change


sktoiva:
I guess this can be removed now: 

https://github.com/sharetribe/sharetribe/pull/2969/files#diff-21497849d8f00507c9c8dcaf6288b136L53

PR #2974, opened 2017-06-01
Change help center admin panel links and utm tags

thomasmalbaux:
This adds links to help.sharetribe.com instead of support.sharetribe.com and updates copy texts accordingly.


config/locales/welcome_email/fi.yml
```diff
@@ -12,7 +12,7 @@ fi:
       follow_suggested_steps: "Ylläpitonäkymän kautta voit muokata sivustoasi. Seuraa automaattista ohjeistusta aloittaaksesi sivuston pystytyksen."
       not_alone: "Et ole yksin"
       support_team_and_book_call: "Sharetriben asiakaspalvelutiimi on apunasi! %{knowledge_base_link} löydät vastaukset yleisimpiin kysymyksiin. Saat yhteyden meihin lähettämällä sähköpostia osoitteeseen help@sharetribe.com tai yksinkertaisesti vataamalla tähän viestiin. Voit halutessasi myös %{book_a_call_link} markkinapaikka-asiantuntijamme kanssa."
-      knowledge_base_link_text: "Tietopankistamme"
```
sktoiva: This doesn't work in Finnish like this. The earlier translation was conjugated, now the Help Center isn't. Help Centeristämme sounds awful as well. And if we don't want to conjugate it and we treat it as a kind of a trademark, then it should be passed directly from code and without the translation key `knowledge_base_link_text`.

Maybe something like this, would work for the support_team_and_book_call:

```
Sharetriben asiakaspalvelutiimi on apunasi! Help Center auttaa löytämään vastaukset yleisimpiin kysymyksiisi. Saat yhteyden meihin lähettämällä sähköpostia osoitteeseen help@sharetribe.com tai yksinkertaisesti vataamalla tähän viestiin. Voit halutessasi myös %{book_a_call_link} markkinapaikka-asiantuntijamme kanssa.
```

PR #2976, opened 2017-06-05
WIP: Rails5 upgrade

valdis:


db/structure.sql
```diff
@@ -777,15 +623,11 @@ CREATE TABLE `listing_images` (
   `width` int(11) DEFAULT NULL,
   `height` int(11) DEFAULT NULL,
   `author_id` varchar(255) DEFAULT NULL,
+  `position` int(11) DEFAULT '0',
```
rap1ds: Is this left here by accident? This is not part of the Rails 5 upgrade

db/structure.sql
```diff
@@ -1553,16 +1245,12 @@ CREATE TABLE `transaction_transitions` (
   `transaction_id` int(11) DEFAULT NULL,
   `created_at` datetime DEFAULT NULL,
   `updated_at` datetime DEFAULT NULL,
+  `most_recent` tinyint(1) DEFAULT NULL,
```
rap1ds: Same here, is this left here by accident, or is this required for the Rails 5 update?

features/support/warden.rb
```diff
@@ -1,4 +1,4 @@
-include Devise::TestHelpers
+#include Devise::Test::ControllerHelpers
```
rap1ds: Should this be removed, if it's not necessary?

app/views/comments/destroy.js.erb
```diff
@@ -2,4 +2,4 @@
 $("#comment<%= @comment.id.to_s %>").fadeOut(300, function() { $(this).remove(); });
 
 /* Update comment count */
-$("#comment_count").html("(<%= @comment.listing.comments_count - 1 %>)");
+$("#comment_count").html("(<%= @comment.listing.comments_count %>)");
```
sktoiva: Why is the -1 removed?

client/package.json
```diff
@@ -2,7 +2,8 @@
   "repository": "sharetribe/sharetribe",
   "license": "MIT",
   "engines": {
-    "node": "6.9"
```
rap1ds: Remember to update the README.md which currently states that Node 6.9 is required.

app/utils/entity_utils.rb
```diff
@@ -410,6 +410,9 @@ def error_msg(result)
     end
 
     def with_result(specs:, data:, on_success:, on_failure:)
+      if data.is_a?(ActionController::Parameters)
```
rap1ds: Ideally, it would be great to keep the EntityUtils to work with Hashes only, and keep it unaware of Rails. So that it could be one day extracted to a library. So ideally, I'd move the call to `.to_unsafe_hash` to the caller side. However, I understand that it would require a lot of work to hunt down all the places where `build` or `validate` is called with `Parameters`. So adding the call here makes sense in practice.

What if we add a deprecation warning? Something like:

```
if data.is_a?(ActionController::Parameters)
  ActiveSupport::Deprecation.warn("EntityUtils used with ActionController::Parameters. Convert Parameters to Hash using `.to_unsafe_hash` before using EntityUtils")
  data = data.to_unsafe_hash
end
```

db/structure.sql
valdis: this is required for statesman updated gem version

db/structure.sql
valdis: Yes, this is not part of rails 5 upgrade, will remove it

client/package.json
valdis: README.md updated

app/views/comments/destroy.js.erb
sktoiva: Or actually, why was it there in the first place?

app/utils/entity_utils.rb
valdis: I like this idea, added deprecation warning

features/support/warden.rb
valdis: removed

app/views/comments/destroy.js.erb
valdis: seems that rails 5 has fixed some data reloading, so that the -1 is not needed anymore

app/controllers/homepage_controller.rb
```diff
@@ -251,6 +253,7 @@ def filter_unnecessary(search_params, numeric_fields)
   end
 
   def filter_range(price_min, price_max)
+    logger.info "PRICE:#{price_min} #{price_max}"
```
sktoiva: A test log? Should be removed.

valdis: logging removed

client/package.json
babajon: Installation requirements still state Node 6.9

PR #2979, opened 2017-06-08
Make server and client money formatting consistent

sktoiva:
### Makes money handling in server and client consistent.
There are two places where JavaScript handles displaying of money: new listing creation and ordering a listing with shipping enabled with a separate price for additional items. This pr unifies how client and server show money.

  Previously:

image

  After:

image


valdis:
Looks good!

PR #2995, opened 2017-06-22
Updated copy text to remove mention to a link that was deleted

thomasmalbaux:
"Support" link was removed from the admin panel, this copy text change update a sentence in which it was still used.


Luis-RG:
I think it looks fine :)

PR #3000, opened 2017-06-27
rails5 upgrade - fix listing location update

valdis:


app/controllers/listings_controller.rb
```diff
@@ -488,7 +488,8 @@ def update
 
   def finalise_update(listing, shape, community, update_successful, old_availability)
     if update_successful
-      listing.location.update_attributes(params[:location]) if listing.location
+      location_params = permit_origin_loc_params(params.require(:listing))
```
sktoiva: How does this work, as in original code it's `params[:location]`, and nothing reads that in new code?

valdis: It seems that updating of the listing's `location` has been broken, because in the listing edit form there is no such field with `name="location"`, therefore `params[:location]` should have always returned `nil`

sktoiva: Cool, so this fixes a previously unknown bug.

sktoiva: Again, this information would be cool to have in the pr description/comment description.

sktoiva:
Looks good, let's test that the update works in staging and deploy to production.

PR #3001, opened 2017-06-27
Fix Airbrake error filtering

sktoiva:
Airbrake error filtering does not work until Airbrake is correctly
configured and initialized.


valdis:
Looks good.

PR #3007, opened 2017-06-30
Fix: IDOR vulnerability

valdis:


app/models/person.rb
```diff
@@ -425,12 +425,12 @@ def consent
     community_membership.consent
   end
 
-  def is_marketplace_admin?
-    community_membership.admin?
+  def is_marketplace_admin?(community)
+    community_membership.admin? && community_membership.community_id = community.id
```
sktoiva: The latter is now assigning to community_id and returning a truthy value.

sktoiva: This isn't significant, but to me it would make more sense if the community check is first, before the admin check.

valdis: ouch... fixing!

PR #3018, opened 2017-07-19
Stripe integration

valdis:


bladealslayer:
I haven't located the source of the bugs yet, but the payment onboarding wizard pop-up seems broken now.

1. The pop-up has placeholder text / button, instead showing anything payment-related for the payments step.
2. After setting up stripe keys successfully and setting commission / transaction settings, the wizard doesn't advance to the next step.

config/locales/en.yml
```diff
@@ -233,6 +233,7 @@ en:
         last_community_updates: "Please note that after deletion, you and your marketplace users may still receive the last marketplace update emails."
         you_will_be_redirected_to: "After you have deleted your marketplace you will be redirected to %{destination}. You will not be able to access your marketplace anymore."
         delete_this_marketplace: "Delete this marketplace"
+        payment_preferences: "Payment System"
```
thomasmalbaux: Payment system (no capital S)

config/locales/en.yml
```diff
@@ -318,11 +319,13 @@ en:
           paid: Paid
           confirmed: Confirmed
           canceled: Canceled
-          initiated: "Waiting PayPal payment"
```
thomasmalbaux: It would be great to create a new translation key instead of editing the existing one. We rely on open-source contributors for the translations and it is possible that, for some languages, they won't be updated quickly. In other languages, they could still display "Waiting PayPal payment" (in the translated language) for a while, even if the EN version isn't the initial one. To avoid this, it's better to create a new key: it might be that it isn't translated shortly but at least the English message that will be visible will be the correct one.

bladealslayer:
There are lots of intermediate migrations in this PR now. Fields added, then removed, more fields added, etc. Could we "merge" all migrations once the PR is ready to be merged, so that we don't have too many unnecessary ones?

app/controllers/preauthorize_transactions_controller.rb
```diff
@@ -547,20 +561,30 @@ def ensure_can_receive_payment
...
+      gateway_fields =
+        {
+          merchant_brand_logo_url: logo_url,
+          success_url: success_paypal_service_checkout_orders_url,
+          cancel_url: cancel_paypal_service_checkout_orders_url(listing_id: opts[:listing].id)
+        }
+    when :stripe
+      gateway_fields =
+        {
+          stripe_email: @current_user.primary_email.address,
```
bladealslayer: Related to that, Stripe supports sending arbitrary metadata (key/value pairs) for any object that is created (stripe managed account, charge, transfer, customer, etc). It might be good idea to use the metadata to tag corresponding objects with the Sharetribe database IDs, making it easier for support to debug issues (they can ask for the metadata, the metadata is visible in stripe dashboard).

For instance, a charge could be tagged with the ID of the conversation. A stripe account can be tagged with the ID of the user, etc.

config/locales/en.yml
```diff
@@ -506,6 +509,52 @@ en:
...
       minimum_listing_price_below_tx_fee: "The minimum transaction size has to be greater than or equal to the minimum transaction fee: %{minimum_transaction_fee}."
       minimum_listing_price_below_min: "The minimum transaction size has to be greater than the minimum commission %{minimum_commission}."
+    payment_preferences:
+      title: Payment system preferences
+      integration_info_text_paypal_and_stripe: "<p>If you want to allow your users to accept payments, you  need to connect at least one of the two payment providers supported by Sharetribe: Stripe and PayPal. You can choose to use both, which means yout users will be able to accept payments with either payment method.</p> You ca create a new account with either service if you don't have one already. Once you have connected an account, you can choose a minimum transaction size and a possible commission fee.</p> What is a difference between Stripe and PayPal? Stripe allows you to delay payout to the seller up to 3 months and offeres one payment method (credit card). Paypal provide two payment methods (credit card and PayPal account) and offers a protection program for buyers and sellers</p>"
```
sktoiva: There shouldn't be any html-tags in translations. These end up in webtranslateit and all translators are then required to replicate them across the translations. This means that a single translator that forgets to put in a closing tag or puts in an incomplete tag like `</p` breaks the whole layout for that language.

For that, and security reasons, the translations should be handled like user inputs and sanitized accordingly. If tags are necessary, then those need to be built in backend.

app/controllers/preauthorize_transactions_controller.rb
valdis: 1) this is the buyers email, a buyer will not have a stripe custom account
2) adding metadata information is a good idea, will add this to a separate task

app/controllers/admin/payment_preferences_controller.rb
```diff
@@ -0,0 +1,254 @@
...
+
+    onboarding_popup_locals = OnboardingViewUtils.popup_locals(
+      flash[:show_onboarding_popup],
+      admin_getting_started_guide_path,
+      Admin::OnboardingWizard.new(@current_community.id).setup_status)
+
+    view_locals = {
+      min_commission_percentage: MIN_COMMISSION_PERCENTAGE,
+      max_commission_percentage: MAX_COMMISSION_PERCENTAGE,
+      available_currencies: MarketplaceService::AvailableCurrencies::CURRENCIES,
```
valdis: solved in https://github.com/sharetribe/sharetribe/pull/3018/commits/afd4f143d612ada4b6608232eeb121fa09734c9f

app/assets/stylesheets/views/_account-settings.scss
```diff
@@ -260,3 +260,25 @@
 .paypal-currency-change-warning-text {
   display: none;
 }
+
+.StripeElement {
```
valdis: stripe.js uses .StripeElement, this must be kept as is

app/services/stripe_service/api/stripe_api_wrapper.rb
```diff
@@ -0,0 +1,273 @@
...
+      end
+    end
+
+    def perform_transfer(community, account_id, amount_cents, amount_currency, initial_amount, charge_id)
+      with_stripe_payment_config(community) do |payment_settings|
+        charge = Stripe::Charge.retrieve(charge_id)
+        balance_txn = Stripe::BalanceTransaction.retrieve(charge.balance_transaction)
+        balance_currency = balance_txn.currency.upcase
+        # when platform balance is, say in EUR, but prices are in USD, recalc amount
+        fixed_amount = balance_currency == amount_currency ? amount_cents : (amount_cents * 1.0 / initial_amount * balance_txn.amount).to_i
```
valdis: it is not relevant since we force the sellers to have bank accounts in the same currency as marketplace, and by default use destination payments

vasco2ramos:
We have started implementing Stripe a while ago, and believe this PR could benefit from some refactoring, specially on the payments_preference_controller I think that class should delegate some behavior to other classes. If you think about adding a new payment method in the future you will see that class becomes hard to maintain pretty fast. I can give more specific comments via a review if you guys want.

In any case, thank you for the efforts, Stripe is a great addition to Sharetribe. We have a fork from Sharetribe which we used to build our Marketplace and it's been great for us.

PR #3024, opened 2017-08-08
Fixed: can update custom fields

valdis:
BUG: changing the names of custom listing field options is not possible


valdis:
Tests are random-failing

PR #3026, opened 2017-08-09
Stripe destination charges

valdis:


app/models/stripe_account.rb
```diff
@@ -2,31 +2,26 @@
...
+#  community_id        :integer
+#  stripe_seller_id    :string(255)
+#  first_name          :string(255)
+#  last_name           :string(255)
+#  address_country     :string(255)
+#  address_city        :string(255)
+#  address_line1       :string(255)
+#  address_postal_code :string(255)
+#  address_state       :string(255)
+#  birth_date          :date
```
bladealslayer: Could we go without storing address, birth date, last digits of bank account, etc? As well as any other field that we don't strictly need? (E.g. first and last name probably too). If we need any information of that, we could call Stripe API and get it (i.e. if it's needed to render editing form).

app/services/stripe_service/api/payments.rb
```diff
@@ -29,7 +29,8 @@ def create_preauth_payment(tx, gateway_fields)
         fee        = Money.new(0, subtotal.currency)
 
         description = "Payment #{tx[:id]} for #{tx[:listing_title]} via #{gateway_fields[:service_name]} "
-        stripe_charge = stripe_api.charge(tx[:community_id], source_id, seller_id, total.cents, commission.cents, total.currency.iso_code, description)
+        metadata = {order_id: tx[:id], seller_id: tx[:listing_author_id], payer_id: tx[:starter_id], mode: stripe_api.charges_mode(tx[:community_id])}
```
bladealslayer: I'd prefix the metadata keys with `sharetribe_`.

app/services/stripe_service/api/payments.rb
```diff
@@ -57,7 +58,7 @@ def create_preauth_payment(tx, gateway_fields)
       def cancel_preauth(tx, reason)
         payment = PaymentStore.get(tx[:community_id], tx[:id])
         seller_account = accounts_api.get(community_id: tx[:community_id], person_id: tx[:listing_author_id]).data
-        stripe_api.cancel_charge(tx[:community_id], payment[:stripe_charge_id], seller_account[:stripe_seller_id], reason)
+        stripe_api.cancel_charge(tx[:community_id], payment[:stripe_charge_id], seller_account[:stripe_seller_id], reason, {order_id: tx[:id]})
```
bladealslayer: Last argument is metadata, right? Same as above, let's prefix keys with `sharetribe_`.

app/services/stripe_service/api/payments.rb
```diff
@@ -107,15 +113,28 @@ def payout(tx)
...
           if seller_gets > 0
-            result = stripe_api.perform_transfer(tx[:community_id], seller_account[:stripe_seller_id], seller_gets.cents, payment[:sum].currency, payment[:subtotal].cents, payment[:stripe_charge_id])
+            result = stripe_api.perform_transfer(tx[:community_id],
+              seller_account[:stripe_seller_id],
+              seller_gets.cents,
+              payment[:sum].currency,
+              payment[:subtotal].cents,
+              payment[:stripe_charge_id],
+              {order_id: tx[:id]}
+            )
```
bladealslayer: The stripe API methods tend to have quite a few arguments. How about switching to keyword arguments so that it's clear what is what?

app/services/stripe_service/api/stripe_api_wrapper.rb
```diff
@@ -203,11 +220,12 @@ def get_card_info(customer)
...
 
-    DESTINATION_TYPES = [:platform, :seller]
-    # :platform - marketplace API account is responsible for fees, refunds, etc. funds are captured to platform account and then moved to seller by Transfer
-    # :seller   - Custom Account for seller is responsible for fees, funds are sent by Payout
-    def destination(community)
-      APP_CONFIG.stripe_charge_destination.to_sym
+    DESTINATION_TYPES = [:separate, :direct, :destination]
+    # :separate - Separate charges and transfers
+    # :direct   - Direct charges
+    # :destination - Direct charges
```
bladealslayer: `Direct charges` -> `Destination charges`

config/config.defaults.yml
```diff
@@ -440,16 +440,10 @@ default: &default_settings
...
-  # Disable displaying stored cards for customer
-  stripe_hide_stored_cards: true
-
-  # :connect - use only Stripe Connect
-  # :managed - use only Managed
-  stripe_accounts_mode: :managed
+  # :separate    - Separate charges and transfers 
+  # :direct      - Direct charges
+  # :destination - Destination charges
+  stripe_charges_mode: :destination
```
bladealslayer: I think it's useful to have a short description of each type of charging model here. Or at least a link to the Stripe documentation: https://stripe.com/docs/connect/charges

PR #3033, opened 2017-08-16
Stripe checkout ui fixes

valdis:


app/controllers/payment_settings_controller.rb
```diff
@@ -176,20 +172,26 @@ def load_stripe_account
...
-        :last_name,
+        :legal_name,
         :address_country,
         :address_city,
         :address_line1,
         :address_postal_code,
         :address_state,
         :birth_date,
-        :ssn_last_4
+        :ssn_last_4,
```
sktoiva: At least ssn_last_4 and personal_id_number should be filtered from logs.

valdis: done

PR #3035, opened 2017-08-16
custom landing page: include custom head scripts

valdis:


valdis:
test failures are unrelated to implementation

app/controllers/homepage_controller.rb
```diff
@@ -139,6 +140,13 @@ def index
   # rubocop:enable AbcSize
   # rubocop:enable MethodLength
 
+  include ActionView::Helpers::JavaScriptHelper
+
+  def custom_head_scripts
```
sktoiva: I guess these changes should be in landing_page_controller?

thomasmalbaux: @sktoiva @valdis Is this still a relevant change request?

valdis: @sktoiva @thomasmalbaux issues resolved

PR #3036, opened 2017-08-16
Stripe integration comment fixes

valdis:


db/migrate/20170817035830_provision_stripe_payment_settings_to_all_users.rb
```diff
@@ -0,0 +1,33 @@
+class ProvisionStripePaymentSettingsToAllUsers < ActiveRecord::Migration[5.1]
+  def up
+    communities_with_stripe_ids    = Community.connection.select_values("SELECT community_id FROM payment_settings WHERE payment_gateway = 'stripe'")
```
sktoiva: Models shouldn't be used in migrations. They are evolving and might break the migrations later. Community might be a quite safe bet, but it's a good general rule to either not use them or circumvent the issue by defining the model explicitly in the migration. So adding lines:

```ruby
class Community < ActiveRecord::Base
end
```
would fix the issue.

valdis: done

app/views/admin/payment_preferences/_stripe_connected.haml
```diff
@@ -9,17 +9,17 @@
       .paypal-connected-email.paypal-account-connected
         -# Not a subject of I18n, forced fixed text
         .col-6
-          Secret API Key: 
+          Publishable API Key:
```
sktoiva: This should be `Publishable key`

app/views/admin/payment_preferences/_stripe_form.haml
```diff
@@ -9,15 +9,15 @@
         .row
           .col-4
             -# Not a subject of I18n, forced fixed text
-            = form.label :api_private_key, "Secret API Key"
+            = form.label :api_publishable_key, "Publishable API Key"
```
sktoiva: Placeholders are missing for `Publishable key` and `Secret key`

app/views/admin/payment_preferences/_stripe_connected.haml
valdis: adressed in https://github.com/sharetribe/sharetribe/pull/3042

app/views/admin/payment_preferences/_stripe_form.haml
valdis: addressed in https://github.com/sharetribe/sharetribe/pull/3042

PR #3038, opened 2017-08-17
Added info_trial_creation_status Intercom custom attribute

thomasmalbaux:
This new attribute helps to track if a trial was correctly created.


thomasmalbaux:
This attribute is hard-coded and doesn't come from any database value or so. 

I think I only have to update it here, and my tests in staging worked smoothly. But do you @rap1ds think I'm missing something?

PR #3042, opened 2017-08-22
Stripe admin ui additional fixes

valdis:


app/controllers/admin/payment_preferences_controller.rb
```diff
@@ -200,16 +200,20 @@ def update_payment_preferences
         end
       end
 
-      # Onboarding wizard step recording
-      state_changed = Admin::OnboardingWizard.new(@current_community.id)
-        .update_from_event(:payment_preferences_updated, @current_community)
-      if state_changed
-        report_to_gtm([{event: "km_record", km_event: "Onboarding payments setup"},
-                       {event: "km_record", km_event: "Onboarding paypal connected"}])
+      if form.mode == 'transaction_fee'
```
sktoiva: Just as a side note that this whole method `update_payment_preferences` would benefit from some refactoring ;)

No need to do anything now.

PR #3044, opened 2017-08-23
disable reminders after marketplace subscription expired

valdis:


thomasmalbaux:
Works way better now, good update, makes more sense. Thanks!

PR #3046, opened 2017-08-23
fix datepicker behaviour for listing night availability

valdis:


thomasmalbaux:
Works fine now, thanks.

PR #3047, opened 2017-08-23
force meta tags content to be html-escaped

valdis:


thomasmalbaux:
Looks great and testing worked fine.

PR #3049, opened 2017-08-24
Stripe admin ui additional fixes

valdis:


config/locales/en.yml
```diff
@@ -1028,6 +1028,7 @@ en:
       transaction_requested_by_user: "Great news! %{requester} is interested in \"%{listing_title}\" and has authorized payment for this listing."
       you_have_time_to_accept: "You need to accept or reject the request within %{payment_expires_in}. If you accept within this timeframe, the payment will go through. If you don't accept, the request will be automatically rejected and you will not get paid."
       if_you_do_accept: "If you accept the request within this timeframe, you will receive the money immediately to your account."
+      if_you_do_accept_stripe: "If you accept the request within this timeframe, the payment will go through. You will receive the money directly to your bank account after you have delivered %{listing_title} to %{requester}"
```
sktoiva: Should there be a period at the end?

config/routes.rb
```diff
@@ -171,6 +171,7 @@
       get  "/payment_preferences"                     => "payment_preferences#index"
       put  "/payment_preferences"                     => "payment_preferences#update"
       # PayPal Connect
+      get  "/paypal_preferences" => redirect("/%{locale}/admin/payment_preferences")
```
sktoiva: Will this work now even without the Stripe feature flag? From now on, the payment preferences are always in this URL with or without stripe flag enabled?

valdis: Yes

config/locales/en.yml
valdis: yeah, it probably should :) thanks

PR #3052, opened 2017-08-28
Stripe provider ui fixes and updates

valdis:


config/locales/en.yml
```diff
@@ -2347,5 +2347,6 @@ en:
     to_accept_paypal: To accept PayPal payments, you need to connect your PayPal acclount so we can move money there
     connect_paypal: Connect PayPal account
     wrong_setup: "There is something wrong with the payment system setup in this marketplace. Please contact the marketplace owners."
+    invalid_bank_account_number: "Please check that your bank account is entered correctly and matches your country"
```
sktoiva: Is there a dot missing at the end?

PR #3054, opened 2017-08-30
Stripe checkout additional fixes

valdis:


app/views/transactions/_shipping_address.haml
```diff
@@ -1,6 +1,6 @@
 - fields = [:name, :phone, :street1, :street2, :postal_code, :city, :state_or_province, :country]
-
 - if shipping_address && shipping_address.slice(*fields).values.any?
+  - shipping_address[:country] ||= CountryI18nHelper.translate_country(shipping_address[:country_code]) rescue shipping_address[:country_code]
```
sktoiva: Umm, is this now protecting against a `NoMethodError` from `CountryI18nHelper.translate_country` for undefined country codes? 

I think the nonexistent case should be handled in the helper and here use e.g. Maybe...

valdis: done

PR #3095, opened 2017-10-10
Remove stripe feature flag

thomasmalbaux:


thomasmalbaux:
Not removed yet from:

- app/controllers/admin/getting_started_guide_controller.rb:66
- app/views/listing_conversations/initiate.haml:68
- app/view_utils/stripe_helper.rb:63

PR #3156, opened 2017-11-19
Map icons from bundled fontawesome to main UI

madumlao:
The default iconset used by the sharetribe site is Symbolset Pika,
which is proprietary. Whereas sharetribe comes bundled with
fontawesome + mappings of some actions to fontawesome.

Problem is, a lot of icons in SS Pika are not mapped to an equivalent
fontawesome icon.

This commit adds mappings for SS Pika icons into the bundled
fontawesome.


valdis:
Thanks for your PR. Much appreciated!

thomasmalbaux:
Definitely a big thank you indeed, quite a few people reported this over time and your work will definitely help others :-)

Special ping about this for @rap1ds: I know that he, especially, appreciates your contribution.

PR #3157, opened 2017-11-20
Allow users to load static landing page from config/initializers

madumlao:
By default, if clp_static_enabled: true, then landing page data is read
from app/services/landing_page/example_data.rb. The docs then recommend
editing that file directly to specify the landing page structure.

This is error prone, git-unfriendly, and un-rails-like. The more railsy
approach would be to specify the landing page as configuration data but
default to example_data.rb when the config data is absent.

This commit does just that. If clp_static_enabled, then landing page
data is loaded from CustomLandingPage::StaticData, but if that is not
present, will fall back to loading CustomLandingPage::ExampleData.

CustomLandingPage::StaticData can be populated from a file in
config/initializers. This initializer can automatically be generated
through a rake task, i.e., `sharetribe:landing_pages:install_static`.
The landing page can now be edited from the generated file.


app/services/custom_landing_page/landing_page.rb.template
```diff
@@ -0,0 +1,304 @@
+module CustomLandingPage
```
valdis: This duplicates `app/services/custom_landing_page/example_data.rb`. can you reuse?

app/controllers/landing_page_controller.rb
```diff
@@ -123,8 +123,14 @@ def preview
     begin
       structure = CLP::LandingPageStore.load_structure(cid, preview_version)
 
-      # Uncomment to use static data instead of dynamic from DB
-      # structure = JSON.parse(CustomLandingPage::ExampleData::DATA_STR)
+      # if static landing page enabled, load data from config/initializers/landing_page.rb
+      if (APP_CONFIG.clp_static_enabled)
+        begin
```
valdis: don't use exceptions for flow control

app/services/custom_landing_page/landing_page_store_static.rb
```diff
@@ -10,8 +10,15 @@ def released_version(*)
     end
 
     def load_structure(*)
-      LandingPageStoreDefaults.add_defaults(
+      structure = LandingPageStoreDefaults.add_defaults(
         JSON.parse(CustomLandingPage::ExampleData::DATA_STR))
+      
+      begin
```
valdis: don't use exceptions for flow control

lib/tasks/sharetribe.rake
```diff
@@ -7,6 +7,15 @@ namespace :sharetribe do
     end
   end
 
+  namespace :landing_pages do
+    desc "Install sample landing page into initializers/landing_page.rb"
+    task :install_static => :environment do
+      source = File.join(Rails.root, "app", "services", "custom_landing_page", "landing_page.rb.template")
```
valdis: a newcomer to ShareTribe OS might be confused why static landing pages don't work after running this rake task. Some output to remind setting `clp_static_enabled: true` might be helpful

madumlao: Added note in ec4170606b687f1e0bccc1bb33a3b9e89a7f7274

app/controllers/landing_page_controller.rb
madumlao: Changed flow control to use Module#const_defined? in 2451599639851d7721dc02dfc78078e7bdcabdf9

app/services/custom_landing_page/landing_page_store_static.rb
madumlao: Changed flow control to use Module#const_defined? in 2451599639851d7721dc02dfc78078e7bdcabdf9

app/services/custom_landing_page/landing_page.rb.template
madumlao: Could you give suggestions on how I can reuse here? I duplicated some but not all of `ExampleData` and the module name differs by a bit.

Would a copy + regex substitution suffice?

app/controllers/landing_page_controller.rb
```diff
@@ -123,8 +123,14 @@ def preview
     begin
       structure = CLP::LandingPageStore.load_structure(cid, preview_version)
 
-      # Uncomment to use static data instead of dynamic from DB
-      # structure = JSON.parse(CustomLandingPage::ExampleData::DATA_STR)
+      # if static landing page enabled, load data from config/initializers/landing_page.rb
+      if (APP_CONFIG.clp_static_enabled)
```
valdis: this whole block seems unnecessary, structure is set to the correct value already in `app/services/custom_landing_page/landing_page_store_static.rb` when you call `load_structure`

IMHO you can just update the comment section to say that one should run the rake task and set clp_static_enabled: true

app/services/custom_landing_page/landing_page.rb.template
valdis: 1) Stick to `example_data.rb` file. 
2) You can keep your version of `DATA_STR`, just set the `copyright` text back to "This website is powered by Sharetribe marketplace platform."
3) in the rake task copy the `example_data.rb` and replace the module name to `StaticData` in the new file
4) remove the landing_page.rb.template

app/services/custom_landing_page/landing_page_store_static.rb
```diff
@@ -10,8 +10,14 @@ def released_version(*)
...
     def load_structure(*)
-      LandingPageStoreDefaults.add_defaults(
+      structure = LandingPageStoreDefaults.add_defaults(
         JSON.parse(CustomLandingPage::ExampleData::DATA_STR))
+      
+      if CustomLandingPage.const_defined?("StaticData")
+        structure = JSON.parse(CustomLandingPage::StaticData::DATA_STR)
+      end
+
+      structure
```
valdis: how about
```
def load_structure(*)
  data = if CustomLandingPage.const_defined?("StaticData")
           CustomLandingPage::StaticData::DATA_STR
         else
           CustomLandingPage::ExampleData::DATA_STR
         end
  LandingPageStoreDefaults.add_defaults((JSON.parse(data))
end
```

madumlao:
Updated rake task to pull / patch from CustomLandingPage::ExampleData

Simplified logic of choosing landing pages according to suggestions

valdis:
Good stuff! Thanks!

aivils:
Opened fresh PR https://github.com/sharetribe/sharetribe/pull/3316 . This is closed.

PR #3184, opened 2017-12-19
changing the regexp and test_regexp for NZ accounts.

Luis-RG:
People have complained a lot about the last 2 numbers. They mention that they want 3 numbers are required, if you account has 2 numbers at the end you need to add a 0 before those 2 numbers.


valdis:
It seems this needs additional code to actually insert the 0 if only 2 numbers are given for the last part

PR #3189, opened 2017-12-21
Deploy with Systemd services

leomp12:
Added example .service files for production deployment


valdis:
Hi @leomp12 
First of all, thanks for helping improve sharetribe.

Unfortunately these changes are rather specific to one environment and might conflict with the setup of other sharetribe users, therefore we will not be merging your branch into master.

PR #3218, opened 2018-01-19
Allow filtering signups by regular expressions and limit name lenght

gnomet:
This allows allowed_email field to contain regular expressions that are handled more directly. (not replacing a . with \. like it normally does.

Also the max length of given_name and family_name are limited to 30 characters as longer fields were used for spammy content.

### DIFF
--- a/app/services/email_service.rb
+++ b/app/services/email_service.rb
@@ -87,7 +87,7 @@ class EmailService
     allowed = false
     allowed_array = allowed_emails.split(",")
     allowed_array.each do |allowed_domain_or_address|
-      if allowed_domain_or_address =~ /^\/.+/
+      if allowed_domain_or_address =~ /^\//
         # treat as if this is meant to be a regular expression
         if address =~ Regexp.new(allowed_domain_or_address.delete("/"))
           allowed = true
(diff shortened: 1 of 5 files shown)


app/services/email_service.rb
```diff
@@ -87,11 +87,20 @@ def self.email_address_allowed?(address, allowed_emails)
     allowed = false
     allowed_array = allowed_emails.split(",")
     allowed_array.each do |allowed_domain_or_address|
-      allowed_domain_or_address.strip!
-      allowed_domain_or_address.gsub!('.', '\.') #change . to be \. to only match a dot, not any char
-      if address =~ /#{allowed_domain_or_address}$/i
-        allowed = true
-        break
+      if allowed_domain_or_address =~ /^\/.+/
```
bladealslayer: You don't need the `.+` here.

gnomet: good point, there's no $ in the end
Maybe I won't change now anyway as ready and tested with this, but good comment to improve my RegExps in the future 👍

gnomet: or actually I did fix that too, as some updates (WTI) also still coming in..

PR #3250, opened 2018-03-01
[Booking][Bug] Fixed: Booking does not end properly

valdis:


aivils:
Already fixed in master

PR #3280, opened 2018-04-04
Update Dockerfile

meiremans:
Added ENV BUNDLE_BIN= into Dockerfile as a workaround for the ruby bundler failing to install


bladealslayer:
Thanks! This is now addressed in #3321

PR #3283, opened 2018-04-09
Fix ssn_last_4 not passed to Stripe when creating account token

mdominiak:
Recently merged #3265, adds `ssn_last_4` to the stripe connect form for US accounts but `ssn_last_4` is not included when creating Stripe account token on client side. As a result, the created Stripe account does not have `ssn_last_4` populated.

image


thomasmalbaux:
Thanks a lot for the contribution, @mdominiak! Good catch and great fix. Thanks!

PR #3331, opened 2018-05-23
Add data deletion rake tasks

bladealslayer:
Rake tasks for permanently deleting marketplace data

- delete marketplace data
- delete Stripe Connect account (through Stripe API)

### DIFF
--- a/lib/tasks/sharetribe_data_deletion.rake
+++ b/lib/tasks/sharetribe_data_deletion.rake
@@ -1,0 +1,322 @@
+# Marketplace data deletion tasks
+# DANGER! Tasks here result in data being permanently deleted.
+
+namespace :sharetribe do
+
+  def confirm!(message)
+    print "#{message} (yes/no) "
+    $stdout.flush
+    input = $stdin.gets.chomp
+    unless input == 'yes'
+      raise "Task aborted."
+    end
+  end
+
+  # rubocop:disable MethodLength
+  def delete_marketplace_queries(id)
+  sql = <<~SQL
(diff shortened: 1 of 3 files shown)


lib/tasks/sharetribe_data_deletion.rake
```diff
@@ -0,0 +1,282 @@
...
+    unless input == 'yes'
+      raise "Task aborted."
+    end
+  end
+
+  def delete_marketplace_queries(id)
+  sql = <<SQL
+UPDATE communities SET deleted = 1 WHERE id = #{id};
+
+DELETE FROM feature_flags WHERE community_id = #{id};
```
aivils: @bladealslayer 
With any SQL DELETE You hardcode current relations in the project. Sane is to make revision of each relation and add to the parent model dependent: destroy. 
https://github.com/sharetribe/sharetribe/blob/master/app/models/community.rb#L108

In the ideal case community.destroy will make all children destroy. 
Of course adjust all models requires more effort and typically fails because model foo refuse to destroy himself as result parent destroy fails.

bladealslayer: Yeah, I know, but letting models handle dependent destroys also results in a lot more queries / is slower than issuing single DELETE queries like these here. I know this hardcoded SQL isn't ideal, but I'm willing to go with that for a first version.

That said, in the current task implementation, I am anyway looping over people and listings (which is the tables with potentially largest amount of rows per marketplace), because I wanted to let Rails handle deletion of images on S3. I'm hoping to optimise that at some point (e.g. issue S3 bulk delete requests separately).

PR #3383, opened 2018-07-30
[Google maps] Change the autocomplete feature to use fields

aivils:


app/assets/javascripts/location_search.js
```diff
@@ -11,7 +11,8 @@ window.ST = window.ST || {};
     var form = document.querySelector(selectors.form);
     var autocomplete = new window.google.maps.places.Autocomplete(searchInput, { bounds: { north: -90, east: -180, south: 90, west: 180 } });
     var locationQueryMade = false;
-    autocomplete.setTypes(['geocode']);
```
Gnito: Why is this removed? Sorry, if I haven't been clear`setTypes` is a different setting than `setFields`, both need to be set.

client/app/components/composites/SearchBar/SearchBar.js
```diff
@@ -40,7 +40,7 @@ class SearchBar extends Component {
     const bounds = { north: -90, east: -180, south: 90, west: 180 };
     if (window.google) {
       const autocomplete = new window.google.maps.places.Autocomplete(this.locationInput, { bounds });
-      autocomplete.setTypes(['geocode']);
```
Gnito: Same here.

Gnito:
I found one more place where fields should be used:
https://github.com/sharetribe/sharetribe/blob/master/client/app/utils/places.js#L61
https://developers.google.com/maps/documentation/javascript/places#place_details_requests

aivils:
@Gnito Changes applied

PR #3399, opened 2018-08-14
Added Puerto Rico as Stripe supported country

thomasmalbaux:


aivils:
@thomasmalbaux 
Stripe has large front-end part https://github.com/sharetribe/sharetribe/blob/master/app/assets/javascripts/stripe_form.js

thomasmalbaux:
Very good point, thanks. I'll work on that too (it works exactly as in the US, so should be quite straightforward).

PR #3401, opened 2018-08-15
[Google maps] Change the autocomplete feature to use Session tokens

aivils:


client/app/utils/places.js
```diff
@@ -141,8 +141,10 @@ export const maxDistance = (place) => {
 export const getPrediction = (location) => new Promise((resolve, reject) => {
   const serviceStatus = window.google.maps.places.PlacesServiceStatus;
   const service = new window.google.maps.places.AutocompleteService();
+  const sessionToken = new window.google.maps.places.AutocompleteSessionToken();
```
Gnito: For the session token to take any effect we need to use the same token with predictions and with place details.

The pricing goes so that any autocomplete prediction queries are free if the token is used to fetch place details once (the place detail query ends the "session"). In other words, prediction queries are grouped to place details queries with shared session token and then the only thing that costs is place details query.

aivils: Yes we do. You should forward your request to the google developers :). getDetails() does not have session token. https://developers.google.com/maps/documentation/javascript/places#place_details

Gnito: @aivils sorry, I didn't notice your answer earlier.

Anyway, let me rephrase my point: 
don't create another sessionToken for placeDetails call. You should use the same token that was created for queryPredictions.

The way `sessionToken`s work is that
1) we need to create token (let's asume it is **"asdf"**)
2) then we search predictions for user input `N` **with token "asdf"**
3) then we search predictions for user input `Ne` **with token "asdf"**
3) then we search predictions for user input `New` **with token "asdf"**
4) user selects prediciton `New York`
5) then we get place details for the selected place **with token "asdf"**

All the character queries for new predictions and the final selection that creates a place details query create a single session if the same token is used.

If predictions (aka character queries) and place details query have different sessionToken they are priced twice.

valdis: @Gnito Here we do not send a request to google for each letter that was entered. The autocomplete is handled by googles places autocomplete widget. This code is executed when a user has already selected a location.

Gnito: Yes, I understand. This is a UX improvement for a situation where user presses "enter" instead of choosing something from the widget's dropdown.

This code is making two calls to Google Places API, one for fetching predictions and another for place details. 

My point is just to group those together. It doesn't make sense to create two separate tokens since that will duplicate place details costs. It should be possible to create the token once and pass it into those functions as a function parameter or something.

Gnito: Tried to scan the code a bit deeper about how these call work together:
1) This is the place where predictions are fetched.
2) It should be grouped with getDetails call: https://github.com/sharetribe/sharetribe/blob/master/client/app/utils/places.js#L65

aivils: @Gnito Yep. In abstract. It is not documented. getDetails() has not documented option similar to sessionToken. However this getDetails() does not refuse option sessionToken. I will apply the patch in hopes google developers think similar way as I think.

aivils:
@Gnito 
Changes applied. It is very suspective as did not found documneted option of session token for getDetails()

client/app/utils/places.js
Gnito: @aivils yeah, this pricing change has been one of the worst roll-out Google has ever made. Their API examples and docs are not yet up to date, but they still started to charge 100 times more from their API usage.

Anyway, here's the only place I have found about token usage:
https://developers.google.com/maps/documentation/javascript/reference/places-service#PlaceDetailsRequest

PR #3407, opened 2018-08-24
Fix load_structure method for static landing page.

mduruisseau:
File config/initializers/landing_page.rb was not used by LandingPageStore.


aivils:
It does not change the behavior of the code. It does not load static data even if CustomLandingPage::StaticData is defined.

PR #3418, opened 2018-09-01
Refactor project

levioza:
- Update package.json to work with node greater or equal to 7.8
- Use prettier for the whole project instead of eslint.
- Use commitizen to manage changelogs automatically.
- Use yarn instead of npm.
- Fix the numbering of some markdown files.
- Add the new auto-generated changelog.


thomasmalbaux:
Thanks for the suggestions!

Quite a lot of changes there, and some good ideas that make sense.
We won't move forward with that right now, as we are focusing on other developments, but we'll keep the ideas in mind for future improvements.

Keep up the good work!

PR #3501, opened 2018-11-13
[Security] Add a tag/parameter that added to the marketplace URL disa…

aivils:
…bles the Javascript.


aivils:
@bladealslayer 
Seems this branch is not updated. Later Luis suggest to me to remove people controller changes and allow to disable custom script by param disarm=true only. Now I updated this branch with code delivered to the staging.

PR #3512, opened 2018-11-27
Add support for per-marketplace HSTS configuration

bladealslayer:
Set HSTS headers for marketplaces without custom domain and support per-marketplace HSTS setting for marketplaces with custom domain in use.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -40,6 +40,11 @@ See instructions how to set application in [maintenance mode in Heroku](https://
 
 ## Unreleased
 
+HTTP Strict Transport Security is now enabled by default for sites that do not
+use custom domain (`communities.use_domain = false`) when `always_use_ssl` is
+set to `true` in the configuration. If you wish to disable it, set
+`hsts_max_age` to `0`.
+
 ## Upgrade from 7.3.1 to 7.4.0
 
 Nothing special. See the [#general-update-instructions].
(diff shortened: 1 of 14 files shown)


UPGRADE.md
```diff
@@ -40,6 +40,10 @@ See instructions how to set application in [maintenance mode in Heroku](https://
 
 ## Unreleased
 
+HTTP Strict Transport Security is now enabled by default for sites where
+`always_use_ssl` is set to `true` in the configuration. If you wish to disable
+it, set `hsts_max_age` to `0`.
```
aivils: This is not true for communities uses the domain. I mean open source most comon case is community.use_domain == true . When use_domain is true and hsts_max_age has default nil value then HSTS is disabled and should be enabled by setting hsts_max_age = some value.

bladealslayer: Hm, I think most common open source setup is to actually not use `use_domain`, but rather set `APP_CONFIG.domin`, no?

bladealslayer: Ah, we do have instruction to set `use_domain`. That's problematic. I'd rather then document how open source installations can enable HSTS, instead of changing the default value. The default value is such that it ensures HSTS is only taken intentionally into use in the hosted environment.

bladealslayer: @aivils I now documented enabling HSTS in the readme. I won't make it enabled by default, as this can be quite unexpected.

PR #3525, opened 2018-12-06
Admin acts as another user

aivils:


aivils:
@bladealslayer 
Changes applied in the last commit.

bladealslayer:
@aivils you could still add test case with two marketplaces where admin of one can't post listing as user of another marketplace.

PR #3532, opened 2018-12-10
[Performance] Emails to all members - ActiveRecord solution

aivils:


app/models/person.rb
```diff
@@ -127,6 +131,20 @@ class Person < ApplicationRecord
       .where('given_name like :pattern OR family_name like :pattern OR display_name like :pattern
         OR emails.address like :pattern', pattern: pattern)
   }
+  scope :has_listings, ->  do
+    joins(:listings).distinct
```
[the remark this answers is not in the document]
bladealslayer: Please, also double check that all the other newly added joins have proper indexes to make them efficient.

app/models/transaction.rb
```diff
@@ -118,6 +118,8 @@ class Transaction < ApplicationRecord
             OR listing_authors_transactions.family_name like :pattern
             OR listing_authors_transactions.display_name like :pattern)", pattern: pattern)
   }
+  scope :paid_or_confirmed, -> { where(current_state: ['paid', 'confirmed']) }
```
bladealslayer: `transactions` table does not have index on `current_state`, so this is a slow query. Index should be added here.

app/jobs/create_member_email_batch_job.rb
```diff
@@ -31,15 +31,15 @@ def community_members(mode, community)
...
+        scope = scope.has_no_stripe_account.has_no_paypal_account.has_listings(community)
       when :with_payment_no_listing
-        scope = scope.has_payment_account.has_no_listings
+        scope = scope.has_payment_account.has_no_listings(community)
       when :no_listing_no_payment
-        scope = scope.has_no_stripe_account.has_no_paypal_account.has_no_listings
+        scope = scope.has_no_stripe_account.has_no_paypal_account.has_no_listings(community)
       when :customers
-        scope = scope.has_started_transactions
+        scope = scope.has_started_transactions(community)
```
bladealslayer: Let's add a comment on top of this fn to be extra careful if changing and to test all queries, as some of these can easily become inefficient.

db/migrate/20181211094456_add_indexes_for_send_emails.rb
```diff
@@ -0,0 +1,7 @@
+class AddIndexesForSendEmails < ActiveRecord::Migration[5.1]
+  def change
+    add_index :listings, [:community_id, :author_id, :deleted], name: 'person_community_exist'
```
bladealslayer: This index is not strictly needed, I think, but let's keep it. As far as I could check, queries that join listings, use community_id in the join, which was enough so that the index on (community_id, person_id) is used. Name could be perhaps `community_author_deleted`. This branch was already deployed to staging, right? What's the best way to change a migration then?

app/models/person.rb
```diff
@@ -127,6 +131,21 @@ class Person < ApplicationRecord
       .where('given_name like :pattern OR family_name like :pattern OR display_name like :pattern
         OR emails.address like :pattern', pattern: pattern)
   }
+  scope :has_listings, ->(community) do
+    joins("INNER JOIN `listings` ON `listings`.`author_id` = `people`.`id` AND `listings`.`community_id` = #{community.id} AND `listings`.`deleted` = 0").distinct
+  end
+  scope :has_no_listings, ->(community) do
+    joins("LEFT OUTER JOIN `listings` ON `listings`.`author_id` = `people`.`id` AND `listings`.`community_id` = #{community.id} AND `listings`.`deleted` = 0")
+    .where(listings: {author_id: nil}).distinct
```
bladealslayer: If this scope is used directly on `people` instead of through `community.members` it results in full scan of `people` table, which is obviously very bad. I suggest that we either have a comment here warning to always use this through `community.members` or adding condition:

```
.where(listings: {author_id: nil}, people: {community_id: community.id})
```

Right now, this scope is used only through `community.members`, so it's not an immediate issue, but I can easily see this being overlooked.

app/models/person.rb
```diff
@@ -127,6 +131,21 @@ class Person < ApplicationRecord
...
+  end
+  scope :has_stripe_account, -> { where(id: StripeAccount.active_users.select(:person_id)) }
```
bladealslayer: `stripe_accounts` table is missing index on `person_id`. Right now, the table is not large, so that's not an issue, but the index should be added.

db/migrate/20181211094456_add_indexes_for_send_emails.rb
```diff
@@ -0,0 +1,7 @@
...
+    add_index :community_memberships, [:community_id, :person_id, :status], name: 'person_community_status'
```
bladealslayer: As far as I could test, this index didn't contribute much. Did you see some query where this helps @aivils? Also index name could be better as `community_person_status` to reflect correct order.

app/jobs/create_member_email_batch_job.rb
```diff
@@ -14,52 +14,37 @@ def perform
     current_community = Community.where(id: community_id).first
 
     Delayed::Job.transaction do
-      recipient_ids(mode, current_community).each do |recipient_id|
-        Delayed::Job.enqueue(CommunityMemberEmailSentJob.new(sender_id, recipient_id, community_id, content, locale))
+      community_members(mode, current_community).find_each do |recipient|
```
bladealslayer: All you need from each person is the ID here, so it would be more efficient to pluck it and loop over the IDs. That way you avoid transferring tons of unnecessary data from DB.

Something like:

```
community_members(mode, current_community).pluck(:id).each do |recipient_id|
   ...
end
```

db/migrate/20181211094456_add_indexes_for_send_emails.rb
```diff
@@ -0,0 +1,7 @@
...
+    add_index :transactions, [:community_id, :starter_id, :current_state], name: 'starter_community_state'
```
bladealslayer: Perhaps rename to `community_starter_state`?

app/jobs/create_member_email_batch_job.rb
aivils: Could not agree Your point of view. That pluck is efficient for 100 members. find_each covers any record count 1 or 1000000 without CPU/RAM overload. Probably ActiveRecord scope can be combined with select scope.select(something) (not sure right now).

bladealslayer: Good point, @aivils. Since pluck would return array of IDs, that's probably not too much memory even there are hundreds of thousands of IDs. An ID is 22 bytes + some overhead, so e.g. 500 000 IDs would be little more than 10 MB in size.

On the other hand find_each is more efficient on the app code side, but would require the complex queries for finding the users to be executed multiple times (for each batch), resulting possibly to a bit higher load on the DB.

I'm not sure which is better, without testing more carefully, so I leave it to you. Either way is probably ok, if we make the queries efficient enough.

PR #3544, opened 2018-12-15
Fix typo for max length of sign up email field

knowler:


thomasmalbaux:
Thanks, very good catch!

PR #3557, opened 2018-12-21
Optimize for queries used during data deletion

bladealslayer:


app/models/person.rb
```diff
@@ -86,7 +86,9 @@ class Person < ApplicationRecord
   # This is in addition to a real persisted field like 'username'
   attr_accessor :login
 
-  has_many :listings, -> { where(deleted: 0) }, :dependent => :destroy, :foreign_key => "author_id"
+  # Utilize index on community_id + author_id
+  has_many :listings, ->(person) { where(community_id: person.community_id).exist }, :dependent => :destroy, :foreign_key => "author_id"
```
aivils: Does it really do not break the whole project?

bladealslayer: At least tests passed. I actually think we didn't use this relationship in the code directly, except in data deletion. But perhaps adding index is safer.

PR #3558, opened 2018-12-21
Set shape in listing_presenter

mveer99:
Fix nil shape error in price.haml when creating a new listing with online payment


aivils:
Thank you very much for your participation. Your patch duplicates this line https://github.com/sharetribe/sharetribe/blob/master/app/controllers/listings_controller.rb#L327 added some time ago.

PR #3616, opened 2019-02-15
Use ipv4.pool.sks-keyservers.net

psalias2006:


bladealslayer:
@psalias2006 can you elaborate, in what cases is that extra keyserver pool needed?

bladealslayer:
Addressed in #3684

PR #3620, opened 2019-02-18
[Admin panel][Listings] Add the ability to have approved listings

aivils:


db/migrate/20190213082646_add_approval_to_listings.rb
```diff
@@ -0,0 +1,6 @@
+class AddApprovalToListings < ActiveRecord::Migration[5.1]
+  def change
+    add_column :listings, :state, :string, default: 'approved'
+    add_index :listings, :state
```
bladealslayer: This index alone doesn't help much. Listings are normally selected first by community_id, then by other columns. The main listings query that must be optimised is the one on the homepage, happening here: https://github.com/sharetribe/sharetribe/blob/ability-to-have-pre-approved-listings/app/services/listing_index_service/search/database_search_helper.rb#L29

That query translates to:

```sql
SELECT  `listings`.* FROM `listings` WHERE `listings`.`community_id` = 7198 AND `listings`.`deleted` = 0 AND `listings`.`open` = 1 AND `listings`.`state` = 'approved' AND (valid_until IS NULL OR valid_until > '2019-02-25 15:22:38.617921') ORDER BY listings.sort_date DESC;
```

The existing index that covers this is: `homepage_query_valid_util` on `(community_id, open, valid_until, sort_date, deleted))`. So now we need a new one that contains also `state`. The state needs to be before the `sort_date` column, as it is used for final ordering. I suggest the following index:

```sql
CREATE INDEX listings_homepage_query ON listings(community_id, open, state, deleted, valid_until, listing_shape_id, sort_date);
```

Column order in the index should be slightly better and I suggest adding the `listing_shape_id`, which sometimes is used in the homepage query.

The new index can be added and the old `homepage_query` and `homepage_query_valid_until` can be removed, as this index should cover all uses of the old ones. Make sure the migration first adds the new index and then deletes the old ones, so that the database is not left without proper index at any point.


Similar situation with the old `updates_email_at` index. It should cover the `state` column. Add new index:

```sql
CREATE INDEX updates_email ON listings(community_id, open, state, deleted, valid_until, updates_email_at);
```

While checking that, I noticed that `deleted` listings don't seem to be excluded from update email. @aivils should we add check for `deleted` in the `currently_open` scope? Is there any place that should still be able to see deleted listings? It seems to be easy to forget including `.exist` in all places where listing scope is used. Or not?

bladealslayer: The speedup for the homepage query with the index I suggest is ~9-11% in some non-scientific test I did.

db/migrate/20190305112030_add_state_indexes_to_listings.rb
```diff
@@ -0,0 +1,6 @@
+class AddStateIndexesToListings < ActiveRecord::Migration[5.1]
+  def change
+    add_index :listings, [:community_id, :open, :state, :deleted, :valid_until, :sort_date], name: 'listings_homepage_query'
+    add_index :listings, [:community_id, :open, :state, :deleted, :valid_until, :updates_email_at, :created_at], name: 'listings_updates_email'
```
aivils: @bladealslayer 
Literary implemented. EXPLAIN does not show improvement for me. Not merged in staging due to hard rollback.

bladealslayer: @aivils the old `homepage_query` and `updates_email_listings` can be removed, no need to keep them around. Just make sure they are removed after the new ones are created.

bladealslayer: @aivils did you see some need to add `:created_at` to the second index? I'm not saying it shouldn't be there, I just didn't notice that it was needed.

aivils: @bladealslayer 
Both updates_email_at and created_at are used.
https://github.com/sharetribe/sharetribe/blob/master/app/models/community.rb#L543

bladealslayer: Ok, so as soon as you add code to remove the old unnecessary indexes, this PR LGTM @aivils

aivils:
@bladealslayer 
Changes applied

PR #3625, opened 2019-02-22
JsAdapter class naming convention changes.

kpadma0928ut:
When we try to call ListingImageJSAdapter/ListingImageS3OptionsJSAdapter getting the below error because of wrong naming convention.

Unable to annotate app/models/listing_image_js_adapter.rb: Unable to autoload constant ListingImageJsAdapter, expected /home/unify/git_repo/p21-os-sharetribe/app/models/listing_image_js_adapter.rb to define it
Unable to annotate app/models/listing_image_s3_options_js_adapter.rb: Unable to autoload constant ListingImageS3OptionsJsAdapter, expected /home/unify/git_repo/p21-os-sharetribe/app/models/listing_image_s3_options_js_adapter.rb to define it

### DIFF
--- a/app/controllers/listing_images_controller.rb
+++ b/app/controllers/listing_images_controller.rb
@@ -59,7 +59,7 @@ class ListingImagesController < ApplicationController
     if !listing_image
       render body: nil, status: 404
     else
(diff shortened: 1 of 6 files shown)


aivils:
Thank you very much.

PR #3630, opened 2019-02-26
Use search service when listing shape is used as filter

bladealslayer:


app/services/listing_index_service/search/database_search_helper.rb
```diff
@@ -45,7 +45,7 @@ def needs_search?(search)
       :longitude,
       :distance_max,
       :sort,
-      :listing_shape_id,
+      :listing_shape_ids,
```
aivils: This does not work with other tricky code. 

https://github.com/sharetribe/sharetribe/blob/master/app/services/listing_index_service/search/sphinx_adapter.rb#L19

bladealslayer: Thanks @aivils nice catch! I updated the code now to check both params and tested locally with both sphinx and our search service. Seems to work ok.

PR #3659, opened 2019-03-18
Update robots.txt to slow down mj12bot

bladealslayer:


lib/rack_middleware/robots_generator.rb
```diff
@@ -68,6 +68,10 @@ def self.redirect_to(location)
 
   def self.index_content(req)
     [
+      "User-agent: MJ12bot",
+      "Disallow: /*auth$",
+      "Crawl-Delay: 20",
+      "Sitemap: #{req.scheme}://#{req.host_with_port}/sitemap.xml.gz",
```
thomasmalbaux: No need to repeat the sitemap instructions (once is enough, for all bots).

thomasmalbaux: There should be a line break between the two sets of instructions. Not sure if the current code creates it?

bladealslayer: @thomasmalbaux fixed, this is how it looks now:

```
User-agent: MJ12bot
Disallow: /*auth$
Crawl-Delay: 20

User-agent: *
Disallow: /*auth$
Crawl-Delay: 5
Sitemap: https://MARKETPLACE_URL/sitemap.xml.gz
```

PR #3672, opened 2019-03-22
Stripe listen for more information from stripe

aivils:


app/jobs/stripe_payout_job.rb
```diff
@@ -15,5 +15,6 @@ def perform
     StripeService::API::Api.payments.payout(tx)
   rescue => exception
     error(self, exception)
+    raise
```
bladealslayer: This will cause the delayed job worker to retry the job, wouldn't it? Retrying isn't safe until there are proper idempotency keys defined for the stripe API calls, as the Stripe call may end up being executed multiple times. In this case paying out money to providers earlier. Or am I missing something, @aivils?

aivils: @bladealslayer 
This will keep job in the DB instead vanish into thin air. Yes, I forgot to limit retry count to 1.

PR #3696, opened 2019-04-04
Updated Facebook login button - changed logo and color

aivils:


app/models/marketplace_configurations.rb
```diff
@@ -10,9 +10,9 @@
 #  created_at            :datetime         not null
 #  updated_at            :datetime         not null
 #  limit_search_distance :boolean          default(TRUE), not null
-#  display_about_menu    :boolean          default(TRUE), not null
-#  display_contact_menu  :boolean          default(TRUE), not null
-#  display_invite_menu   :boolean          default(TRUE), not null
+#  display_about_menu    :boolean          default(TRUE)
```
thomasmalbaux: Hmm I'm curious and confused: why is that related to the Facebook button guideline changes? Or was it important for some reason?

aivils: @thomasmalbaux 
File removed. Unfortunately, sometime annotation works weird.

PR #3714, opened 2019-04-12
Use index hint for homepage query

bladealslayer:
DB server doesn't seem to always pick the correct index, so we give it a hint.

### DIFF
--- a/app/services/listing_index_service/search/database_search_helper.rb
+++ b/app/services/listing_index_service/search/database_search_helper.rb
@@ -16,18 +16,13 @@ module ListingIndexService::Search::DatabaseSearchHelper
         listing_shape_id: Maybe(search[:listing_shape_ids]).or_else(nil)
       })
 
-    query = Listing
-            .where(where_opts)
-            .includes(included_models)
-            .order("listings.sort_date DESC")
-            .paginate(per_page: search[:per_page], page: search[:page])
-
-    listings =
-      if search[:include_closed]
-        query
-      else
-        query.currently_open
-      end
+    scope = Listing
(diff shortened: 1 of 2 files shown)


app/services/listing_index_service/search/database_search_helper.rb
```diff
@@ -16,11 +16,21 @@ def fetch_from_db(community_id:, search:, included_models:, includes:)
...
-    query = Listing
-            .where(where_opts)
-            .includes(included_models)
-            .order("listings.sort_date DESC")
-            .paginate(per_page: search[:per_page], page: search[:page])
+    query = if !search[:include_closed] && !search[:author_id]
+              # Add index hint for homepage query
+              Listing
+                .use_homepage_index
+                .where(where_opts)
```
aivils: @bladealslayer 
scope is chainable

```
scope = Listing
scope = scope.use_homepage_index !search[:include_closed] && !search[:author_id]
scope = scope.currently_open unless search[:include_closed]
query = scope.where(where_opts)
                .includes(included_models)
                .order("listings.sort_date DESC")
                .paginate(per_page: search[:per_page], page: search[:page])
```

bladealslayer: Indeed, thanks!

bladealslayer:
@aivils does this look ok to you now?

bladealslayer:
@aivils you you means something like:

```ruby
  HOMEPAGE_INDEX = "listings_homepage_query"
  scope :use_index, ->(index) { from("#{self.table_name} USE INDEX (#{index})") }
  scope :use_homepage_index, -> { use_index(HOMEPAGE_INDEX) }
```

PR #3722, opened 2019-04-17
Rails 5.2.3

aivils:


app/controllers/community_memberships_controller.rb
```diff
@@ -30,7 +30,7 @@ def pending_consent
 
   def give_consent
     form_params = params[:form] || {}
-    values = Form.call(form_params)
+    values = Form.call(form_params.to_unsafe_hash)
```
bladealslayer: Would it be too complex to start whitelisting parameters and avoid using unsafe hashes, @aivils? Here and all other places `.to_unsafe_hash` is called.

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -0,0 +1,38 @@
+# Be sure to restart your server when you modify this file.
+#
+# This file contains migration options to ease your Rails 5.2 upgrade.
+#
+# Once upgraded flip defaults one by one to migrate to the new default.
```
bladealslayer: @aivils once rails upgrade is rolled out, will we go through these and enable the new defaults? Do you see any potential issues?

app/controllers/community_memberships_controller.rb
aivils: @bladealslayer 
Avoiding unsafe hashes would be breaking changes. There is a warning https://github.com/sharetribe/sharetribe/blob/master/app/utils/entity_utils.rb#L416

Removing warning and refactoring EntityUtils is a serious step to affect the whole project.

config/initializers/new_framework_defaults_5_2.rb
aivils: @bladealslayer 
Yes, this is forgotten. I will check it out.

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -8,7 +8,7 @@
 
 # Make Active Record use stable #cache_key alongside new #cache_version method.
 # This is needed for recyclable cache keys.
-# Rails.application.config.active_record.cache_versioning = true
+Rails.application.config.active_record.cache_versioning = true
```
aivils: @bladealslayer 
cache_versioning work with "dalli" gem. "readthis" gem never noticed cache_versioning sooner does not work with cache_versioning. Please check out what is written in the variables for this https://github.com/sharetribe/sharetribe/blob/rails-5.2.3/config/environments/production.rb#L72

There is 3rd party gem "memcachier" involved here. I don't know it is working or not.

PR #3749, opened 2019-05-09
Fixed spelling mistake

landsverk:

PR #3751, opened 2019-05-09
Update edit.haml

landsverk:
Fixed spelling error.

### DIFF
--- a/app/views/admin/communities/footer/edit.haml
+++ b/app/views/admin/communities/footer/edit.haml
@@ -34,7 +34,7 @@
     %h3= t("admin.communities.footer.footer_style")
     .row
       .col-12.footer-theme-radio-buttons
-        = form.collection_radio_buttons :footer_theme, @service.foter_themes, :key, :value
+        = form.collection_radio_buttons :footer_theme, @service.footer_themes, :key, :value
 
     %h3= t("admin.communities.footer.menu_links")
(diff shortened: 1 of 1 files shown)


app/views/admin/communities/footer/edit.haml
```diff
@@ -34,7 +34,7 @@
     %h3= t("admin.communities.footer.footer_style")
     .row
       .col-12.footer-theme-radio-buttons
-        = form.collection_radio_buttons :footer_theme, @service.foter_themes, :key, :value
+        = form.collection_radio_buttons :footer_theme, @service.footer_themes, :key, :value
```
aivils: Please create sane patches! The single patch must implement a single feature and do not break anything.

thomasmalbaux: @aivils My understanding is that this was done through another PR, at https://github.com/sharetribe/sharetribe/pull/3749.

It definitely would be better to have everything in a single PR, and if @landsverk can do that it would be perfect. But maybe we could also accept both PR at the same time, to still welcome the contribution?

@aivils I trust how you prefer to do in regards to best practices.
@landsverk Ideally you should apply both changes into that single PR and close the other one.

aivils:
PR closed  by giving preference to this https://github.com/sharetribe/sharetribe/pull/3757

PR #3761, opened 2019-05-28
Category translation caching improvements

bladealslayer:


app/views/homepage/index.haml
```diff
@@ -96,19 +96,19 @@
               .toggle-header-container
                 .toggle-header
                   - if selected_category
-                    = selected_category.display_name(I18n.locale)
+                    = @category_display_names[selected_category.id]
```
aivils: @bladealslayer 
This is solved by RoR. You do not need your own implementation. https://guides.rubyonrails.org/caching_with_rails.html#fragment-caching

bladealslayer: @aivils I tried that, but the issue is then there will be one fragment cached for each state of the navigation menu - i.e. when user selects different category, different item will be created in the cache, so the cache isn't that useful. Imagine marketplace with 100+ categories and multiple languages - for a cold cache, each click on the category menu will need to create new cached menu fragment. Do you know some way to take care of that issue with fragment cache?

aivils: @bladealslayer 
From my point of view keeping code closer to genuine RoR is more efficient in the long term. RoR solves same troubles we solve. Here you should make precise measurements to decide what is more correct.

app/controllers/homepage_controller.rb
```diff
@@ -204,6 +208,23 @@ def find_listings(params:, current_page:, listings_per_page:, filter_params:, in
     end
   end
 
+  # Time to cache category translations per locale
+  CATEGORY_DISPLAY_NAME_CACHE_EXPIRE_TIME = 24.hours
```
aivils: @bladealslayer 
For such a long period of time Category model must have a hook after_save cleans up the cache for this community category names. Admin changes in the categories must be available immediately in front-end.

bladealslayer: The cache key is formed based on top categories and their modification time. Child categories and translations all have `touch: true` for "parent", so whenever anything is changed, the cache is effectively invalidated. Seems to work fine in my testing.

app/views/homepage/index.haml
bladealslayer: Indeed. However, I couldn't get satisfactory result. Imagine the following scenario (which is actually happening): marketplace with 300 categories and 20+ languages, where a web crawler starts crawling. As I said above, if the menu is cached as fragment, separate cache entries will exist for each combination of locale and `current_category` that are there. So if a crawler / user clicks through all categories one by one, each consequent request will not take advantage of the menu cached on the previous click.

PR #3770, opened 2019-06-11
[bug] fix i18n error with same delimiter and separator in RU and HR l…

aivils:
…ocales

### DIFF
--- a/config/locales/hr.yml
+++ b/config/locales/hr.yml
@@ -3,7 +3,7 @@ hr:
     currency:
       format:
         separator: ","
-        delimiter: ","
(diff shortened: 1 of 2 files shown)


thomasmalbaux:
Thanks, good catch! I didn't know that it could break things to have separator = delimiter. Good to know.

Changes to the translation files should happen via WebTranslate It (except for en.yml, or they'll be overwritten by the next update).

I've now made the change and ran a wti pull at https://github.com/sharetribe/sharetribe/commit/ce211f32648b65375091835a97187c73324fbe79. I'll close that PR.

PR #3774, opened 2019-06-18
Landing page editor

aivils:


app/jobs/cleanup_landing_page_assets_job.rb
```diff
@@ -0,0 +1,21 @@
...
+    # Set the correct service name to thread for I18n to pick it
+    ApplicationHelper.store_community_service_name_to_thread_from_community_id(community_id)
+  end
+
+  def perform
+    community = Community.find(community_id)
+    existing_asset_ids = LandingPageVersion.where(community_id: community_id).map do |lpv|
+      (lpv.parsed_content['assets'] || []).map{|asset| asset['asset_id']}
+    end.flatten.compact.uniq
+    community.landing_page_assets.where.not(id: existing_asset_ids).destroy_all
```
aivils: @bladealslayer 
Please pay attention here.  community.landing_page_assets is ActiveSorage never cleaned up except this job. community.landing_page_assets are shared between different landing page versions as URL of assets. Landing page versions do not copy assets from the previous version, instead just use existing assets from the previous version. As a result, direct clean up of assets is not possible.
This job will be performed after the release of landing page version. This job will collect existing, used assets from all landing page versions and clean up unused assets.

PR #3778, opened 2019-07-03
Landing page editor hero

valdis:


app/models/landing_page_asset.rb
```diff
@@ -0,0 +1,40 @@
...
+        access_key_id: APP_CONFIG.aws_access_key_id,
+        secret_access_key: APP_CONFIG.aws_secret_access_key
+      }
+  else
+    has_attached_file :image,
+      :path => ":rails_root/public/landing_page/:site_name/:filename",
+      :url => ":filename"
+  end
+
+  Paperclip.interpolates :site_name do |attachment, style|
```
bladealslayer: We should not be using Paperclip for new features, as it is deprecated. Is there a reason why we can't use the new Rails ActiveStorage here?

PR #3791, opened 2019-07-19
Stripe sca 1 paymentintent based preauth process

aivils:


app/services/stripe_service/api/payments.rb
```diff
@@ -48,8 +62,25 @@ def create_preauth_payment(tx, gateway_fields)
...
           fee_cents: fee.cents,
           subtotal_cents: subtotal.cents,
-          stripe_charge_id: stripe_charge.id
-        })
+          stripe_charge_id: stripe_charge.try(:id)
+        }
+        if intent
+          payment_data[:stripe_payment_intent_id] = intent.id
+          if intent.status == 'requires_action' &&
+             intent.next_action.type == 'use_stripe_sdk'
```
bladealslayer: I think we shouldn't only handle `use_stripe_sdk`. It's possible to get a redirect action too, I think, but Stripe.js should be able to handle that in the UI too via iframe. Was there a reason whey next_action.type is checked here?

app/services/stripe_service/api/stripe_api_wrapper.rb
```diff
@@ -299,5 +299,41 @@ def delete_account(community:, account_id:)
...
     end
+
+    def create_payment_intent(community:, seller_account_id:, payment_method_id:, amount:, currency:, fee:, description:, metadata:)
+      with_stripe_payment_config(community) do |payment_settings|
+        Stripe::PaymentIntent.create(
+          capture_method: 'manual',
+          payment_method: payment_method_id,
+          amount: amount,
+          currency: currency,
+          confirmation_method: 'manual',
```
bladealslayer: Looks like you chose to implement the manual confirmation flow, instead of automatic confirmation one. Was there some reason for that? I originally spec'ed this to use automatic confirmation. I guess either one can work, just wondering if you had some reason to change that.

app/services/stripe_service/api/payments.rb
aivils: @bladealslayer 
No way to check something in the other place. This wrapped into the large count of methods. Most safe communication with front-end is via DB as this was implemented. Stripe docs advise to use this code. I could not reply about some Stripe internals.

bladealslayer: Accordig to Stripe docs another `next_action` type could be `redirect_to_url`, but to be honest I don't know under what conditions that can happen - when is it going to be `use_stripe_sdk` and when `redirect_to_url`.

https://stripe.com/docs/api/payment_intents/object#payment_intent_object-next_action

Can you show me which example you followed?

aivils: Example https://stripe.com/docs/payments/payment-intents/quickstart#creating-with-manual-confirmation 
Yes, the current implementation does not take into account redirect_to_url. Do you have any idea how to trigger this case? Our case is confirming a PaymentIntent with Stripe.js. As far I get our case does not trigger redirect_to_url in the Stripe response.

app/services/stripe_service/api/stripe_api_wrapper.rb
```diff
@@ -309,9 +309,11 @@ def create_payment_intent(community:, seller_account_id:, payment_method_id:, am
           currency: currency,
           confirmation_method: 'manual',
           confirm: true,
-          application_fee_amount: fee,
           on_behalf_of: seller_account_id,
-          transfer_data: {destination: seller_account_id},
+          transfer_data: {
+            destination: seller_account_id,
+            amount: amount - fee
```
aivils: @bladealslayer This fixes amount of transfer later paid out for seller.

app/services/stripe_service/api/payments.rb
bladealslayer: I've contacted Stripe support about this to get a clarification.

app/services/stripe_service/api/payments.rb
```diff
@@ -29,17 +30,30 @@ def create_preauth_payment(tx, gateway_fields)
...
+          stripe_charge = stripe_api.charge(
+            community: tx.community_id,
+            token: source_id,
+            seller_account_id: seller_id,
+            amount: total.cents,
+            fee: commission.cents + buyer_commission.cents,
+            currency: total.currency.iso_code,
+            description: description,
+            metadata: metadata)
+        elsif payment_method_id.present?
```
bladealslayer: Let's change the reporting here to log correctly whether this was a payment using payment intent or a charge. Currently we have `report.create_charge_start` and `report.create_charge_success`. We should have equivalent ones for payment intents. Likewise, let's support reporting when capturing is done via a payment intent.

When looking at this, I spotted that there wasn't reporting for cancelling in `cancel_preauth` function. Let's add that. We could report the charge ID / payment intent ID.

app/services/stripe_service/api/stripe_api_wrapper.rb
bladealslayer: @aivils true! A month ago, Stripe did not support passing amount here when creating payment intents, so I didn't know this is now possible. Looks like it works!

app/services/stripe_service/api/payments.rb
aivils: @bladealslayer 
Changes applied. It leads to massive refactoring to avoid duplicate code.

app/services/stripe_service/api/payments.rb
bladealslayer: This is ok, type will always be `use_stripe_sdk` with this type of PaymentIntent handling.

app/services/stripe_service/report.rb
```diff
@@ -127,6 +253,33 @@ def create_payout
...
+      "transaction_id": tx.id,
+      "stripe_seller_id": stripe_account.stripe_seller_id
+    }
+  end
+
+  def cancel_intent
+    {
+      "stripe_op": "cancel_intent",
+      "transaction_id": tx.id,
+      "stripe_seller_id": stripe_account.stripe_seller_id
```
bladealslayer: Minor thing: we could report the `stripe_payment_intent_id` here. We have the ID available both when starting the call and when it completes.

app/services/stripe_service/report.rb
```diff
@@ -117,6 +235,14 @@ def create_charge
     }
   end
 
+  def cancel_charge
+    {
+      "stripe_op": "cancel_charge",
+      "transaction_id": tx.id,
+      "stripe_seller_id": stripe_account.stripe_seller_id
```
bladealslayer: Minor thing: we could report the `stripe_charge_id` here. We have the ID available both when starting the call and when it completes.

PR #3808, opened 2019-08-01
Add query_track to find slow SQL queries

kirillshevch:
I suggest adding a gem query_track which will be show in the console info about SQL queries which slower than 1s (for example, this is configurable). Can add for `development` and `test` environments, because for `staging` and `production` already used newrelic. Developers will be easier to profile the database and improve performance on development.

Example of log (from another project):
console

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -106,6 +106,7 @@ end
 group :development, :test do
   gem 'rubocop', '~> 0.67.2', require: false
   gem 'factory_girl_rails', '~> 4.8.0'
+  gem 'query_track', '~> 0.0.7'
 end
 
 group :development, :staging do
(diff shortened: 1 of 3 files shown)


thomasmalbaux:
Thanks a lot for the suggestion, Kirill! This is indeed super useful when monitoring performances.

However we won't move forward with it here. I believe this is something that should be monitored and managed by each owner of a marketplace open-source instance.

We do monitor that activity for our hosted version, and build improvements based on that, but it's hard to know for sure that someone else would get similar data for their specific website. 

So definitely a good recommendation of something that most people should monitor!

Maybe you could suggest it and share your experience with that gem on Sharetribe's Community forum, at https://www.sharetribe.com/community/? I'm sure that would be helpful to other open-source users!

PR #3829, opened 2019-08-12
[Fix] Let's not a HTML link in en.yml

aivils:


app/views/admin/communities/edit_look_and_feel.haml
```diff
@@ -36,7 +36,8 @@
       .col-12
         = form.label :cover_photo, t(".community_cover_photo"), :class => "input"
         = render :partial => "layouts/info_text", :locals => { :text => (t(".cover_photo_visibility") + " " +  t(".cover_photo_instructions_text_with_dimensions", :width => 1920, :height => 450, :see_how_it_looks_like => link_to(t("admin.communities.edit_details.see_how_it_looks_like"), "/?big_cover_photo=true", :id => "view_cover_photo_link"))).html_safe}
-        = render :partial => "layouts/info_text", :locals => { :text => t(".cover_photo_ready_made").html_safe}
+        - link = link_to t(".we_have_you_covered"), 'https://help.sharetribe.com/en/articles/666084', target: '_blank'
```
thomasmalbaux: Shouldn't that be `#{knowledge_base_url}/en/articles/666084` as it is on other locations (for https://github.com/sharetribe/sharetribe/blob/9c67a1334bfd65dfafa5c6c0d3133957368325f2/app/views/admin/payment_preferences/_payment_settings.haml#L15)? To be consistent and avoid challenges in the future if/when that URL changes.

PR #3906, opened 2019-10-29
email server side newsletter image

aivils:


app/models/listing_image.rb
```diff
@@ -204,4 +214,33 @@ def self.construct_big_style(dimensions, desired_dimensions, max_crop_difference
...
+
+  def newsletter_email_image_url
+    if email_image_hash != self.email_hash || !email_image.present?
+      compose_email_image
+    end
+    email_image.url
+  end
+
+  def email_image_hash
+    Digest::MD5.hexdigest([image_file_name, listing.author.image_file_name].join("|"))
```
bladealslayer: I suppose hash based on file names alone is good enough in practice. Do you see any reason not to include e.g. `image_update_at` and `listing.author.image_updated_at` too? That way probably any image change will result in the composite image to be rebuilt correctly.

aivils: @bladealslayer 
Changes applied. Merged into main email branch.

PR #3935, opened 2019-12-06
[Bug] check payment method presence when initiating payment

aivils:


aivils:
@bladealslayer 
Changes applied!

PR #3954, opened 2019-12-23
Admin panel v2

Charly3X:


app/views/layouts/admin.haml
```diff
@@ -0,0 +1,29 @@
...
+    #root
+      = render partial: 'admin2/topnav', locals: header_props
+      = flash_messages
+      .layout-container
+        = render 'admin2/sidebar'
+        .content-card
+          = yield
+    .mobilemenu-overlay
+    %script{src: 'https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.14.3/umd/popper.min.js'}
+    %script{src: 'https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/js/bootstrap.min.js'}
```
bladealslayer: It would be better if these are not loaded from external URLs, but added to the assets in the source in `vendor/assests/javascripts`. It's safer that way, as we wouldn't be loading scripts from (untrusted) external URLs by default, as well as we wouldn't depend on the external service to be accessible, as we'd serve the scripts from our own infra.

features/support/hapybara.rb
```diff
@@ -22,7 +22,7 @@
   # https://github.com/teamcapybara/capybara/issues/1890
   # Google Chrome 71.0.3578.80
   require 'webdrivers'
-  Webdrivers::Chromedriver.version = '2.46'
+  # Webdrivers::Chromedriver.version = '2.46'
```
bladealslayer: Was this commented out by mistake? Some tests are failing in CI. I couldn't quickly find out why except if it is because of this change.

I remember there were some issues with running tests locally with Chrome. If this was commented out because of that, it might be that the driver change requires some changes to the failing tests, so they would pass in CI.

Charly3X: Yes, I forgot to uncomment, locally my tests did not run on this version of the driver.

app/views/layouts/admin.haml
Charly3X: done!

PR #3965, opened 2020-01-08
[Bug] cast section position to int on reordering of CLP sections

VladFarion:
There is an issue with CLP editor,  when # of sections is higher than 10, ordering fails. 
This PR fixes the issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ This file follows the best practices from [keepachangelog.com](http://keepachang
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


thomasmalbaux:
Thanks for the contribution! We'll have a look :-)

Ping @Luis-RG as you've been working with the CLP editor.

thomasmalbaux:
Thank you very much for the contribution, @VladFarion! This is great :-)

We've added a link to your GitHub page at https://www.sharetribe.com/products/go/updates/

PR #3967, opened 2020-01-10
Ruby update

bladealslayer:
Update to 2.6.3 or even 2.6.5.

Using Debian Buster as base image, which comes with ImageMagick with HEIC support, so full HEIC support can be added on top of this.

~TODO workaround or fix for failing DB SSL connection, due to change in underlying MySQL client libs (MariaDB compat ones).~ Went to using MySQL 5.7 client and libs.


bladealslayer:
TODO workaround or fix for failing DB SSL connection, due to change in underlying MySQL client libs (MariaDB compat ones).

PR #3976, opened 2020-01-27
[Stripe] Updates to API - rake task to update capabilities

aivils:


app/services/stripe_service/api/stripe_api_wrapper.rb
```diff
@@ -264,6 +264,14 @@ def update_account(community:, account_id:, attrs:)
       end
     end
 
+    def update_account_capabilities(community:, account_id:)
+      with_stripe_payment_config(community) do |payment_settings|
+        account = Stripe::Account.retrieve(account_id)
+        account.requested_capabilities = ['card_payments', 'transfers']
+        account.save
```
bladealslayer: Will this result in an API call to Stripe, if the account already had all the requested capabilities? Or will the save be a no-op in that case? Since the update is a long operation (can take hours), it might be more efficient if saving is skipped when unnecessary, in case we interrupt and restart the update.

app/services/stripe_service/api/stripe_api_wrapper.rb
```diff
@@ -264,6 +264,14 @@ def update_account(community:, account_id:, attrs:)
       end
     end
 
+    def update_account_capabilities(community:, account_id:)
+      with_stripe_payment_config(community) do |payment_settings|
+        account = Stripe::Account.retrieve(account_id)
+        account.requested_capabilities = ['card_payments', 'transfers']
```
bladealslayer: Let's add `legacy_payments` to the list here too, just in case.

aivils:
@bladealslayer 
Changes applied except multi threading. From my point of view faster is running one week rake task without human attention than resolve weird rails multi threading errors. Another PR https://github.com/sharetribe/sharetribe/pull/3980

bladealslayer:
@aivils alright, we'll run it linearly.

app/services/stripe_service/api/stripe_api_wrapper.rb
```diff
@@ -264,6 +264,17 @@ def update_account(community:, account_id:, attrs:)
       end
     end
 
+    def update_account_capabilities(community:, account_id:)
+      with_stripe_payment_config(community) do |payment_settings|
```
bladealslayer: @aivils Let's wrap this entire operation with a `rescue` so that if there is an exception (e.g. failing network API call) the account is simply logged and skipped, so that the long operation to update all accounts can continue.

aivils:
@bladealslayer 
Changes applied. Now StripeAccount has column api_version. Update will select accounts does not have correct version and update these accounts only. So it is safe to run multiple times over single community or whole DB. Next time we will have another stripe version and we can perform an update to next api version in safe way.
Newly created accounts will contain the api version according the marketplace feature flags.

bladealslayer:
@aivils could you rebase this to get it ready for merging?

aivils:
@bladealslayer 
api_version column is for safe update. In the event of a successful account update, the current version is placed in the api_version column. On failed update api_version stays NULL. rake task does not try to update account with current version in the column api_version. The task can be runned multiple times without any worries. This column will not be used daily.

bladealslayer:
👍 ok. It's a bit wasteful to store and maintain all that data if it isn't used much, but we can remove it in the future, if needed.

PR #3980, opened 2020-01-30
Stripe updates to api rake task run in parallel

aivils:
Unstable: LoadError: Unable to autoload constant ... , Circular dependency detected while autoloading constant  ...


bladealslayer:
Closing this as we decided not to complicate things with parallelism.

PR #4020, opened 2020-03-04
Harmony removal 1 create a data schema

aivils:


db/migrate/20200303075727_create_listing_blocked_dates.rb
```diff
@@ -0,0 +1,10 @@
+class CreateListingBlockedDates < ActiveRecord::Migration[5.2]
+  def change
+    create_table :listing_blocked_dates do |t|
+      t.references :listing
+      t.date :blocked_at
```
bladealslayer: Let's make [community_id, lsiting_id, blocked_at] a unique key in the DB to ensure that two entries for the same date for one listing are not possible, removing chance of race conditions during writing.

db/migrate/20200303075727_create_listing_blocked_dates.rb
```diff
@@ -0,0 +1,10 @@
+class CreateListingBlockedDates < ActiveRecord::Migration[5.2]
+  def change
+    create_table :listing_blocked_dates do |t|
+      t.references :listing
```
bladealslayer: History shows that it's very useful to have community_id in each table. Let's add it here too, even though it's not technically needed.

app/models/listing/blocked_date.rb
```diff
@@ -0,0 +1,26 @@
...
+# Indexes
+#
+#  index_listing_blocked_dates_on_listing_id  (listing_id)
+#
+
+class Listing::BlockedDate < ApplicationRecord
+  belongs_to :listing
+
+  scope :in_period, ->(start_on, end_on) do
+    where('blocked_at >= ? AND blocked_at <= ?', start_on, end_on)
```
bladealslayer: We could add index also on community_id, listing_id and blocked_at, so that queries like this use index.

aivils:
@bladealslayer 
Sorry, I can't accept the suggestion to add community_id. RoR simply does not work like that. The outcome of this change is predictable - people start fighting against the framework instead of using the framework. Indexes added.  You analysed small part only. listings - transactions - bookings already store all "blocked dates" except dates seller marked as "blocked" knowingly without real transaction. Listing::BlockedDate  -  seller marked the date as "blocked" knowingly without real transaction.

bladealslayer:
@aivils ok about the `community_id`. It's not a big deal.

I understand that the total availability is affected by transactions and bookings as well. When talking about race conditions, I meant specificaly only the "BlockedDate". Right now, I don't see anything preventing from a date being blocked twice by accident, in case of two concurrent requests to "create a blocked date on day X". There is a race condition where Rails will check that the date is not there and then proceed to create the row in the DB and if those two happen in parallel, two rows may be created. Am I missing something? Does rails do table locking automatically somehow?

So, I suggest we make the index you added also a `UNIQUE` key, meaning that the database itself will prevent two rows for same `[listing_id, date]` to be created.

aivils:
@bladealslayer 
Unnecessary code removed. UNIQUE added in the same last commit as it is very small change in the migration.

PR #4037, opened 2020-03-16
Harmony removal 2 implement double writes

aivils:


aivils:
@bladealslayer 
This is finalised. Unfortunately, the last commit took a long time.

app/state_machines/transaction_process_state_machine.rb
```diff
@@ -76,13 +82,28 @@ class TransactionProcessStateMachine
     send_new_transaction_email(transaction) if transaction.conversation.payment?
   end
 
+  # "guard_transition" is before SQL BEGIN-COMMIT block
+  # instead, "before_transition" is inside the block
+  before_transition(to: :preauthorized) do |transaction, transition|
```
bladealslayer: Should there be the same kind of block for transition to `:payment_intent_requires_action`? That state also is blocking availability, so it would be better if we guard when it is first entered. No?

app/state_machines/transaction_process_state_machine.rb
```diff
@@ -76,13 +82,29 @@ class TransactionProcessStateMachine
...
+    validate_before_preauthorized(transaction, transition)
+  end
+
+  before_transition(to: :payment_intent_requires_action) do |transaction, transition|
```
aivils: @bladealslayer 
Oops, now it validates both.

bladealslayer: 👍 I guess we also need one:

```ruby
after_transition_failure(to: :payment_intent_requires_action) do |transaction|
    void_payment(transaction)
  end
```

right, @aivils?

PR #4039, opened 2020-03-17
Remove Amazon as overwritten ActiveStorage config on development

danielpaul:
`/config/environments/development.rb` already has `config.active_storage.service = :local` in line 73. The Amazon storage config is overwriting this setting at the bottom of the file. Must have been a wrongly checked in line? 

Was causing issues getting the codebase running locally with this config without setting up Amazon for local file storage.


bladealslayer:
Closing this in favor of #4143

PR #4040, opened 2020-03-19
[Stripe] Support bank accounts in a different country than the provider's address

aivils:


thomasmalbaux:
We've decided not to move forward with that feature for the time being.

PR #4053, opened 2020-04-04
Fix license formatting issue

kemitchell:
I noticed a formatting error in the license text, where a following
paragraph was enumerated in a list, rather than the final element.

I am NOT reviewing or approving the legal substance of the license.  Do
not rely on this patch for legal advice.  See https://notlegaladvice.law
for more information.


thomasmalbaux:
Hi there,

and thanks a lot, good catch! This was embarrassing, indeed I've just made that typo/mistake when adding the licence file.

So, thanks a lot for the help. And thanks for the [great and interesting review of the Sharetribe Community Public License](https://writing.kemitchell.com/2020/04/04/Sharetribe-1.0.html) you've shared!

PR #4093, opened 2020-05-25
admin v2 order types

Charly3X:


app/controllers/admin/listing_shapes_controller.rb
```diff
@@ -222,6 +220,7 @@ def common_locals(form:, count:, process_summary:, available_locs:)
       uneditable_fields: uneditable_fields(process_summary, form[:author_is_seller]),
       shape: FormViewLayer.shape_to_locals(form),
       count: count,
+      harmony_in_use: APP_CONFIG.harmony_api_in_use.to_s == "true",
```
bladealslayer, line 223: I don't think this should be here, as Harmony is no longer used at all and this configuration variable was removed. Is there a reason why it was added here?

app/controllers/admin2/listings/order_types_controller.rb
```diff
@@ -0,0 +1,173 @@
...
+    def filter_uneditable_fields(shape, process_summary)
+      uneditable_keys = uneditable_fields(process_summary, shape[:author_is_seller]).select { |_, uneditable| uneditable }.keys
+      shape.except(*uneditable_keys)
+    end
+
+    def common_locals(form:, count:, process_summary:, available_locs:)
+      { uneditable_fields: uneditable_fields(process_summary, form[:author_is_seller]),
+        shape: FormViewLayer.shape_to_locals(form),
+        count: count,
+        harmony_in_use: APP_CONFIG.harmony_api_in_use.to_s == 'true',
```
bladealslayer, line 161: Same as above: I don't think this should be here, as Harmony is no longer used at all and this configuration variable was removed. Is there a reason why it was added here?

PR #4107, opened 2020-06-03
Fix translation for de.listings.unit_types.person

andreasmischke:
Minor bug, but the misspelled word is visible very prominently.


thomasmalbaux:
Thanks a lot for the report!

We actually use another tool to manage translations and I've made the fix there already, it will be reflected here too soon.

Thanks again!

PR #4129, opened 2020-06-17
Fix break pending consent page

Charly3X:


thomasmalbaux:
Fixed differently, in `https://github.com/sharetribe/sharetribe/pull/4132`, by always passing the %{count} argument.

PR #4145, opened 2020-07-06
Fix fuzzy location map

Charly3X:


thomasmalbaux:
**Active map view on listing page doesn't load after clicking on the screenshot view**

With fuzzy location enabled and keyword search as the search type, clicking on the map screenshot on a listing page wouldn't load the active map view at all.

This is now fixed.

PR #4160, opened 2020-07-22
fix bug with has_paypal field after user CSV export

mrdya4enko:


bladealslayer:
Paypal account data should be fetched via the `PaypalService` and not directly via Rails models. I'm closing this in favour of #4196

PR #4236, opened 2020-10-26
Add script for pulling all WTI translation projects

bladealslayer:
Running `./script/wti-pull-all.sh` from the project root directory will
pull all translations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,9 @@ coverage/
 .env
 .session.vim
 
+# WTI real config file. See .wti.main and .wti.admin instead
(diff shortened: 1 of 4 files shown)


thomasmalbaux:
I'll close this for now but might get back to it in the future. @bladealslayer 

Right now, I prefer to have a bit more control over when new Admin panel translations should be pulled, as it feels important not to release partial translations. So I don't mind the extra manual step and prefer not to bring changes to the flow for the rest of the team, who is quite familiar with the current setup.

PR #4253, opened 2020-11-06
Example route rename handling for admin2

bladealslayer:
Allows all path helpers to remain the same


thomasmalbaux:
Used at https://github.com/sharetribe/sharetribe/pull/4263

PR #4264, opened 2020-11-18
Added 'for' attribute on agreement checkbox label

bipashant:


thomasmalbaux:
Thanks for the contribution!

Is there any specific reason on why to focus on that specific label? That pattern hasn't been used much elsewhere in the code (which is probably not the best idea) so I wonder why improve only that specific one?

bipashant:
If you see the transaction agreement checkbox, the user needs to click on the checkbox input to accept it, clicking on the label does not do anything. So adding the `for` attribute on the label will link the checkbox with its label. 

It can be done in two way,
1) By nesting checkbox tag within the label, as we are doing on the homepage filter
homepage-filter
2) By adding the for attribute on the label, as we are doing on the Admin panel
Admin delete logo

thomasmalbaux:
Thanks for the details. And this is a good contribution indeed!

I was just curious why this label specifically: for example, have you noticed others that could be fixed in a similar manner?

bipashant:
Thanks, Thomas!
it is the only thing I noticed. I will keep exploring.

PR #4280, opened 2020-11-30
Typo correction on StateMachine comments.

bipashant:
Renaming `:initialized` to `:initiated` since the `:initialized` is not a valid state.

### DIFF
--- a/app/services/transaction_service/state_machine.rb
+++ b/app/services/transaction_service/state_machine.rb
@@ -27,7 +27,7 @@ module TransactionService
           state_machine.transition_to(new_status, metadata_hash)
         rescue StandardError
           # If transaction failed to transition to it's first state (e.g.
-          # :initialized or :free), mark it as deleted. Reload is needed, in
+          # :initiated or :free), mark it as deleted. Reload is needed, in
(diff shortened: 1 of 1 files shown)


thomasmalbaux:
Thanks for the contribution! Indeed, having correct information in comments is important.

I've noticed that, in the past few days, you've suggested a few similar fixes about typos. Maybe, instead of creating one commit and one PR for each, it would make sense to bundle all the ones you've found and think are worth fixing in a single PR, focused on that "fixing typos" theme?

bipashant:
Hi @thomasmalbaux 
Thanks. I’m working on an amazing Sharetribe customisation project that need to understand code in deeper level.So, whenever I see the thing that can be improved, I’m making the pull requests. 
Sending a single pull request is also a good idea but I’m personally prefer small pull requests and also I’m not sure much I can contribute in future. So, If you’re okay, we can merge this and I will send a separate one in future If I find anything to contribute.

thomasmalbaux:
All right then. Fixing typos is always appreciated but indeed, that feels a bit weird to create commits and PRs for each letter fix in comments ;-)

Good to hear about that exciting project, keep up the good work!

PR #4289, opened 2020-12-10
[Custom script] Make the Custom script plan_feature dependent

Charly3X:


app/services/plan_service/api/no_plans.rb
```diff
@@ -24,7 +24,7 @@ def get_current(community_id:)
         Plan.call(
         community_id: community_id,
         status: :active,
-        features: { deletable: true, admin_email: true, whitelabel: true },
+        features: { deletable: true, admin_email: true, whitelabel: true, custom_script: false },
```
bladealslayer, line 27: If I remember correctly, this defines the features for a standalone Go CE installation. There, custom_script should be true, right @thomasmalbaux?

thomasmalbaux: Indeed. Changed to true.

PR #4295, opened 2020-12-18
Add Stripe support for Hungary

thomasmalbaux:


app/services/inbox_service.rb
```diff
@@ -190,6 +190,13 @@ def extend_conversation(conversation)
     )
   end
 
+  # https://stripe.com/docs/currencies#special-cases
+  def divisible_100(sum, currency)
```
bladealslayer, line 194: This seems unused now.

bladealslayer (commented):
Looks good to me! There's some leftover code that seems unused.

PR #4307, opened 2021-01-04
[Email layout v2] Remove ability to change feature flag from Admin panels

Charly3X:


thomasmalbaux:
The first step of the sunset of the legacy layout for emails is planned for 2021-02-01. This PR won't be merged until them, on purpose.

PR #4335, opened 2021-02-08
[Admin2] Custom Landing Page editor

Charly3X:


db/structure.sql
```diff
@@ -367,6 +367,20 @@ CREATE TABLE `community_customizations` (
   KEY `index_community_customizations_on_community_id` (`community_id`) USING BTREE
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 /*!40101 SET character_set_client = @saved_cs_client */;
+DROP TABLE IF EXISTS `community_domain_checkers`;
```
bladealslayer, line 370: This is probably added to the PR by mistake as it is related to old migration that was removed from the code. You might need to reset your local databases to avoid getting that back in.

app/models/domain_setup.rb
```diff
@@ -15,7 +15,6 @@
 #
 #  index_domain_setups_on_community_id          (community_id) UNIQUE
 #  index_domain_setups_on_critical_error        (critical_error)
-#  index_domain_setups_on_domain                (domain) UNIQUE
```
bladealslayer, line 18: This is also related to the accidental update to db structure that should not be in this PR.

PR #4387, opened 2023-08-20
Use a valid git clone link in README.md

romanzes:
I was puzzled as to why my `git clone` command wasn't doing anything and then I realised that the link in the installation guide is invalid.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Before you get started, the following needs to be installed:
 1.  Get the code. Clone this git repository and check out the latest release:
 
     ```bash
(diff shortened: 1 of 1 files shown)


thomasmalbaux:
Many thanks @romanzes! Good catch!

This has been fixed at https://github.com/sharetribe/sharetribe/commit/c3b07d829c77887b8cd11558c1a98d9a44ae538c and your handle has been added to the changelog file at https://github.com/sharetribe/sharetribe/commit/424b78389f5398070305c00dfeb57b4ee054789b.

Thanks again for making this less confusing for future Go CE users!

PR #4388, opened 2023-11-07
Update et.yml

villu164:
**näide.com** is not the same as **example.com**

Read more - https://secret.näide.com/

## Summary
Localization was fixed, to use example.com domain not a private domain

## Motivation
Security related. Not an urgent(probably seen as a no-issue by many) security issue, but could/should trigger the thought process of the same "example.com" email in other languages

## Testing
Compare the output of
[https://example.com/](https://youtu.be/dQw4w9WgXcQ?Seriously+Watch+Where+You+Are+Clicking)
https://näide.com/
https://secret.example.com/
https://secret.näide.com/
To verify that they are not the same

### DIFF
--- a/config/locales/et.yml
+++ b/config/locales/et.yml
@@ -1122,7 +1122,7 @@ et:
         plaanite saata palju kutseid, peaksite kasutama e-posti saatmise tööriista.
         Vaadake %{this_article_link}, et leida sobivate e-teenuste kohta allikaid.'
       this_article_link: see artikkel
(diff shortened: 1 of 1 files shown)


thomasmalbaux:
Thanks, good catch!

We use another tool to update translations, and it has now been fixed via https://github.com/sharetribe/sharetribe/commit/a582f6ffdd08a8d0ed257180d168caf006e9b723. This has also been noted in the changelog at https://github.com/sharetribe/sharetribe/commit/c134646648082d6387d5850e78c949560dc8f4c2.

Thanks again for your help!

